PR #91, opened 2016-07-08
managed row-copy

shlomi-noach:
WIP

phase 2: copy row data via app (configurable)

This will be an alternative to the `insert into...select` that executes on the master. Instead, we can read row data from the replica, into the app, then write it back to the master.

This further decouples the write load from master's load, or in fact decouples the entire migration from the master. The master just so happens to be the one where we apply the new table on.


DamianZaremba:
As an extension to this, it could be interesting to support filtering the table by some arbitrary query; this would essentially support sharding a table onto a new replica set (assuming hook functionality to change the write host externally rather than doing the table flip). It may well be outside the scope of the tool though.

tomkrouper:
@DamianZaremba, I would say adding a `--where` type option, like you are suggesting would be separate from this, and a good idea. Can you open an issue referencing this one with any details you'd like to see in such a feature?

PR #124, opened 2016-08-02
all: use time.Since() instead of time.Now().Sub

dgryski:
Patch created with:
    gofmt -w -r 'time.Now().Sub(a) -> time.Since(a)' .


SuperQ:
:+1:

PR #126, opened 2016-08-03
Set executable bits on build.sh

dveeden:
change permissions on build.sh from 644 to 755.
This makes it possible to use ./build.sh instead of bash build.sh
I believe this is a more common way of executing shell scripts.


SuperQ:
Yup, better.  Please update documentation references that refer to `bash build.sh`.

Issue #127, opened 2016-08-03
Mix of --<option> and -<option>


dveeden:
The output of `gh-ost --help` shows short options like `-database` in go-style.
However `gh-ost` without options shows `2016-08-03 09:16:34 FATAL --database must be provided and database name must not be empty` with `--database` instead.

Let's settle for the go-style one-dash options instead of the GNU style long options.
This prevents the later use Unix style short options, but that should be fine

shlomi-noach:
Oh I'm so ambivalent about this. While I realize this is a `golang` project, I really do like adhering to unix-style `--long-options`.

Given that 

> The output of gh-ost --help shows short options like -database in go-style

it makes sense to approve https://github.com/github/gh-ost/pull/128, but... I'd like to see if we can have it otherwise.

zmoazeni:
This really confused me using the `--help`. Personally I'm in favor of using the double dash since that's a bit more standard. But I don't have a strong preference.

However, having documentation+help consistent would be a great.

zmoazeni:
> it makes sense to approve #128, but... I'd like to see if we can have it otherwise.

Does this mean you'd prefer the opposite of 128? As in, instead of changing the documentation, change the output of `--help` to double dash?

I can take a stab at that PR if so.

PR #128, opened 2016-08-03
Use go style short options everywhere

dveeden:
e.g. change `--database` to `-database`

Also change references to `gh-osc` to `gh-ost`. In README.md there is one reference left, but that's to describe the history.

Related Issue: https://github.com/github/gh-ost/issues/127


SuperQ:
:+1:

PR #131, opened 2016-08-03
Don't continue if build fails

dveeden:
Stop build.sh if any error pops up.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -1,22 +1,36 @@
 #!/bin/bash
 #
 #
-RELEASE_VERSION="1.0.8"
 
+function build {
+    osname=$1
+    osshort=$2
+    GOOS=$3
+    GOARCH=$4
+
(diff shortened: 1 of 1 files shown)


SuperQ:
I usually prefer to catch errors explicitly rather than do `set -e`.  The reasoning behind this is bash doesn't give you a traceback when an error happens.  It makes failures difficult to debug.

dveeden:
I converted build to a function and let it check the return code of `go build`. This also allows easier adding of other operating systems. I tested FreeBSD and that worked. A build on windows failed because syslog.* is not working there.

I don't think building on FreeBSD should be done by default as testing is probably only done for macOS and Linux.

SuperQ:
Very nice. :+1:

shlomi-noach:
Merged manually, thank you!

Issue #136, opened 2016-08-05
Error message about privileges is not helpful


dveeden:
Example:
```
2016-08-05 08:57:13 ERROR User has insufficient privileges for migration.
2016-08-05 08:57:13 FATAL 2016-08-05 08:57:13 ERROR User has insufficient privileges for migration.
```

It should tell the user:
* what is needed
* what it found

shlomi-noach:
Closed by #137

Issue #132, opened 2016-08-03
Getting by without SUPER privilege


teu:
Hi, 
we have case in which we would like to use gh-ost with AWS RDS. However, the problem with this managed database store is that it does not grant us the SUPER privilege.
Is there any way of using gh-ost and not having SUPER on the server?
Thanks

shlomi-noach:
looking into this

shlomi-noach:
The first blocker I see right away is that `SUPER` is required for `STOP SLAVE` and `START SLAVE`.

Why do we need these two in the first place? Because `gh-ost` may be asked to switch `binlog_format` to `ROW`. In which case it issues:

- `set global binlog_format='ROW'`
- `set session binlog_format='ROW'`

However if the replica was running with `STATEMENT`, merely changing to `ROW` means nothing; the replication must be `STOP`ped and `START`ed for the setting to take effect. `gh-ost` does not want to assume it was set up correctly before running; therefore it _always_ issues a `STOP & START`.

We _could_ add a flag saying `--trust-me-this-is-rbr` to skip the above. I don't like it, but it's possible.

SuperQ:
It should be possible to pre-test for RBR and not require `SUPER`.

shlomi-noach:
@SuperQ how can you you confirm without doubt the replica is using RBR? Please read again my comment.

shlomi-noach:
To elaborate: you see a replica. It _says_ it has `binlog_format=ROW`. Fact: it may replicate via `SBR`, because _maybe_ someone `set global binlog_format=ROW` without restarting replication.

It sounds far fetched, but in fact is a classic case in `--switch-to-rbr`

shlomi-noach:
https://github.com/github/gh-ost/pull/156 will close this Issue, and will make it possible to run `gh-ost` without `SUPER` privilege.

shlomi-noach:
I'm happy to merge #156 , and see thoughts on SBR/RBR in comment: https://github.com/github/gh-ost/pull/156#issuecomment-240351262

shlomi-noach:
`SUPER` is now not required for normal migration, see #156 and release https://github.com/github/gh-ost/releases/tag/v1.0.10

It is still required in `--test-on-replica`, see discussion on https://github.com/github/gh-ost/issues/162 and on #163

shlomi-noach:
[...]

PR #137, opened 2016-08-05
More user friendly error reporting for privileges

dveeden:
### Description

This makes errors about privileges a bit more user friendly.
I had granted REPLICATION CLIENT instead of REPLICATION SLAVE and this was more difficult to troubleshoot that it should have been.

Related Issue: https://github.com/github/gh-ost/issues/136

New output:
```
2016-08-05 11:39:07 DEBUG Privileges: Super: true, REPLICATION SLAVE: false, ALL on *.*: false, ALL on `sakila`.*: true
2016-08-05 11:39:07 ERROR User has insufficient privileges for migration. Needed: SUPER, REPLICATION SLAVE and ALL on `sakila`.*
2016-08-05 11:39:07 FATAL 2016-08-05 11:39:07 ERROR User has insufficient privileges for migration. Needed: SUPER, REPLICATION SLAVE and ALL on `sakila`.*
```

### DIFF
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -190,10 +190,11 @@ func (this *Inspector) validateGrants() error {
 		return nil
 	}
 	if foundSuper && foundReplicationSlave && foundDBAll {
-		log.Infof("User has SUPER, REPLICATION SLAVE privileges, and has ALL privileges on `%s`", this.migrationContext.DatabaseName)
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Thank you. I'm also going to pursue https://github.com/github/gh-ost/issues/132, so this is likely to change.

Issue #138, opened 2016-08-05
gh-ost tries to update generated columns


dveeden:
If the old table has a generated column then `gh-ost` tries to insert on the generated column on the new table.
It does not detect the error and tries to continue.

The solution is:
 1. detect the failure and stop 
 1. skip generated columns on insert to avoid the error
 1. Get Oracle to add an option to MySQL to blackhole inserts on generated columns (This is wrong..)

Example table:
```SQL
CREATE TABLE `salaries` (
  `emp_no` int(11) NOT NULL,
  `salary` int(11) NOT NULL,
  `from_date` date NOT NULL,
  `to_date` date NOT NULL,
  `yearly_per_year` int(11) GENERATED ALWAYS AS ((`salary` * 12)) VIRTUAL,
  PRIMARY KEY (`emp_no`,`from_date`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4
```

Example output
```
$ sudo rm -rf /tmp/gh-ost.employees.salaries.sock; /usr/lib/gh-ost/gh-ost/gh-ost -table salaries -throttle-control-replicas replica1,replica2 -alter "ADD COLUMN test1 tinyint unsigned DEFAULT NULL" -conf /home/dvaneeden/.my_gh-ost.cnf -database employees -host $(hostname -f) -initially-drop-old-table -initially-drop-ghost-table -execute -test-on-replica
# Migrating `employees`.`salaries`; Ghost table is `employees`.`_salaries_gho`
# Migrating dvaneeden-test-2002.example.com:3306; inspecting dvaneeden-test-2002.example.com:3306; executing on dvaneeden-test-2002.example.com
# Migration started at Fri Aug 05 13:32:35 +0000 2016
# chunk-size: 1000; max-lag-millis: 1500ms; max-load: ; critical-load: ; nice-ratio: 0.000000
# throttle-additional-flag-file: /tmp/gh-ost.throttle 
# Serving on unix socket: /tmp/gh-ost.employees.salaries.sock
Copy: 0/2838426 0.0%; Applied: 0; Backlog: 0/100; Time: 0s(total), 0s(copy); streamer: binlog.000017:100461102; ETA: N/A
2016-08-05 13:32:36 ERROR Error 3105: The value specified for generated column 'yearly_per_year' in table '_salaries_gho' is not allowed.
[...]

PR #140, opened 2016-08-05
Support virtual columns

dveeden:
### Description

Don't try to insert on virtual columns when copying rows.
I only did a very basic test, so maybe run this through some more testing?

Related Issue: https://github.com/github/gh-ost/issues/138

### DIFF
--- a/doc/requirements-and-limitations.md
+++ b/doc/requirements-and-limitations.md
@@ -13,7 +13,6 @@
 
 - Foreign keys not supported. They may be supported in the future, to some extent.
 - Triggers are not supported. They may be supported in the future.
-- MySQL 5.7 generated columns are not supported. They may be supported in the future.
 - The two _before_ & _after_ tables must share some `UNIQUE KEY`. Such key would be used by `gh-ost` to iterate the table.
   - As an example, if your table has a single `UNIQUE KEY` and no `PRIMARY KEY`, and you wish to replace it with a `PRIMARY KEY`, you will need two migrations: one to add the `PRIMARY KEY` (this migration will use the existing `UNIQUE KEY`), another to drop the now redundant `UNIQUE KEY` (this migration will use the `PRIMARY KEY`).
 - The chosen migration key must not include columns with `NULL` values.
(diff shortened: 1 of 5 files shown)


shlomi-noach:
Yes, this will be a bit lengthier to handle; we may eventually also need to consider what happens when the alter statement changes a generated column into a non generated column. I need to wrap my head around this, but this sounds non-trivial.

shlomi-noach:
Any thoughts on possible scenarios is appreciated!

mizioumt:
This should work I think:

* do not INSERT into any generated column
* if a column is changed from generated to non-generated (much better done by DROP COLUMN/ADD COLUMN), then the resulting column should be treated as a new one, that is we do not INSERT to it.

So, if column `x` exists as a generated column in either old or new table, we do not insert into it.

There is an exception, though: ALTER drops a generated column `g` and also renames existing column `c` to `g`.

shlomi-noach:
> There is an exception, though: ALTER drops a generated column g and also renames existing column c to g. 

this kind of complication can happen with normal columns as well; and it is a perfectly good way to shoot one's own foot. I'm not very concerned about that one.

> if a column is changed from generated to non-generated (much better done by DROP COLUMN/ADD COLUMN), then the resulting column should be treated as a new one, that is we do not INSERT to it.

What is the reasoning for not inserting to it?

mizioumt:
well maybe I was wrong here the value of the original column should be INSERTed into the new one in the spirit of CHANGE/MODIFY as opposed to ADD/DROP. 
For the columns that end up virtual in the new table, DEFAULT should be inserted which is allowed, or they should not be in the list of columns in the INSERT.

artemy-osipov:
@shlomi-noach, it would be nice to add only support for tables with virtual columns ignoring insert on such columns while changing virtual columns can be implemented in another time/PR. For now we can just forbid changing virtual columns.

shlomi-noach:
_Finally_ getting to it. Since this PR is so old and has many conflicts, and seeing that I implemented it slightly differently, I resubmitted as https://github.com/github/gh-ost/pull/595, using parts of @dveeden 's contributions.

shlomi-noach:
I'm closing this PR in favor of #595

Issue #142, opened 2016-08-05
Too many colons in address when using IPv6 address literals for replication


Roguelazer:
If replication is set up using IPv6 address literals for the master host name, `gh-ost` dies with "2016-08-05 23:25:51 ERROR dial tcp: too many colons in address fd00:aaaa:bbbb::1:3306".

As far as I can tell, the mysql driver seems to expect IPv6 addresses to be written as `[a:b:c::d]:3306`, so this may just be a matter of detecting when `Master_Host` is an IPv6 literal and adding some square brackets?

shlomi-noach:
@Roguelazer thank you -- did you happen to confirm your change works with `ipv6`? I don't have an immediate environment to test this.

Roguelazer:
yeah, it seems to.

Roguelazer:
(fixed in 1.0.9)

PR #143, opened 2016-08-05
detect IPv6 literals and wrap them in square brackets in GetDBUri

Roguelazer:
Related issue: https://github.com/github/gh-ost/issues/142

### Description

This PR detects IPv6 address literals (using `net.ParseIP`) and wraps them in square brackets for the dburi so that we can actually use IPv6 replication configurations.

I don't typically write go, so I'm not sure if this is the idiomatic way to detect IP address families. It appears that `net.IP` sometimes packs an IPv4 address into a 16-byte char slice, so we can't just compare the length of the slice with `net.IPv4len`, but maybe there's some better function out there somewhere. This seems to work in limited testing.

### DIFF
--- a/go/mysql/connection.go
+++ b/go/mysql/connection.go
@@ -7,6 +7,7 @@ package mysql
 
 import (
 	"fmt"
+	"net"
 )
 
 // ConnectionConfig is the minimal configuration required to connect to a MySQL server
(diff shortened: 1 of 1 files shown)


go/mysql/connection.go
```diff
@@ -47,5 +48,11 @@ func (this *ConnectionConfig) Equals(other *ConnectionConfig) bool {
 }
 
 func (this *ConnectionConfig) GetDBUri(databaseName string) string {
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s", this.User, this.Password, this.Key.Hostname, this.Key.Port, databaseName)
+	var ip = net.ParseIP(this.Key.Hostname)
+	if (ip != nil) && (ip.To4() == nil) {
```
shlomi-noach: Is there a reason why not to test `To16()` instead of `To4()`?

dgryski: You don't need to check `ip != nil` before calling `ip.To4()`; `To4` will return `nil` as expected.

shlomi-noach: @dgryski I believe the `ParseIP` function can return `nil`, hence the check

dgryski: `ParseIP` can return nil, but it's a typed `nil` that you can still call methods on.  Nil receivers are valid in Go.

https://play.golang.org/p/imTbiRbYmB

shlomi-noach: OMG

Roguelazer: To answer your original question, the Go standard library will pack an IPv4 address (with zeros) into a 16-byte slice if you call .To16 on it, so it's not a useful check.

It frustrates me immensely that the language doesn't ship with a `.AddressFamily()` discriminator or a `.IsIpv4()` function or a way to call the `inet_pton` function.

For your second question, it seems stylistically preferable to me to check `ip != nil` because that does something conceptually different – it checks whether the passed input is an IP address literal at all. I'm of the school that explicit is always better than implicit.

shlomi-noach: LGTM, thanks for clarifying

Issue #141, opened 2016-08-05
gh-ost should display a warning if it's run against a table with triggers defined


cswingler:
From the docs:

> Triggers are not supported. They may be supported in the future.

The output of `gh-ost` should probably mention this as at least a warning if it detects any triggers on the table to be altered. As it stands now, it silently leaves them attached to the old table, and if `--ok-to-drop-table` is defined, silently deletes them.

shlomi-noach:
`gh-ost` should fail if triggers exist on the table.

I will later open an Issue for community contribution for supporting triggers -- this is possible to do.

shlomi-noach:
closed by https://github.com/github/gh-ost/pull/152

PR #152, opened 2016-08-11
testing for trigger existence

shlomi-noach:
Related issue: https://github.com/github/gh-ost/issues/141

### Description

This PR checks for existence of triggers on the migrated table; if any trigger exists, `gh-ost` bails out with error.


shlomi-noach:
Closes #141

Issue #135, opened 2016-08-04
[Potential Bug] log_slave_updates needed for master only migration


kbussche:
Using gh-ost 1.0.8 on a 5.7 master, setup with (log-bin,binlog_format=ROW, sync_binlog=1).

When I try to run using the --allow-on-master and --allow-master-master it exits with:

FATAL myserver:3306 must have log_slave_updates enabled

I have read the docs but could of missed something. Its my understanding that when doing the master only migration that it should be reading from the bin log, and log_slave_updates would be un-needed in this situation.

dveeden:
This is checked in `validateBinlogs()` in `go/logic/inspect.go`

`-allow-master-master` is stored in `AllowedRunningOnMaster`, but checking that is not enough.

Looks like if `InspectorIsAlsoApplier()` is true we should not check for log_slave_updates.

shlomi-noach:
Thank you. This will be addressed in a few days.

shlomi-noach:
@dveeden your analysis is correct; I will need to push the test till later, though, since the distinction of `InspectorIsAlsoApplier()` is only possible at this time after the `insupector`'s tests

shlomi-noach:
@kbussche @dveeden , https://github.com/github/gh-ost/pull/155 will fix and close this issue; will merge tomorrow.

shlomi-noach:
closed by https://github.com/github/gh-ost/pull/155

PR #155, opened 2016-08-11
relaxed check for log_slave_updates

shlomi-noach:
Related issue: #135 

`log_slave_updates` doesn't have to be set if we're directly:

- migrating on master
- migrating the replica

`log_slave_updates` _is_ required when doing normal read-from-replica-write-on-master or when `--test-on-replica`

### DIFF
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -243,16 +243,13 @@ func (this *Inspector) applyBinlogFormat() error {
 
 // validateBinlogs checks that binary log configuration is good to go
 func (this *Inspector) validateBinlogs() error {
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Closes #135

go/logic/inspect.go
```diff
@@ -283,6 +280,21 @@ func (this *Inspector) validateBinlogs() error {
 	return nil
 }
 
+// ValidateLogSlaveUpdates checks that binary log log_slave_updates is set. This test is not required when migrating on replica or when migrating directly on master
+func (this *Inspector) ValidateLogSlaveUpdates() error {
```
SuperQ: This seems to be an internal func for the logic package, should be `validateLogSlaveUpdates`

PR #156, opened 2016-08-12
accepting `--assume-rbr`, remove `SUPER` requirement

shlomi-noach:
Related issue: https://github.com/github/gh-ost/issues/132

### Avoid restarting replication, avoid need for SUPER

`gh-ost` cannot verify the replication thread truly uses `RBR`, and wishes to restart replication (`stop slave; start slave;`). However that requires `SUPER` privilege.
This PR supports `--assume-rbr` where the user clearly indicates "yes, this is truly and really RBR, and the replica is using RBR". If `binlog_format=ROW` then `gh-ost` is happy to proceed without restarting replication, hence without requiring `SUPER` privilege.

This PR [briefly explain what is does]

### DIFF
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -16,6 +16,12 @@ When your migration issues a column rename (`change column old_name new_name ...
 
 If you think `gh-ost` is mistaken and that there's actually no _rename_ involved, you may pass `--skip-renamed-columns` instead. This will cause `gh-ost` to disassociate the column values; data will not be copied between those columns.
 
+### assume-rbr
+
+If you happen to _know_ your servers use RBR (Row Based Replication, i.e. `binlog_format=ROW`), you may specify `--assume-rbr`. This skips a verification step where `gh-ost` would issue a `STOP SLAVE; START SLAVE`.
+Skipping this step means `gh-ost` would not need the `SUPER` privilege in order to operate.
+You may want to use this on Amazon RDS
+
 ### conf
 
 `--conf=/path/to/my.cnf`: file where credentials are specified. Should be in (or contain) the following format:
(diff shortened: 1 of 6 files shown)


SuperQ:
Normally I would call this `--assume-rbr`, but "trust" is also fine.

:+1:

pbitty:
Hi @shlomi-noach, have you considered the approach of detecting whether `ROW` format is actually being used?  By sending some test updates to a table, we could see the format of the events arriving in the binlog.

Something like this:
1. On master: `insert into _rbrtest_gho (a) values (1), (2), (3);`
2. Wait for the corresponding `Write_rows` events in the binlog

If the event arrives as a `Query` event then we know we're using `STATEMENT` format and can return the appropriate error message.

PS. Kudos on coming up with such a novel approach to live schema changes.

shlomi-noach:
@pbitty this is a good suggestion, thank you! It may complicate stuff a bit, but seems worthwhile. I think I might first iterate with `--trust-rbr` and then iterate again with your idea.

shlomi-noach:
@SuperQ `--assume-rbr` does sound better. Will change.

SuperQ:
Maybe we can file a feature request upstream to extend the SHOW SLAVE STATUS to include the binlog mode for the running thread.

SuperQ:
:+1: Nice

shlomi-noach:
Noting down that in the event `gh-ost` _thinks_ replication uses RBR where in fact it does not, the current behavior for migration is to _just not work_, because `gh-ost` will not be able to intercept its very own changelog table events.

I'm actually good to merge this, and iterate later with double-safety checks, such as:

- oh, I'm waiting for too long for my own changelog event, and replication seems fine, let's inject a noop DDL statement (`drop view if exists no_such_view_1234567890randomnumber`) and see if we're able to intercept _that_, hence SBR.

shlomi-noach:
Need to add documentation:

- mentioning this flag
- discussing `SUPER`
- and anywhere `SUPER` or RDS is discussed

PR #158, opened 2016-08-15
elaborate output on error in apply dml

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/157

This PR outputs more detail in the case where `ApplyDMLEventQuery` fails to apply a query. 
Reminder: this function gets a DML event from the binary log, constructs a query and arguments from that event, and applies the query with given arguments.

In the case of failure (and after retry count), this function will return the original SQL error, plus query, plus argument values.


shlomi-noach:
Included in https://github.com/github/gh-ost/pull/160

PR #160, opened 2016-08-17
Fix unsigned columns

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/157

This PR checks and fixes unsigned datatypes. 
It seems the binary log does not tell us whether int values are signed or unsigned. The `go-mysql` library assumes signed.

`gh-ost` uses its own understanding of migrated table structure to modify column type (on app side) on the fly from `int` to `uint` on all variations (8, 16, 32, 64).

This fixes a data-corruption problem which can go undetected without `STRICT_ALL_TABLES`.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 #
 #
 
-RELEASE_VERSION="1.0.9"
+RELEASE_VERSION="1.0.10"
 
 function build {
     osname=$1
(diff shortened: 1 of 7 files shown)


shlomi-noach:
This passes production tests

dveeden:
I also hit this issue and can confirm that the fix seem to work.
Please make a new release (1.0.10?).

shlomi-noach:
https://github.com/github/gh-ost/releases/tag/v1.0.10 is released

Issue #161, opened 2016-08-17
Timezone issue with timestamp columns if server doesn't use UTC.


dveeden:
After running a gh-ost with -test-on-replica I found differences in two timestamp columns.

I found this by doing a 
```
select 
  * 
from 
  t1 a 
  inner join _t1_gho b on a.id=b.id 
where 
  a.c1<>b.c1 
  or a.c2<>b.c2
  ...
```

Record 1
```
   timestamp_notified: 2016-08-17 13:08:04
   timestamp_notified: 2016-08-17 11:08:04
```

Record 2
```
   timestamp_notified: 2016-08-17 13:08:19
  timestamp_finalized: 2016-02-01 23:39:30
   timestamp_notified: 2016-08-17 11:08:19
  timestamp_finalized: 2016-02-01 22:39:30
```

From the definition:
```
  `timestamp_notified` timestamp NULL DEFAULT NULL,
  `timestamp_exported` timestamp NULL DEFAULT NULL,
  `timestamp_finalized` timestamp NULL DEFAULT NULL,
```

dveeden:
Environment: CentOS 6, MySQL 5.6.27
/etc/localtime is a link to /usr/share/zoneinfo/Europe/Amsterdam

@@global.time_zone is set to SYSTEM.

shlomi-noach:
Are these all the rows of the table? How many rows does the table have?
Also, to speed up the investigation, would you guess these rows were inserted/added/manipulated throughout the migration process?

shlomi-noach:
Also, please confirm that on `record 2` there are:

- one TZ difference of 2 hours
- one TZ difference of 1 hour

-- in the same record. Please double check this?

dveeden:
Rows with differences: 7085
Total number of rows (from show table status): 31982317

Yes I do think these were modified while gh-ost was running.

dveeden:
Yes I verified the 2h/1h difference... I have no clue about why that is..

dveeden:
This is with gh-ost from the v1.0.8 tag/release.

shlomi-noach:
was `gh-ost` executing from the very same replica host, or from another host? If from another host, can you check what the TZ is on said host?

dveeden:
The very same replica host.

dveeden:
Tested with v1.0.9 but hit #157 so used a build from today's master. It shows the same issue.

```
*************************** 1. row ***************************
                id: 29793038
[...]

PR #165, opened 2016-08-18
Set session timezone to UTC to prevent data drift for timestamp columns

dveeden:
Related Issue: https://github.com/github/gh-ost/issues/161

Do we need to do the SET SESSION and apply statement in a transaction?
Could this introduce issues with date columns?

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -878,6 +878,13 @@ func (this *Applier) ApplyDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) error {
 	// 	return nil
 	// }()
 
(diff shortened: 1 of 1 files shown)


shlomi-noach:
https://github.com/github/gh-ost/pull/172 addresses the transaction requirement

shlomi-noach:
@dveeden I do not see there's an issue with `DATE` columns, since those to not take `@@time_zone` into consideration either way.

shlomi-noach:
@dveeden can you clarify whether the `SET SESSION time_zone = '+00:00'` is tested to work on your side?

shlomi-noach:
Obsoleted by #175

PR #170, opened 2016-08-18
clarifying meaning of sleep-ratio

shlomi-noach:
Documentation change


tomkrouper:
Thank you so much for the clarification :heart: :+1:

Issue #164, opened 2016-08-18
Set `sql_mode=STRICT_ALL_TABLES` for all gh-ost operations


shlomi-noach:
This will set up a safety net for potential data integrity considerations such as depicted in https://github.com/github/gh-ost/issues/157

We will need to set this per-statement, in the transactional space `gh-ost` uses for:

- applying DML event
- copying a row chunk (less important at this time, since data moves around within server space; but this behavior can change in the future)

shlomi-noach:
applied for per-DML event. Closing for now as chunk-copy is not yet in user space anyhow.

Fixed by #191

Issue #149, opened 2016-08-11
support `--sql-log-bin` (bool)


shlomi-noach:
`--sql-log-bin` would default `true`. When `false`, writes to _ghost_ table will not be logged to binlogs.

This may only work on:

- `--test-on-replica`
- `--migrate-on-replica`
- `--allow-on-master`

This cannot work on normal connect-to-replica-and-write-on-master.

When GTID is enabled such writes would cause errant transactions.

_Note_: the _changelog_ table also gets writes. In the event you're using GTID this still calls for a problem if you're migrating on a replica. See https://github.com/github/gh-ost/issues/146 for more discussion.

freediver9000:
Would it be acceptable  to update the master with a dummy transaction for
the single update on the changelog table?

Dave

shlomi-noach:
@freediver9000 would you know if this needs to be in sync with the write on replica, or how do I make MySQL understand the two are related?

shlomi-noach:
I believe the aforementioned write to the _changelog_ table can be skipped in the event of 

- running directly on master
- migrating on replica
- testing on replica

i.e. any case where `migrationContext.InspectorIsAlsoApplier()`. Will wrap my head around this next days.

shlomi-noach:
so much GTID pain. So many operations must prepend with `set sql_log_bin:=0`; this includes the very creation of changelog table, writes onto it, the cut-over operations and others.

PR #172, opened 2016-08-18
DML write wrapped in transaction

shlomi-noach:
Required for:

- https://github.com/github/gh-ost/issues/164
- https://github.com/github/gh-ost/issues/149
- https://github.com/github/gh-ost/issues/161

also solving the `golang` limitation: 

```
sql: converting Exec argument #2's type: uint64 values with high bit set are not supported
```
by converting `uint64` silently to strings when applying DML changes.
Related: https://github.com/github/gh-ost/issues/157

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -880,15 +880,14 @@ func (this *Applier) ApplyDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) error {
 		return nil
 	}()
 
-	if err == nil {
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Please see https://github.com/github/gh-ost/pull/172

shlomi-noach:
I'm running this through testing for the duration of the day. When this passed, I'm happy to merge in the TZ fix. 

cc @dveeden

shlomi-noach:
This passes tests well

Issue #62, opened 2016-06-10
Support hooks


shlomi-noach:
- On startup
- On end
- Before beginning rowcopy
- When prepared to cut-over
- Right after cut-over
- When begins throttling (spammy)
- When receiving interactive command

shlomi-noach:
Also: 
- on `--test-on-replica`, when about to `stop slave`

shlomi-noach:
- on "all validations ok". This is before issing the `count(*)` when `--exact-rowcount` is given.
  The `count(*)` may take some time.
  The "Before beginning rowcopy" (above) will notify just prior to starting copy, which is _after_ `count(*)`

shlomi-noach:
Documentation is on https://github.com/github/gh-ost/blob/master/doc/hooks.md

shlomi-noach:
Closed by #222

PR #174, opened 2016-08-18
Add --test-on-replica-manual-replication-control flag

### LINKED ISSUE #162: Support manual replication stop in --test-on-replica
Hello, I am trying to use gh-ost with Amazon's RDS.  Unfortunately, RDS doesn't give you `SUPER` privileges, so the `stop slave` statement doesn't work, when running it in `--test-on-replica` mode.

To work around this, I'm working off my own branch and I've [patched it to wait for replication to stop](https://github.com/Wattpad/gh-ost/commit/222fca486148bb8821860dba0a010370a2f25a42).

RDS does have its own command for stopping replication: ```call mysql.rds_stop_replication```  However, I wanted to avoid polluting the tool with RDS-specific statements, so I opted for just manually running the replication stop command externally.

Do you have any thoughts on whether/how this could be supported?  Perhaps a CLI argument like `--manual-replication-stop`?

I am happy to contribute a PR.

pbitty:
This will wait indefinitely for the replication status to change.
This allows us to run test schema changes in RDS without needing
custom RDS commands in gh-ost.

Closes #162

### DIFF
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -60,8 +60,8 @@ func main() {
 	flag.BoolVar(&migrationContext.SkipRenamedColumns, "skip-renamed-columns", false, "in case your `ALTER` statement renames columns, gh-ost will note that and offer its interpretation of the rename. By default gh-ost does not proceed to execute. This flag tells gh-ost to skip the renamed columns, i.e. to treat what gh-ost thinks are renamed columns as unrelated columns. NOTE: you may lose column data")
 
 	executeFlag := flag.Bool("execute", false, "actually execute the alter & migrate the table. Default is noop: do some tests and exit")
-	testOnReplicaWithManualReplicationControl := flag.Bool("test-on-replica-manual-replication-control", false, "Same as --test-on-replica, but waits for replication to be stopped, instead of stopping it automatically. (Useful in RDS.)")
 	flag.BoolVar(&migrationContext.TestOnReplica, "test-on-replica", false, "Have the migration run on a replica, not on the master. At the end of migration replication is stopped, and tables are swapped and immediately swap-revert. Replication remains stopped and you can compare the two tables for building trust")
+	flag.BoolVar(&migrationContext.TestOnReplicaSkipReplicaStop, "test-on-replica-skip-replica-stop", false, "When --test-on-replica is enabled, do not issue commands stop replication (requires --test-on-replica)")
 	flag.BoolVar(&migrationContext.MigrateOnReplica, "migrate-on-replica", false, "Have the migration run on a replica, not on the master. This will do the full migration on the replica including cut-over (as opposed to --test-on-replica)")
 
 	flag.BoolVar(&migrationContext.OkToDropTable, "ok-to-drop-table", false, "Shall the tool drop the old table at end of operation. DROPping tables can be a long locking operation, which is why I'm not doing it by default. I'm an online tool, yes?")
@@ -150,13 +150,13 @@ func main() {
 	if migrationContext.SwitchToRowBinlogFormat && migrationContext.AssumeRBR {
 		log.Fatalf("--switch-to-rbr and --assume-rbr are mutually exclusive")
 	}
-	if *testOnReplicaWithManualReplicationControl {
-		if migrationContext.TestOnReplica {
-			log.Fatalf("--test-on-replica-manual-replication-control and --test-on-replica are mutually exclusive")
+	if migrationContext.TestOnReplicaSkipReplicaStop {
+		if !migrationContext.TestOnReplica {
+			log.Fatalf("--test-on-replica-skip-replica-stop requires --test-on-replica to be enabled")
 		}
-		migrationContext.TestOnReplica = true
-		migrationContext.ManualReplicationControl = true
+		log.Warning("--test-on-replica-skip-replica-stop enabled. We will not stop replication before cut-over. Ensure you have a plugin that does this.")
 	}
+
 	switch *cutOver {
 	case "atomic", "default", "":
 		migrationContext.CutOverType = base.CutOverAtomic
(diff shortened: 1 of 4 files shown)


go/cmd/gh-ost/main.go
```diff
@@ -149,6 +150,13 @@ func main() {
 	if migrationContext.SwitchToRowBinlogFormat && migrationContext.AssumeRBR {
 		log.Fatalf("--switch-to-rbr and --assume-rbr are mutually exclusive")
 	}
+	if *testOnReplicaWithManualReplicationControl {
+		if migrationContext.TestOnReplica {
+			log.Fatalf("--test-on-replica-manual-replication-control and --test-on-replica are mutually exclusive")
+		}
+		migrationContext.TestOnReplica = true
+		migrationContext.ManualReplicationControl = true
+	}
```
pbitty: @shlomi-noach , in my opinion the code reads more complicated to me when there are two flags that have the same core functionality, with this one slight variation.

I think the semantics would be more clear if we still used the `--test-on-replica` flag and had a separate flag that just customized this one aspect of its behavior.  So if one wants to have manual replication control, one would use the args:

```
--test-on-replica --manual-replication-control
```

Without the extra flag, `--test-on-replica` would behave as it always has.

What do you think?

shlomi-noach: That was my original intention. The flag is good to be named `--test-on-replica-skip-stop-replica` or whatever, but it would be an _addon_ to `--test-on-replica`. Not only are the flags not mutually exclusive, the `--test-on-replica-skip-stop-replica` is only applicable when `--test-on-replica` exists.

go/logic/applier.go
```diff
@@ -566,14 +566,41 @@ func (this *Applier) StartSlaveSQLThread() error {
...
-		return err
+	if this.migrationContext.ManualReplicationControl {
+		for {
+			log.Info("Waiting for replication to stop...")
+			if this.isReplicationStopped() {
+				log.Info("Replication stopped.")
+				break
+			}
+			time.Sleep(5 * time.Second)
+		}
```
shlomi-noach: We discussed the use of plugins. In such case the plugin would be called synchronously and would stop replication.
The way I see it there is no need to wait for replication to stop: I'm assuming your call to amazon to stop replication is also synchronous, i.e. when the plugin returns, replication is stopped (or there is an error).
In either case, waiting is not the desired behavior.

I would suggest the PR could be made much smaller: merely introducing the flag; and if it is set, we skip the `StopSlave...` operation.

Later on I will add the plugin injection.

WDYT?

shlomi-noach:
See comments inline

go/cmd/gh-ost/main.go
pbitty: Ah, I understand now.  My apologies for the confusion.  I'll update the logic.

go/logic/applier.go
pbitty: I like that approach, it keeps things clean and simple in the tool.  I would probably block this PR until #62 is done, since simply skipping replication without the ability to stop it externally, would break the behavior of `--test-on-replica`.

I think we should also log a Warning or Info line explaining the above - eg. "You must use a plugin to stop replication."  The wording would depend on how #62 turns out.

Thoughts?

pbitty: As for the RDS calls to start/stop replication, they are indeed synchronous.  If there's ever a case of an async operation, the plugin could block while it waits for the desired state.

pbitty:
I've updated the flag and behavior we discussed.  It is now an add-on flag to `--test-on-replica` and has some warnings about the impact of using it without another means of stopping replication.

go/cmd/gh-ost/main.go
```diff
@@ -149,6 +150,13 @@ func main() {
 	if migrationContext.SwitchToRowBinlogFormat && migrationContext.AssumeRBR {
 		log.Fatalf("--switch-to-rbr and --assume-rbr are mutually exclusive")
 	}
+	if migrationContext.TestOnReplicaSkipReplicaStop {
+		if !migrationContext.TestOnReplica {
+			log.Fatalf("--test-on-replica-skip-replica-stop requires --test-on-replica to be enabled")
+		}
+		log.Warning("--test-on-replica-skip-replica-stop enabled. We will not stop replication before cut-over. Ensure you have a plugin that does this.")
```
pbitty: I've made mention of plugins (prematurely) in the warning because I don't think this feature will really work without them.

go/cmd/gh-ost/main.go
```diff
@@ -149,6 +150,13 @@ func main() {
 	if migrationContext.SwitchToRowBinlogFormat && migrationContext.AssumeRBR {
 		log.Fatalf("--switch-to-rbr and --assume-rbr are mutually exclusive")
 	}
+	if migrationContext.TestOnReplicaSkipReplicaStop {
+		if !migrationContext.TestOnReplica {
+			log.Fatalf("--test-on-replica-skip-replica-stop requires --test-on-replica to be enabled")
```
shlomi-noach: This will crash the migration at its very end -- after potentially hours of working. Instead, we can verify this in `main.go` -- at the very beginning of the migration.
Otherwise looking good!

pbitty: As far as I can tell, the above is happening before the migration.  This is in already in `main.go`, and the migration starts on [line 198](https://github.com/Wattpad/gh-ost/blob/2e43718ef38032b9589c32e59c2e9174d9fe44f7/go/cmd/gh-ost/main.go#L198).  Unless I am missing something...

shlomi-noach: My mistake!

go/logic/applier.go
```diff
@@ -568,12 +568,17 @@ func (this *Applier) StartSlaveSQLThread() error {
 
 // StopReplication is used by `--test-on-replica` and stops replication.
 func (this *Applier) StopReplication() error {
-	if err := this.StopSlaveIOThread(); err != nil {
-		return err
-	}
-	if err := this.StopSlaveSQLThread(); err != nil {
-		return err
+	if this.migrationContext.TestOnReplicaSkipReplicaStop {
```
shlomi-noach: Let me know if I'm getting too bossy.
I'd rather do the `if this.migrationContext.TestOnReplicaSkipReplicaStop` externally to this function. So I would rather have the logic in `migrator.go` instead.

pbitty:
Not at all.  I'll take a look first thing tomorrow.

pbitty:
Actually, that was simple, just did it now.  :)

shlomi-noach:
@pbitty this is what it would look like once https://github.com/github/gh-ost/pull/190 is merged: https://github.com/github/gh-ost/pull/190/files#diff-e2990679ccfa83d6fc15414958333bcbR523

PR #175, opened 2016-08-19
setting time_zone on DML apply

shlomi-noach:
Storyline: #161 

This addresses the `time_zone` setting by applying a session`time_zone` for each and every DML applied, within same transaction

cc @dveeden 

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 #
 #
 
-RELEASE_VERSION="1.0.10"
+RELEASE_VERSION="1.0.11"
 
 function build {
     osname=$1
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Obsoletes https://github.com/github/gh-ost/pull/165

shlomi-noach:
@dveeden are you able to test this branch on your side?

shlomi-noach:
Insofar this passes tests on my side; however the problem never manifested on my side in the first place. I'll only merge this on Monday, pending approval by @dveeden

shlomi-noach:
Testing passes. Am also working on local tests to verify this particular error, in separate PR.

PR #182, opened 2016-08-22
row_event converts timestamps to UTC

shlomi-noach:
Continued discussion on https://github.com/github/gh-ost/issues/161

This explicitly sets UTC for timestamps read by the `go-mysql` library.
Unfortunately when returning timestamp, it is returned as a `string`, just so that it is able to support `0000-00-00 00:00:00`.
But then it uses the local machine's timezone, converting a real time into string, losing context. Because this is a string, it is not converted back to number when applied as DML event.
This is an attempt to fix that, with local tests already proving well.

cc @dveeden 

### DIFF
--- a/vendor/github.com/siddontang/go-mysql/replication/row_event.go
+++ b/vendor/github.com/siddontang/go-mysql/replication/row_event.go
@@ -609,7 +609,7 @@ func decodeTimestamp2(data []byte, dec uint16) (string, int, error) {
 		return "0000-00-00 00:00:00", n, nil
 	}
 
-	t := time.Unix(sec, usec*1000)
+	t := time.Unix(sec, usec*1000).UTC()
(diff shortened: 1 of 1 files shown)


dveeden:
Ran a test and it looks ok. :+1:

siddontang:
@shlomi-noach 

Do I need to fix it in go-mysql?

shlomi-noach:
@siddontang I'm not sure what the correct behavior is in general.

The thing is you have to convert the `TIMESTAMP` to `string` so as to support `0000-00-00 00:00:00`, but then you lose the context of your time zone.

My take was to turn my app to `UTC` completely, and let MySQL figure out the timezone differences and do the math.

I think it would be at least nice to have this configurable: do I want to read timestamps in local timezone or in UTC.

siddontang:
Got it, thanks @shlomi-noach

Issue #178, opened 2016-08-19
Error with renamed column 


ggunson:
Running a migration that changed the name of a column, as well as added an index on the (newly named) column, it eventually died with this message:

```
2016-08-19 14:37:46 ERROR shared columns is not a subset of table columns in BuildDMLUpdateQuery
2016-08-19 14:37:46 FATAL shared columns is not a subset of table columns in BuildDMLUpdateQuery
```

This was after a 40 minute long `select count(*)`, which looks like it could be hitting a bug with MySQL 5.7 (https://bugs.mysql.com/bug.php?id=80580), so I'm guessing it got past the initial schema checks.

Command was:

```
gh-ost-wrapper-script -c mysql1 -t tablename -d 'CHANGE old_column new_column int(11) DEFAULT NULL,ADD INDEX index_1 (new_column),ADD INDEX index_2 (column_whocares, column_fine),DROP INDEX shitty_index' -x -- --approve-renamed-columns 
```

cc @github/database-infrastructure this is blocking a migration

shlomi-noach:
I am able to reproduce this. This error happens for any `rename` and `update`.

shlomi-noach:
> This was after a 40 minute long select count(*), which looks like it could be hitting a bug with MySQL 5.7 (https://bugs.mysql.com/bug.php?id=80580), so I'm guessing it got past the initial schema checks.

The bug happens while handling an `UPDATE` event in the binary log. This is the reason why `gh-ost` only crashed 40 minutes into the execution: everything was sound and fine in terms of schema change, and `count(*)` took a long time; the problem only manifested while applying changes read from the binary log.

shlomi-noach:
Fix is under way.

shlomi-noach:
Fix sent to testing

shlomi-noach:
`UPDATE` is fixed. Still have `INSERT` issue.

shlomi-noach:
Sent to testing

shlomi-noach:
Testing passes. `INSERT` fix merged. Will soon release.

shlomi-noach:
`1.0.13` fixes this

PR #183, opened 2016-08-22
fix INSERT DML handling on renamed column

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/178

Fixing the `shared columns is not a subset of table columns in BuildDMLInsertQuery`, found on column rename and while applying `INSERT` DML

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 #
 #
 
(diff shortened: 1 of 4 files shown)


shlomi-noach:
Sent to testing

PR #185, opened 2016-08-23
WIP: Local tests

shlomi-noach:
Work in progress: local tests

This is in preparation for a CI job that would run explicit tests on a local replication setup.
In these tests we expect a master-replica setup. 

A test:

- creates a table
- creates an event that populates the table and runs for 60 seconds
- issues `gh-ost --test-on-replica` that runs for at least `5` seconds
- expects identical `md5sum` on original and ghost tables

All tests share the same `gh-ost` invocation logic, being able to override arguments and column names.


shlomi-noach:
Merging for now and will iterate later. Want those initial tests in.

PR #191, opened 2016-08-23
DML write sets sql_mode to STRICT ALL TABLES

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/164

Each DML write forces session `sql_mode` to `STRICT_ALL_TABLES`.

This serves as a safety net in case the server's `sql_mode` is not strict. `gh-ost`'s mode _should_ be strict because it just copies values around, so there should not be any case where data is truncated.

On interesting potential case is if the user reduces `BIGINT` to `INT`, or `VARCHAR(128)` to `VARCHAR(64)`, for example, and some existing values do not fit in reduced column size.
This PR will assist the user in avoiding shooting herself in the foot.

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -870,7 +870,10 @@ func (this *Applier) ApplyDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) error {
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Sent to testing

shlomi-noach:
Testing goes well. Merging.

Issue #180, opened 2016-08-21
Data copy continues after the rename completed


borizio:
While running the gh-ost  v1.0.10 the following output was observed:

```
Copy: 349054/326425 106.9%; Applied: 450; Backlog: 0/100; Time: 1m6s(total), 59s(copy); streamer: mysql-bin.023347:187676098; ETA: Due
2016-08-21 16:31:42 INFO Setting RENAME timeout as 3 seconds
2016-08-21 16:31:42 INFO Session renaming tables is 77018987
2016-08-21 16:31:42 INFO Issuing and expecting this to block: rename /* gh-ost */ table `gsstats_sh6`.`gs_raw_comments` to `gsstats_sh6`.`_gs_raw_comments_del`, `gsstats_sh6`.`_gs_raw_comments_gho` to `gsstats_sh6`.`gs_raw_comments`
2016-08-21 16:31:42 INFO Found atomic RENAME to be blocking, as expected. Double checking the lock is still in place (though I don't strictly have to)
2016-08-21 16:31:42 INFO Checking session lock: gh-ost.77019071.lock
2016-08-21 16:31:42 INFO Connection holding lock on original table still exists
2016-08-21 16:31:42 INFO Will now proceed to drop magic table and unlock tables
2016-08-21 16:31:42 INFO Dropping magic cut-over table
2016-08-21 16:31:42 INFO Releasing lock from `gsstats_sh6`.`gs_raw_comments`, `gsstats_sh6`.`_gs_raw_comments_del`
2016-08-21 16:31:42 INFO Tables unlocked
2016-08-21 16:31:42 INFO Tables renamed
2016-08-21 16:31:42 INFO Lock & rename duration: 1.44793849s. During this time, queries on `gs_raw_comments` were blocked
2016-08-21 16:31:42 INFO Looking for magic cut-over table
2016-08-21 16:31:42 INFO Droppping table `gsstats_sh6`.`_gs_raw_comments_ghc`
2016-08-21 16:31:42 INFO Table dropped
2016-08-21 16:31:42 INFO Droppping table `gsstats_sh6`.`_gs_raw_comments_del`
2016-08-21 16:31:43 ERROR Error 1146: Table 'gsstats_sh6._gs_raw_comments_gho' doesn't exist; query=
                        replace /* gh-ost `gsstats_sh6`.`_gs_raw_comments_gho` */ into
                                `gsstats_sh6`.`_gs_raw_comments_gho`
[...]

PR #192, opened 2016-08-23
avoiding writing rows when rowcopy complete

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/180

This should address the problem of rows still attempted to copy after completion flag is set.
This should solve the thread synchronization problem because rowcopy executes sequentially.

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 #
 #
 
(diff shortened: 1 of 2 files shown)


shlomi-noach:
Sent to testing

shlomi-noach:
testing is good

Issue #194, opened 2016-08-23
FR: if --exact-rowcount=true, don't wait for count to return before starting


ggunson:
`SELECT COUNT(*) FROM table` can take a long time with large tables, and is slower on MySQL 5.7 than 5.6, according to this bug report: https://bugs.mysql.com/bug.php?id=80580

Using the `--exact-rowcount` flag, gh-ost gets a count of the table before it starts copying rows, so that the progress and estimated completion times are more accurate than using table statistics for the calculations.

If the flag is used on a big migration, the migration can be delayed several minutes (we saw upwards of 40 minutes in practice using a 5.7 replica) waiting for the count, and the row count is now 40 minutes old and is no longer accurate on an insert-heavy table. 

Is it possible to start the row copy prior to getting the count and check on the result later as it would with the various panic/postpone flag files? It can take a while for gh-ost to display an estimated completion time already, even after it has the count. 

/cc @github/database-infrastructure

shlomi-noach:
> If the flag is used on a big migration, the migration can be delayed several minutes (we saw upwards of 40 minutes in practice using a 5.7 replica) waiting for the count, and the row count is now 40 minutes old and is no longer accurate on an insert-heavy table. 

That is true, and the number is never really absolutely correct, but experience shows that it is nonetheless very close to the real value; on our largest migrations I've see it off by less than `1%`. So I suggest in terms of accuracy we've achieved a good one.

I think the big deal here is "are we willing to pay an extra 30 minutes for the `count(*)` to complete before migration starts".

Running the `count(*)` in parallel to running the row-copy and updating the estimate later is entirely possible.

This will have the:

-  positive effect of reducing runtime and getting some stats early on the migration
- confusing effect of changing ETA some 30 minutes after migration started

We can make this optional and the user can choose desired behavior.

shlomi-noach:
Waiting the first 30 minutes of a migration by just waiting for `count(*)` to complete is also something that's not human friendly -- another reason for allowing `count(*)` to run concurrently with migration.

dveeden:
On an insert heavy table running the `count(*)` for 40 minutes might also grow the undo logs in ibdata1. Maybe changing the transaction isolation is useful.

Does it adjust the count when gh-ost sees `INSERT`s or `DELETE`s? If so, are these two (count and adjustments) properly aligned? Maybe it should start a transaction (and enforce InnoDB to take part in it) and then get the binlog position and then start the count and the watching-of-binlogs?

[...]

PR #196, opened 2016-08-24
concurrent row-count

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/194

This introduces concurrent `select count(*)` even as row-copy begins.
WIP


shlomi-noach:
This works

Issue #184, opened 2016-08-23
 NOT  work   on  mysql  5.5


nilly:
test  on mysql 5.5.9  

```
gh-ost \
--max-load=Threads_running=25 \
--critical-load=Threads_running=1000 \
--chunk-size=1000 \
--throttle-control-replicas="127.0.0.1,127.0.0.1" \
--max-lag-millis=1500 \
--user="root" \
--password="" \
--host=127.0.0.1 \
--allow-on-master \
--database="test" \
--table="lsl" \
--verbose \
--alter="engine=innodb" \
--switch-to-rbr \
--allow-master-master \
**--allow-nullable-unique-key \**
--cut-over=default \
--exact-rowcount \
--default-retries=120 \
--panic-flag-file=/tmp/ghost.panic.flag \
--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag \


2016-08-23 11:01:02 INFO Chosen shared unique key is idx_id1
2016-08-23 11:01:02 WARNING Chosen key (idx_id1: [id1]; has nullable: true) has nullable columns. You have supplied with --allow-nullable-unique-key and so this migration proceeds. As long as there aren't NULL values in this key's column, migration should be fine. NULL values will corrupt migration's data
2016-08-23 11:01:02 FATAL binlog_row_image is '' and chosen key is idx_id1: [id1]; has nullable: true, which is not the primary key. This operation cannot proceed. You may `set global binlog_row_image='full'` and try again
```

BUT   mysql 5.5.9  has  no this  global variable  **binlog_row_image** 

and  I  have already   use --allow-nullable-unique-key

shlomi-noach:
True, `binlog_row_image` was only introduced in `5.6.2`. Will fix.

shlomi-noach:
Seems like I did foresee this: https://github.com/github/gh-ost/blob/master/go/logic/inspect.go#L300
checking what went wrong.

shlomi-noach:
Fixed by https://github.com/github/gh-ost/pull/200

shlomi-noach:
Will likely publish a release tomorrow that will include this fix.

shlomi-noach:
released in https://github.com/github/gh-ost/releases/tag/v1.0.14

PR #200, opened 2016-08-25
fixed 5.5 binlog_row_image default value

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/184

While the code did make sure not to fail on `5.5`, it didn't set a reasonable default, namely `FULL`.
It does so now.


shlomi-noach:
woohoo! PR number 200!

PR #217, opened 2016-09-01
Modified definition of `critical-load`

ggunson:
This PR is just to update the help docs to clarify `--critical-load`, because it didn't give the format for the option like it did for `--max-load`, and people who aren't used to running `pt-online-schema-change` won't necessarily recognize that they're related.

I mostly copied the `--max-load` definition and added some embedded list items to break lines, because I couldn't figure out from markdown how to do newlines within list items. 

/cc https://github.com/github/shell/pull/3842 as the use case for the option
/cc @github/database-infrastructure for feedback on docs change


shlomi-noach:
FYI multi-line within bullet is as follows:

```
- main bullet
- main bullet 2
  2nd line in main bullet 2
- main bullet 3
  - sub bullet 3
    2nd line in sub bullet 3
- main bullet 4
```

- main bullet
- main bullet 2
  2nd line in main bullet 2
- main bullet 3
  - sub bullet 3
    2nd line in sub bullet 3
- main bullet 4

shlomi-noach:
I thought I knew how to do multi-line bullet but seems I was wrong.

PR #220, opened 2016-09-01
Fail operation if onStopReplication hook fails

pbitty:
Without this check, the process will continue even if the hook call fails.  I figure we'd want to abort and try again if the hook fails.


shlomi-noach:
Thank you -- oh actually we want to _retry_ this. Let me look into it today.

shlomi-noach:
All right. Meanwhile failing on error is good to go. I think the issue of retrying hooks is on a grander scale and non trivial. Merging.

Issue #212, opened 2016-08-30
Specify master directly when connecting to replica


dataindataout:
I see when using the "connect to replica, migrate on master" method, the master is detected via the function here https://github.com/github/gh-ost/blob/f06b469227208000d70739233fc222b1b4998f04/go/mysql/utils.go#L77.

Is there a way (that I've missed) to specify the master directly via a flag? If not, would making this configurable cause any architecture issues or predictable bugs?

Thank you!

shlomi-noach:
@dataindataout there is no way to do that right now. Adding this would not cause a significant architectural change. 

Can you elaborate on the need?

The risk I see is someone wrongly specifying a master which is not on same topology (though `gh-ost` can presumably check it out)

shlomi-noach:
Or, another more likely risk is the user specifying a non-master server.

dataindataout:
Thank you for the reply! The use case would be non-native replication (e.g., Tungsten replicator), so a show slave status doesn't provide information. By the way, this is working functionally with the "Connect to master" method, of course.

shlomi-noach:
ahh... didn't cross my mind. Then, yes, it's possible. I might do this in the next week or two.

dataindataout:
Thank you for taking a look at it!

shlomi-noach:
@dataindataout out of curiosity, is your topology that uses tungsten strictly RBR?

dataindataout:
It's actually SBR (and can't be changed), so we've found that we do need to run it from a replica.

shlomi-noach:
@dataindataout I don't have a tungsten setup on me, could you please answer the following: if I were to look at one of your replica servers, replicating via tungsten, is there any indication at all in that server that it acts as a replica? I assume `show slave status` shows empty result. What would be a way for `gh-ost` (or any client) to say "oh yeah, this is a tungsten-based replica" ?

dataindataout:
Yes, show slave status is blank:
mysql> show slave status\G
Empty set (0.00 sec)

From mysql, you can see the tungsten connections, but of course there's no guarantee someone has named their user "tungsten":

mysql> show processlist\G
*************************** 1. row ***************************
           Id: 48
         User: tungsten
         Host: tungstenslave:52707
           db: tungsten_cookbook
      Command: Sleep
         Time: 2654
        State:
         Info: NULL
    Rows_sent: 0
Rows_examined: 0
    Rows_read: 0

[...]

PR #221, opened 2016-09-02
Add -override-applier-host for use with -allow-master-master

twotwotwo:
This is for configurations where writes are meant to go to one master, but gh-ost can't automatically determine which.

Related discussion on https://github.com/github/gh-ost/issues/212 -- it looks like there are a couple use cases where this comes in handy, and we've done a migration using the flag on our cluster.

I could rename it something like -active-master since that's closer to how the user thinks of it. It could be required when master-master is allowed and there are at least three nodes. I should get an idea what you think before going down any of those paths.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -109,6 +109,7 @@ type MigrationContext struct {
 	CutOverType                  CutOver
 
 	Hostname                               string
+	OverrideApplierHostname                string
 	TableEngine                            string
 	RowsEstimate                           int64
 	RowsDeltaEstimate                      int64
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Thank you for this PR: it makes perfect sense.

I'm going to merge it, but will then iterate and override as follows:

- rename argument to `--assume-master-host`
- decouple from `--allow-master-master` so as to support `tungsten`, as per #212

shlomi-noach:
Reiterated via https://github.com/github/gh-ost/pull/223

PR #227, opened 2016-09-06
added charset support & tests

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/226

This adds tests for `utf8` and `latin1` text columns with "special" characters

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -2,7 +2,7 @@
 #
 #
 
-RELEASE_VERSION="1.0.17"
+RELEASE_VERSION="1.0.18"
 
 function build {
     osname=$1
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -301,7 +301,7 @@ func (this *Applier) ExecuteThrottleQuery() (int64, error) {
 // ReadMigrationMinValues returns the minimum values to be iterated on rowcopy
 func (this *Applier) ReadMigrationMinValues(uniqueKey *sql.UniqueKey) error {
 	log.Debugf("Reading migration range according to key: %s", uniqueKey.Name)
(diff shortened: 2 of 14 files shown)


SuperQ:
Don't forget to test utf8mb4 with some emoji chars. :smile: :hankey: :pizza:

shlomi-noach:
true, true. Bummed by the `latin1` problem at the moment.

SuperQ:
Yup, char encoding is a long sad MySQL story :disappointed: :sob: :crying_cat_face:

shlomi-noach:
`utf8mb4` is good (rather, made it good easily, need to merge).
`latin1` still a pain.

shlomi-noach:
The problem goes away when the MySQL connection is configured to use `charset=latin1`. But this is not a valid solution because we wish to support tables that have multiple charset columns. Maybe as a quick workaround...

shlomi-noach:
Character set recognition and manipulation

- Identifying textual characters sets; converting into specific type when applying dml events
- Refactored `ColumnsList`: introducing `Column` type
- Refactored `unsigned` handling, as part of `Column`
- `Column` type supports `convertArg()`: converting value of argument according to column data type
- DB URI attempts `utf8mb4,utf8,latin1` charsets in that order (first one to be recognized wins)
- Local tests filter by pattern
- Local tests append table schema on failure
- Local tests do not have postpone flag file
- Added character set local tests: `utf8`, `utf8mb4`, `latin1`

SuperQ:
:+1: :cake: :ice_cream:

shlomi-noach:
all tests pass. merging.

PR #231, opened 2016-09-12
Named cut over

shlomi-noach:
This PR introduces the possibility to require `gh-ost` to only respond to `cut-over` given the name of the migrated table:

- when `--force-named-cut-over` is given, the interactive `cut-over|postpone` command _must_ include name of table, as so:
  `echo unpostpone=thetable | socat - /tmp/gh-ost.sock`
- when the flag is not given, the user may issue `cut-over` without table name
- or may choose to issue `cut-over` with table name

Whenever table name is given (ether forced or voluntary), `gh-ost` compares it to migrated table name. It will only cut-over if they're identical, and will just ignore the command otherwise.


shlomi-noach:
There's also been some refactoring here, extracting code from `migrator.go` and into `server.go`

shlomi-noach:
passes tests. merging

PR #232, opened 2016-09-12
rowcount progress at 100% when row-copy completes

shlomi-noach:
When row copy is complete, there is no point in heuristically updating the number of rows. In fact, the number of copied rows is the _de facto_ number of rows, and so this number is made fixed.


jonahberquist:
Hah, nice 👍

Issue #239, opened 2016-09-20
Fail when trying to change column order


pbitty:
Currently changing column order is not supported by gh-ost.

I'm wondering if it should detect this scenario and fail with an error message, or at least have a note in the limitations doc.

shlomi-noach:
What is the output?

pbitty:
gh-ost output is normal, it runs to completion with no errors, but the column values get swapped.

I modified `localtests/rename` to have this alter statement:
```
--alter="change column c2 c3 int not null first" --approve-renamed-columns
```

Here's the [full output of the test](https://gist.github.com/pbitty/c084a6776a06ebf39b5a95a8c9c07f2d#file-output).

shlomi-noach:
The above test both renames _and_ reorders -- is that what the original Issue comment is about?

`gh-ost` will not support rename-and-reorder, and I'll add this to `limitations` docs. However I do believe you opened the issue for reorder only.

pbitty:
Oops, you are right, I opened it only for the reorder functionality but reused the rename test to quickly test it out.  I'll try to avoid that source of confusion in the future.

Agreed on the above, the main point of the issue was to discuss either documenting it or detecting it and exiting.

pbitty:
To make things more clear, I re-tested with only reordering, no rename:
```
--alter="modify column c2 int not null after id"
```

As expected, gh-ost runs to completion but the column values get swapped between `c1` and `c2`.

shlomi-noach:
s/As expected/Not as expected/ ?

pbitty:
I meant that this is the behavior I currently expect from gh-ost since it doesn't support reordering.   It will complete the migration and exit with a 0 exit code.

The desired behavior perhaps would be for it to exit saying it does not support column reordering.

But a note in the docs is probably enough.

shlomi-noach:
I do believe normal column reorder works well. Please take a look at: https://github.com/github/gh-ost/pull/240

The reason your test failed is, I suspect, that the test issued an implicit `select *` on orig & ghost tables, where `*` extrapolates columns in different order in your case.

I've refined the test to support explicit named columns, and this shows migration success. Can you please review?

pbitty:
My bad... :facepalm:  

It looks like even rename-and-reorder will work.  I had the wrong mental model of what was happening in `BuildDMLInsertQuery`.  Will review #240.

pbitty:
I am closing this issue, as it was resolved by #240.  Thanks @shlomi-noach .

PR #240, opened 2016-09-20
added tests for column-reorder

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/239

These tests test a column reorder DDL:

```
--alter="change column c2 c2 int not null after id"
```

and confirms that `gh-ost` supports column reorder.


pbitty:
Do we want to add a test for 'rename-and-reorder' as well?

Either way, LGTM.

shlomi-noach:
rename & reorder added -- actually works! Hehe, was about to document that it is not supported. Still am willing to document that, if only to deter people from doing such a crazy thing.

Issue #247, opened 2016-09-27
Incorrect values when converting DATETIME to TIMESTAMP on non-UTC server


mignaulo:
We are seeing incorrect date values when converting a DATETIME column to TIMESTAMP on a server which is not in UTC. The dates are off by the number of hours between PDT (the server timezone) and UTC. 

To explain why we are doing this, we are attempting to repair the bad decisions we made when we didn't have our MySQL servers in UTC. We are converting our DATETIME columns to TIMESTAMP which will allow us to easily change the database time zone. 

Rows that were copied in batches are correct. I assume this is performed using the default server timezone, the tool does not try to override that.

Rows that are inserted from the binlog are incorrect. I assume the binlog is inherently in the timezone of the server (in this case PDT), while inserts are specifically made with the time_zone set to UTC (https://github.com/github/gh-ost/blob/master/go/logic/applier.go#L875). 
We therefore end up with some rows 7 hours behind what we expect. 

I will need to do further testing, but binlog update statements also do not seem to apply. I am not certain about this, but I assume the UPDATE statement it generates might include every old column values in the WHERE clause? If that is the case, the timestamp in the table would not match, since changing the time_zone changes the table date to UTC while the date from the binlog is still in PDT.

I believe this is related to #161 or #175, and fixing that particular issue probably caused the one I am currently seeing. 

I'm unsure what would be the best way to fix this problem. Removing the session time_zone would certainly fix my problem, but would bring back the problems from the issues mentioned above. Being able to override the timezone this is set to would also help but it's not the cleanest of solutions.

Happy to provide additional information and run tests. 

I appreciate your help here, thanks!

gh-ost: 1.0.20

shlomi-noach:
https://github.com/github/gh-ost/pull/248 adds a test for the above use case. Right now, that test fails.

shlomi-noach:
> To explain why we are doing this, we are attempting to repair the bad decisions we made when we didn't have our MySQL servers in UTC. We are converting our DATETIME columns to TIMESTAMP which will allow us to easily change the database time zone. 

Unrelated to the bug, this is an interesting note. I'm in fact more inclined to get rid of `timestamp` as much as possible. One option is to use `datetime`, another is to store time as `int` in `unix_timestamp` format (ie seconds since leap time).

Reason is the insane conversions when connecting with utc/non-utc settings; the outputting of time as string, which is then difficult to manipulate.

Food for thought.

shlomi-noach:
[...]

PR #248, opened 2016-09-28
adding datetime->timestamp conversion test

shlomi-noach:

testing conversion from `datetime` to `timestamp` value.


shlomi-noach:
I've identified that the problem was not on the binlog-event-apply part, but actually on the row-copy part.

The row-copy statement now works with `time_zone='+00:00'` and this seems to fix the tests, consistently.

shlomi-noach:
Closed in favor of the more encompassing https://github.com/github/gh-ost/pull/269

Issue #256, opened 2016-10-03
panic on master-master setup when on-failure hook exists


jbodah:
Had a master-master setup and a simple `echo hello` hook (and a `on-startup` hook which worked fine). `gh-ost` errored out on the master-master setup and then hit a panic when trying to run the failure hook. Removing the hook allowed `gh-ost` to gracefully fail and tell me that the master-master setup was the problem.

Version 1.0.20, Ubuntu 14.04, installed from release tarball

```
#! /usr/bin/env sh
# hooks/gh-ost-on-failure
echo hello
```

Running gh-ost with the following:

```
gh-ost --database=my_database --conf=my_conf.cnf --table=my_table --alter='ADD COLUMN some_col int' --hooks-path=hooks --hooks-hint=my_name --verbose --switch-to-rbr --test-on-replica --execute --debug
```

Caused a panic:

```
2016-10-03 14:11:22 INFO starting gh-ost 1.0.20
2016-10-03 14:11:22 INFO Migrating `my_database`.`my_table`
2016-10-03 14:11:22 INFO executing gh-ost-on-startup hook: hooks/gh-ost-on-startup--tell-slack
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   310  100     2  100   308      9   1418 --:--:-- --:--:-- --:--:--  1419
ok
2016-10-03 14:11:22 INFO connection validated on xxx.xxx.xxx.xxx:3306
2016-10-03 14:11:22 INFO User has SUPER, REPLICATION SLAVE privileges, and has ALL privileges on `my_database`.*
2016-10-03 14:11:22 INFO binary logs validated on xxx.xxx.xxx.xxx:3306
2016-10-03 14:11:22 INFO Restarting replication on xxx.xxx.xxx.xxx:3306 to make sure binlog settings apply to replication thread
2016-10-03 14:11:22 DEBUG Replication restarted
2016-10-03 14:11:22 INFO Table found. Engine=InnoDB
2016-10-03 14:11:22 DEBUG Estimated number of rows via STATUS: 177523871
2016-10-03 14:11:23 DEBUG Validated no foreign keys exist on table
2016-10-03 14:11:23 DEBUG Validated no triggers exist on table
2016-10-03 14:11:23 INFO Estimated number of rows via EXPLAIN: 191585167
[...]

PR #260, opened 2016-10-04
safe access to applier/inspector hostnames for hooks

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/256

this PR provides safe access to applier/inspector hostnames, such that `nil` references are silenced.


shlomi-noach:
@jbodah are you capable of building a binary from this branch? If not, I'll provide binaries later.

Issue #262, opened 2016-10-05
Can't remove foreign keys with alter statement


Frank-Leemkuil:
> This is the place to report a bug, ask a question, suggest an enhancement.

> This is the place to make a discussion before creating a PR.

> Please label your Issue

> Please understand if this Issue is not addressed immediately or in a timeframe you were expecting.

> Thank you!

Frank-Leemkuil:
When we want to remove two foreign keys from the table we get an error that we can't remove it.
This is needed to use the gh-ost tool.

ggunson:
@Frank-Leemkuil You're just trying to run a straight "ALTER TABLE" statement to remove the foreign keys, so that then you can use gh-ost for future migrations? Or are you trying to use gh-ost to remove the foreign keys?

What is the error?

Frank-Leemkuil:
We need to remove the foreign keys with gh-ost. Just spoken on percona europe 2016 with your colleges regarding this issue. They told is an bug/feature.

shlomi-noach:
The request is to be able to get rid of foreign keys via `gh-ost` migration. Today `gh-ost` says: "oh you have foreign keys therefor I bail out", but these people actually want to get rid of their FK.

This is basically a `--yes-ok-to-completely-ignore-foreign-keys`

tomkrouper:
@Frank-Leemkuil,

During our discussion you showed the output. Could you please add that here? Possibly with a sanitized output of `show create table ...\G` If you aren't aware, it's best to put the output in triple back ticks for a code block.

Frank-Leemkuil:
The create table of the table:

       Table: alarm
Create Table: CREATE TABLE `alarm` (
  `Id` int(11) unsigned NOT NULL AUTO_INCREMENT,
  `Guid` char(36) NOT NULL,
  `AssetId` int(11) unsigned NOT NULL,
  `EventId` int(11) unsigned NOT NULL,
  `CreatedOn` datetime NOT NULL,
  `AcknowledgedOn` datetime DEFAULT NULL,
  `ProcessedOn` datetime DEFAULT NULL,
  PRIMARY KEY (`Id`),
  KEY `alarm_ibfk_2` (`EventId`),
  KEY `Guid` (`Guid`) USING BTREE,
  KEY `AssetIdCreatedOn` (`AssetId`,`CreatedOn`) USING BTREE,
[...]

PR #264, opened 2016-10-07
Discard foreign keys

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/262

This PR supports dropping of foreign keys from a table. There's nothing special about what `gh-ost` does to make that happen: it just doesn't do anything to _create_ those foreign keys on the _ghost_ table.


ggunson:
Can the docs say something as well about what it does to foreign keys that reference the migrated table? Even if that is repeating itself, that would be a good place to repeat it.

shlomi-noach:
@ggunson thanks for your comment, and I realize we need to refine this and only allow on tables with child-side FK. I don't wish to support any kind of parent-size FK handling at this time; it seems too dangerous to me (`_del` table dropped with `on delete cascade` --> whoops, all rows deleted from child table)

shlomi-noach:
@ggunson https://github.com/github/gh-ost/pull/264/commits/a4d566e71c5bdfc6cb7dddf8f35e27581844fe27 takes care of parent-side: not supported. Only child-side foreign keys can be discarded.

For people who are looking to compleltely eliminate foreign keys on their database, you should begin with child-most tables and move upwards.

I don't have an immediate solution to a self-referencing FK table.

ggunson:
> thanks for your comment, and I realize we need to refine this and only allow on tables with child-side FK.

@shlomi-noach I asked if the **docs** for this could say something about foreign keys referencing the table. So it was a suggestion that you add that to the docs you edited for this PR (e.g., the new flag description), that this doesn't apply to tables who are referenced in FKs from other tables.

shlomi-noach:
@ggunson I was sure I did 😮 but then I realize I wrote this on another couple places. Will add.

Anyway, my original comment acknowledged that regardless of the docs, the _code_ was wrong, and that is fixed. Proper error messages will indicate the limitation where someone tried to drop a parent-side foreign key. Docs were supposed to follow suit...

PR #265, opened 2016-10-10
WIP: configurable time-zone, native time parsing

shlomi-noach:

This PR takes a different approach to time-zone handling, and is Work In Progress

- adding `--time-zone` flag so that user can specify e.g. `--time-zone='-07:00'`
- natively parsing/returning `Time` values on `row_event`, i.e. when parsing binary logs. Only returning a `string` when the time is `0000-00-00 00:00:00`.

Working to solve all cosmic timezone problems.


shlomi-noach:
Closed in favor of #269

Issue #246, opened 2016-09-26
Allow --critical-load to set throttle flag


tomkrouper:
I was thinking about `gh-ost` and how it handles `--critical-load`. The issue is, that as soon as critical load is hit, the process drops out without chance of recovery. Any data copied is unusable. Issue #205 is being worked on to address this, but I'm curious if there is an easier way.

Because `gh-ost` has such a light amount of activity happening when the process is set to `throttle`d, I'm proposing we should just set `throttle` to on when critical load is hit. This allows for a DBA to investigate what happened before the schema change continues. This also prevents the loss of any activity that the migration has already completed. Once the problem has been investigated, the schema change can be `un-throttle`d and `gh-ost` can complete its work.

I'd love to get people's thoughts on this.

shlomi-noach:
This makes a lot of sense to me, and we can make that configurable via some `--throttle-and-wait-for-human-ack-on-critical-load`

jonahberquist:
I worry a bit that if we only throttle on critical load, we might miss some weird corner case where gh-ost itself even in throttling is causing load. It seems unlikely, but I want to be able to just let gh-ost run and forget about it, knowing that it will die if it needs to.

What if we went with an optional second chance on `--critical-load`?

The behavior could be:
 * first hit of critical-load throttles the migration until human intervention
 * if load is still at critical-threshold 15 seconds later, gh-ost process dies

shlomi-noach:
@jonahberquist I'm goind to proceed with your suggestion. An optional variable would be `--critical-load-interval-millis`, an `int`, defaults `0`. Values of this flag would imply:

- `0` (default): immediate bailout on meeting `critical-load`
- positive integer (milli seconds): don't bail out; make 2nd check after given number of millis. If `critical-load` is still met, bail out. If not, forget it ever happened.
  As per throttling, that would be taken care of by `max-load`. `critical-load` would not handle throttling.

Basically this is somewhat like a Nagios "number of checks", but reduced to _potentially one extra check_

jonahberquist:
That sounds good to me. 👍

tomkrouper:
This sounds like a decent compromise on what I was looking for.

shlomi-noach:
Handled by #266

PR #266, opened 2016-10-10
support for --critical-load-interval-millis

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/246

introducing `--critical-load-interval-millis`

This optional flag gives `--critical-load` a second chance. When configured to positive value, meeting with critical-load spawns a timer. When this timer expires a second check for critical-load is made; if still met, gh-ost bails out.
By default the interval is zero, in which case gh-ost bails out immediately on meeting critical load.

### DIFF
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -43,6 +43,16 @@ password=123456
 
 See `exact-rowcount`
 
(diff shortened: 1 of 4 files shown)


jonahberquist:
Nice 👍

PR #269, opened 2016-10-13
Unified tz solution

shlomi-noach:

This PR identifies and solves the very special case where a `DATETIME` columns is converted to `TIMESTAMP`. Turns out it is a different case from any other `DATETIME`/`TIMESTAMP` related conversion.

The PR fixes handling of this migration cross different timezones; it will behave correctly even as master, replica and app are all on different time zones.

This PR adds a lot of time-related tests.


go/sql/builder.go
```diff
@@ -83,13 +97,19 @@ func BuildEqualsPreparedComparison(columns []string) (result string, err error)
...
+	if columns.Len() == 0 {
 		return "", fmt.Errorf("Got 0 columns in BuildSetPreparedClause")
 	}
 	setTokens := []string{}
-	for _, column := range columns {
-		setTokens = append(setTokens, fmt.Sprintf("%s=?", EscapeName(column)))
+	for _, column := range columns.Columns() {
+		var setToken string
+		if column.timezoneConversion != nil {
+			setToken = fmt.Sprintf("%s=convert_tz(?, '%s', '%s')", EscapeName(column.Name), column.timezoneConversion.ToTimezone, "+00:00")
```
jonahberquist: Nice, glad this works 👍

shlomi-noach:
Add a (enforced, tested) limitation that a column converting from `DATETIME` to `TIMESTAMP` cannot be part of the shared unique key `gh-ost` iterates on.
This limitation can be lifted in the future given more development.

shlomi-noach:
Sent to production testing

shlomi-noach:
Tests are good. Merging.

Issue #290, opened 2016-10-26
gh-ost aborts on trying to UPDATE, errors on invalid character


mizioumt:
v1.0.21

gh-ost with --test-on-replica failed like this:

```
FATAL Error 1366: Incorrect string value: '\xE1...' for column 'column_name' at row 1; query=

`column_name` is a latin1 VARCHAR and the value coming in the UPDATE statement contained this problematic character:

á
```

So this may be related to the safe SQL mode settings.

The binlogs showed this:

* first, a row with the relevant `id` is INSERTed into the ghost table. At the moment, the data contains no \xE1 character.
* eventually, there is incoming UPDATE for the original table that introduces the \xE1 character. gh-ost refuses to handle it and aborts

mizioumt:
I attempted to gh-ost ALTER a one row table with \xE1 already in the original table. This finished fine. So this sort of character is properly handled by gh-ost when it is INSERTing rows as part of table to table copy. But when the character comes from a binlog UPDATE it is not handled properly.

So it may be possible to construct a simple testcase over a small table when a problematic character is fed through the binlog. I need to learn how stop gh-ost at exactly the right moment. There will likely be no time for interactive command if the table is small.

shlomi-noach:
@mizioumt thank you.

This is strange, because this bug was fixed in `1.0.21`, or at least so I thought. See https://github.com/github/gh-ost/pull/236 and https://github.com/github/gh-ost/pull/227.

`gh-ost` supports [localtests](https://github.com/github/gh-ost/blob/master/doc/local-tests.md), which emulate the exact behavior you'd expect. Let me see if `update`s are properly covered.

shlomi-noach:
please share `show create table` and the `gh-ost` invocation you attempted.
If table secription is too long, please just share the essentials.

shlomi-noach:
See https://github.com/github/gh-ost/pull/291

mizioumt:
the table description is too long and I can't guess what stuff is more essential than whichever other stuff.
So here's the entire thing:

```
*************************** 1. row ***************************
       Table: Table_name
Create Table: CREATE TABLE `Table_name` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `another_column` bigint(20) unsigned NOT NULL DEFAULT '0',
  `column_1` int(10) unsigned NOT NULL DEFAULT '0',
  `column_2` date NOT NULL DEFAULT '0000-00-00',
[...]

PR #291, opened 2016-10-26
more DML tests for latin1

shlomi-noach:
Storyline: #290 

This PR adds `update`, `delete` to `latin1` tests in an attempt to reproduce #290 , but the tests do not fail.


MrMauricioLeite:
This test do not cover TEXT columns. I'm having no trouble with VARCHAR columns but on TEXT columns I'm experiencing the same charset errors other were facing with varchar columns.

Does it make any sense?

shlomi-noach:
I'll add tests for `TEXT` columns as well.

PR #293, opened 2016-10-27
automated builds, CI

shlomi-noach:
This PR begins support for automated, tested builds. This will also make sure CI runs on `push`.

Initially CI will only build the code + run `go` tests. Later we will add the `localtests` (integration tests with master-replica setup)


script/build
```diff
@@ -0,0 +1,21 @@
...
+# We have a few binaries that we want to build, so let's put them into bin/
+
+version=$(git rev-parse HEAD)
+describe=$(git describe --tags --always --dirty)
+
+export GOPATH="$PWD/.gopath"
+cd .gopath/src/github.com/github/gh-ost
+
+# We put the binaries into libexec/ and then put the shim with the appropriate name into bin/
+go build -o "$libexecdir/gh-ost" -ldflags "-X main.AppVersion=${version} -X main.BuildDescribe=${describe}" ./go/cmd/gh-ost/main.go
```
carlosmn: You probably want to put the output in `$bindir/gh-ost` since you don't need the shim and can invoke the actual binary directly.

script/cibuild-gh-ost-build-deploy-tarball
```diff
@@ -0,0 +1,29 @@
...
+if [ $(uname -s) = "Darwin" ]; then
+    build_arch="$(uname -sr | tr -d ' ' | tr '[:upper:]' '[:lower:]')-$(uname -m)"
+else
+    build_arch="$(lsb_release -sc | tr -d ' ' | tr '[:upper:]' '[:lower:]')-$(uname -m)"
+fi
+
+tarball=$build_dir/${commit_sha}-${build_arch}.tar
+
+# Create the tarball
+tar cvf $tarball --mode="ugo=rx" libexec/ bin/
```
carlosmn: If you do the move to `$bindir/` above, you can also remove `libexec/` from here.

script/shim
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash
```
carlosmn: You can get rid of this whole file since you don't need it.

script/godep
```diff
@@ -0,0 +1,18 @@
+#!/bin/bash
```
carlosmn: Do you use godep for managing your dependencies? If not, you can get rid of this file.

script/build
carlosmn: To expand on that, this puts it into `libexec/` but if you don't have the shim in `bin/` then the layout of the resulting deployment will look rather odd, so I'd put it in `bin/` and then you can run `/data/gh-ost/current/bin/gh-ost`.

shlomi-noach: Fixed!

script/godep
shlomi-noach: fixed!

shlomi-noach:
merging now, and may iterate later if need be.

PR #305, opened 2016-11-02
added limitations clarifications: JSON and FEDERATED

shlomi-noach:
cc @ggunson I was already documenting the `JSON` limitation so jumped on the `FEDERATED` as well. I don't think we should add much to the `FEDERATED` case.


ggunson:
I was mostly wanting to know what the federated setup even was that this failed on, because it didn't make a lot of sense to me. (Also, if `SELECT COUNT(*)` causes mysqld to crash, that's not gh-ost's fault).

Issue #314, opened 2016-11-17
gh-ost hangs at `cut-over` phase at "Waiting for events up to lock"


shlomi-noach:
I've had this happen twice now: running a "migrate-on-replica" migration with `postpone-cut-over-flag-file` set; migration properly waits on cut-over, but when instructing `cut-over` the following occurs:

```
Copy: 8488615/8488615 100.0%; Applied: 10182; Backlog: 88/100; Time: 37m0s(total), 36m37s(copy); streamer: mysql-bin.015237:917577121; State: postponing cut-over; ETA: due
2016-11-17 03:00:13 INFO executing gh-ost-on-status hook: /data/github/shell/bin/gh-ost-hooks/prod/gh-ost-on-status-hook
Copy: 8488615/8488615 100.0%; Applied: 10182; Backlog: 89/100; Time: 37m0s(total), 36m37s(copy); streamer: mysql-bin.015237:917607446; State: postponing cut-over; ETA: due
2016-11-17 03:00:13 INFO executing gh-ost-on-status hook: /data/github/shell/bin/gh-ost-hooks/prod/gh-ost-on-status-hook

2016-11-17 03:00:14 INFO Grabbing voluntary lock: gh-ost.1317274408.lock
2016-11-17 03:00:14 INFO Setting LOCK timeout as 6 seconds
2016-11-17 03:00:14 INFO Looking for magic cut-over table
2016-11-17 03:00:14 INFO Creating magic cut-over table `my_schema`.`_my_table_del`
2016-11-17 03:00:14 INFO Magic cut-over table created
2016-11-17 03:00:14 INFO Locking `my_schema`.`my_table`, `my_schema`.`_my_table_del`
2016-11-17 03:00:14 INFO Tables locked
2016-11-17 03:00:14 INFO Session locking original & magic tables is 1317274408
2016-11-17 03:00:14 INFO Writing changelog state: AllEventsUpToLockProcessed
2016-11-17 03:00:14 INFO Waiting for events up to lock
Copy: 8488615/8488615 100.0%; Applied: 10182; Backlog: 94/100; Time: 37m5s(total), 36m37s(copy); streamer: mysql-bin.015237:919839473; State: throttled, max-load Threads_running=434 >= 25; ETA: due
2016-11-17 03:00:21 ERROR critical-load met once: Threads_running=1007, >=1000. Will check again in 3000 millis
[...]

PR #315, opened 2016-11-17
Fixing hanging cut-over waiting on `AllEventsUpToLockProcessed`

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/314

This PR addresses the issue of cut-over hanging while attempting to intercept the `AllEventsUpToLockProcessed` state.

Initial commit:

- `waitForEventsUpToLock()` timeout -- if for whatever reason we are not able to intercept the event, we timeout
- more info (logging) on `AllEventsUpToLockProcessed`, before and after injecting/intercepting

ongoing work.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -156,6 +156,7 @@ type MigrationContext struct {
 	CleanupImminentFlag                    int64
 	UserCommandedUnpostponeFlag            int64
 	CutOverCompleteFlag                    int64
+	InCutOverCriticalSectionFlag           int64
 	PanicAbort                             chan error
 
 	OriginalTableColumnsOnApplier    *sql.ColumnList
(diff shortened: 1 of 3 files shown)


shlomi-noach:
Because `waitForEventsUpToLock()` can time out and retry, we need to make sure that when we consume the `AllEventsUpToLockProcessed`, we're not consuming an older entry.

For that matter, we now inject `AllEventsUpToLockProcessed:<12345>` where `<12345>` is some non repeating number, and are not satisfied with `AllEventsUpToLockProcessed` unless it contains the same number we've just injected. We skip any other entry.

shlomi-noach:
The timeout addition makes it "less terrible" to get the hang scenario, and may cover for yet other cases (e.g. master locks down while we attempt to write changelog event).

shlomi-noach:
We now never throttle within the cut-over critical section. This was supposed to have been the case, however a race condition could happen when throttling began just as unpostpone was requested, at which time the cut-over would block until throttling ended. We now _don't block_ in cut-over.

shlomi-noach:
Also important fix to cut-over retries. It was unrealiable and not adhering to additional throttling; it did not allow to rest the unpostpone command.

shlomi-noach:
sending to testing

shlomi-noach:
testing is good

Issue #318, opened 2016-11-29
False positive for rename column check


dveeden:
Operation:
```
-alter 'add column exchange_rate double comment "exchange rate used for pay in your own currency"'
```

Result:
```
2016-11-29 10:33:17 FATAL gh-ost believes the ALTER statement renames columns, as follows: map[rate:used]; as precation, you are asked to confirm gh-ost is correct, and provide with `--approve-renamed-columns`, and we're all happy. Or you can skip renamed columns via `--skip-renamed-columns`, in which case column data may be lost
```

```
 13 var (
 14         renameColumnRegexp = regexp.MustCompile(`(?i)change\s+(column\s+|)([\S]+)\s+([\S]+)\s+`)
 15 )
```

Would be nice if we could exclude comment from this regex.

shlomi-noach:
Similarly:

```
--alter="add column exchange double comment 'exchange rate used for pay in your own currency'"
```
Results with:

```
2016-11-29 11:03:59 FATAL gh-ost believes the ALTER statement renames columns, as follows: map[rate:used double:comment]; as precaution, you are asked to confirm gh-ost is correct, and provide with `--approve-renamed-columns`, and we're all happy. Or you can skip renamed columns via `--skip-renamed-columns`, in which case column data may be lost
```

shlomi-noach:
local tests to reproduce this error: https://github.com/github/gh-ost/pull/319/files

dveeden:
The mighty ANTLR4 now has a go target... https://github.com/antlr/antlr4/blob/master/doc/go-target.md
I happen to have a MySQLSchemaChange.g4 file...

Would you be interested in switching this from a regex to antlr?

shlomi-noach:
> Would you be interested in switching this from a regex to antlr?

If you're up to do the work, I'm happy to take it

shlomi-noach:
closed by #319. Potentially to be iterated later on via ANTLR4

PR #319, opened 2016-11-29
added tests to verify no false positives rename-column found

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/318

When this branch passes local tests, then https://github.com/github/gh-ost/issues/318 is fixed.

We add tests to catch false-positive rename-column detection on:

- column that has the substring `change`
- comment that has word/substring `change`


shlomi-noach:
The `rename` detection flow now:

- tokenizes `alter` statement into its ingredients (smartly avoids tokenizing on quoted commas, enum commas, etc)
- sanitizes quotes in each such token: completely ignores text within quotes

shlomi-noach:
this passes prod testing

Issue #320, opened 2016-12-01
when alter myisam table and the default engine is innodb, with gtid will be error. 


rj03hou:
we find a **bug** when alter myisam table and the default engine is innodb, with gtid will be error. 

the log is like this:
```
2016-12-01 10:06:29 INFO Creating magic cut-over table `test`.`_test_del`
2016-12-01 10:06:29 INFO Magic cut-over table created
2016-12-01 10:06:29 INFO Locking `test`.`test`, `test`.`_test_del`
2016-12-01 10:06:29 ERROR Error 1785: When @@GLOBAL.ENFORCE_GTID_CONSISTENCY = 1, updates to non-transactional tables can only be done in either autocommitted statements or single-statement transactions, and never in the same statement as updates to transactional tables.
2016-12-01 10:06:29 INFO Looking for magic cut-over table
```

because when load data, the MyISAM engine is better than Innodb,  otherwise Innodb is better, so our default engine is innodb, and some load data table is MyISAM.

So when we alter the myisam table, the  magic cut-over table is default engine innodb. when gtid is on, GLOBAL.ENFORCE_GTID_CONSISTENCY is 1, can't allow we use innodb and myisam in the same transaction, so we execute "LOCK TABLES tbl WRITE, tbl_magic WRITE", the error happens.

I advice the easy way is set magic cut-over table's engine with the same withe the original table.

rj03hou:
I make a pull request, when I make sure the test.sh work, I will pull it.

shlomi-noach:
closed by https://github.com/github/gh-ost/issues/325

PR #321, opened 2016-12-01
If the original table is MyISAM and the default engine is Innodb, and…

rj03hou:
Related issue: https://github.com/github/gh-ost/issues/320

The following is all done.

### Description

just make sure the the magic cut-over table's engine same with the original table.
detail description is in https://github.com/github/gh-ost/issues/320.

… the gtid mode is on, there will be error when execute 'LOCK TABLES tbl WRITE, tbl_magic WRITE'. If make the magic cut-over table's engine same with the original table, there will not have this problem.


shlomi-noach:
Looking good. Thank you

go/logic/applier.go
```diff
@@ -703,8 +703,9 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
 
 	query := fmt.Sprintf(`create /* gh-ost */ table %s.%s (
 			id int auto_increment primary key
-		) comment='%s'
+		) engine=%s comment='%s'
 		`,
+		this.migrationContext.TableEngine,
```
shlomi-noach: `TableEngine` correlates to the 3rd placeholder in the template string, not the 1st

rj03hou:
I change the order, and make the test. 
sorry about this, I make change and test in one environment, and push the code another environment.

I unify the environment, next time, will be ok.

shlomi-noach:
Resubmitted as https://github.com/github/gh-ost/pull/325

Issue #323, opened 2016-12-02
Not support for dirty replication


rj03hou:
Find a bug not  support for clean replication:

1. DBA using reset slave rather than reset slave all. The Master_Host and Master_Port exist.
1. DBA not execute reset slave or reset slave all when they promote the slave to master. The Master_Host and Master_Port exist, but the replication is break.

In the above case, we using recursive to execute show slave status from the host to find the master but not check the slave status, may find the wrong master.

So we can check the Master_Log_File and Slave_IO_Running Slave_SQL_Running.

```
if show slave status output is not empty:
    if Master_Log_File is empty:
        return
    else    Slave_IO_Running or Slave_SQL_Running not yes:
       exit
    else
       this may be the master
```

shlomi-noach:
Thank you,

`Slave_IO_Running` and `Slave_SQL_Running` being `No` is certainly not a reliable indicator that a server is a master.

I'm OK to consider an empty `Master_Log_File` as an indicator that a server is a master.

rj03hou:
In the second case:
DBA not execute reset slave or reset slave all when they promote the slave to master, just execute stop slave. The Master_Host and Master_Port exist, but the replication is break.

In this case, we can find the stale master, and try to write to id.

shlomi-noach:
Also in that same 2nd case, some intermediate master may just have replication broken; it's unsafe to assume that it is the master.

rj03hou:
In that same 2nd case, if the Slave_IO_Running and Slave_SQL_Running being No, we can stop and remind the user to check.
otherwise will find the wrong master.

shlomi-noach:
> we can stop and remind the user to check.

yes, that's an excellent sanity check.

rj03hou:
@shlomi-noach I make a pull request to solve this problem,  can you review the code.

shlomi-noach:
closed by https://github.com/github/gh-ost/issues/324

PR #324, opened 2016-12-02
check the slave status when find recursive find the master, so suppor…

rj03hou:
Related issue: https://github.com/github/gh-ost/issues/323

The following is all don.
- [x] the binary is test.
- [x] the log does not output.

### Description

1. DBA using reset slave rather than reset slave all. The Master_Host and Master_Port exist.
2. DBA not execute reset slave or reset slave all when they promote the slave to master. The Master_Host and Master_Port exist, but the replication is break.

In the above case, we using recursive to execute show slave status from the host to find the master but not check the slave status, may find the wrong master.

So we can check the Master_Log_File and Slave_IO_Running Slave_SQL_Running.

```
if show slave status output is not empty:
    if Master_Log_File is empty:
        return
    else    Slave_IO_Running or Slave_SQL_Running not yes:
       exit
    else
       this may be the master
```

**Have a question,** I think log before we recursive  find the master, but he log.Infof not work in go/logic/inspect.go.

### DIFF
--- a/go/mysql/utils.go
+++ b/go/mysql/utils.go
@@ -83,7 +83,10 @@ func GetMasterKeyFromSlaveStatus(connectionConfig *ConnectionConfig) (masterKey 
 		return nil, err
 	}
 	err = sqlutils.QueryRowsMap(db, `show slave status`, func(rowMap sqlutils.RowMap) error {
-		// Using reset slave not reset slave all, the Master_Log_File is empty.
+		// We wish to recognize the case where the topology's master actually has replication configuration.
+		// This can happen when a DBA issues a `RESET SLAVE` instead of `RESET SLAVE ALL`.
+
+		// An empty log file indicates this is a master:
 		if rowMap.GetString("Master_Log_File") == "" {
 			return nil
 		}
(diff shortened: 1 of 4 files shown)


go/mysql/utils.go
```diff
@@ -83,12 +83,27 @@ func GetMasterKeyFromSlaveStatus(connectionConfig *ConnectionConfig) (masterKey
...
 	}
 	err = sqlutils.QueryRowsMap(db, `show slave status`, func(rowMap sqlutils.RowMap) error {
+		// Using reset slave not reset slave all, the Master_Log_File is empty.
+		if rowMap.GetString("Master_Log_File") == "" {
+			return nil
+		}
+
+		// If not using reset slave or reset slave all and slave is break, something is wrong, so we stop.
+		if rowMap.GetString("Slave_IO_Running") != "Yes" || rowMap.GetString("Slave_SQL_Running") != "Yes" {
+			log.Fatalf("The slave %s is break: Slave_IO_Running:%s Slave_SQL_Running:%s please make sure the replication is correct.",
```
shlomi-noach: `Fatalf` shouldn't be called from within an internal utility. The utility should return with an error, and it is up to the calling code to decide whether to panic or not.

go/mysql/utils.go
```diff
@@ -83,12 +83,27 @@ func GetMasterKeyFromSlaveStatus(connectionConfig *ConnectionConfig) (masterKey
...
+				connectionConfig.Key.Hostname,
+				rowMap.GetString("Slave_IO_Running"),
```
shlomi-noach: only `GetString()` once for each column, into a variable, then use that variable.

rj03hou:
➜  gh-ost git:(replication-reset-slave) sh test.sh
PASS
ok  	github.com/github/gh-ost/go/base	0.010s
PASS
ok  	github.com/github/gh-ost/go/mysql	0.009s
2016-12-05 19:44:27 ERROR []string{"add column t int", "add column e enum('a','b','c')"}
2016-12-05 19:44:27 ERROR []string{"add column t int(11)", "add column e enum('a','b','c')"}
PASS
ok  	github.com/github/gh-ost/go/sql	0.010s


I check something is wrong in the official master, so my test is not pass, I build and test is OK.

Another question, can you make a release include the myisam-gtid and this recently, our production meet this two question.

shlomi-noach:
Thank you,
I will release shortly, sometime this week!

PR #325, opened 2016-12-02
Myisam gtid

shlomi-noach:
(resubmission of https://github.com/github/gh-ost/issues/321 by @rj03hou)

Related issue: https://github.com/github/gh-ost/issues/320

The following is all done.

### Description

just make sure the the magic cut-over table's engine same with the original table.
detail description is in https://github.com/github/gh-ost/issues/320.

… the gtid mode is on, there will be error when execute 'LOCK TABLES tbl WRITE, tbl_magic WRITE'. If make the magic cut-over table's engine same with the original table, there will not have this problem.


shlomi-noach:
Thanks @rj03hou

Issue #329, opened 2016-12-09
don't check the critical-load at cut-over section


rj03hou:
I check the 1.0.30 release note, find that in this version we find fix a bug "Fixed cut-over stall", but we meet another similar case:

```
2016-12-07 19:12:17 INFO Creating magic cut-over table `test`.`_test_del`
2016-12-07 19:12:17 INFO Magic cut-over table created
2016-12-07 19:12:17 INFO Locking `test`.`test`, `test`.`_test_del`
2016-12-07 19:12:17 INFO Tables locked
2016-12-07 19:12:17 INFO Session locking original & magic tables is 2164861
2016-12-07 19:12:17 INFO Writing changelog state: AllEventsUpToLockProcessed:1481109137528939398
2016-12-07 19:12:17 INFO Intercepted changelog state AllEventsUpToLockProcessed
2016-12-07 19:12:17 INFO Handled changelog state AllEventsUpToLockProcessed
2016-12-07 19:12:17 INFO Waiting for events up to lock
2016-12-07 19:12:18 FATAL critical-load met: Threads_running=54, >=32
```

means critical-load check may happens in the cut-over section, when we lock the tables, the  Threads_running may increase quickly but shortly, so I think we should not check the critical-load in the cut-over section.

I try to fix this, but it not  easy to reproduction, I modify the code to reproduce.
```go
//applier.go
func (this *Applier) makeTestLoad() {
	for {
		sqlutils.ExecNoPrepare(this.db, `select count(*) from test`)
	}

}

//migrator.go add the following code to atomicCutOver method
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()
	go this.applier.makeTestLoad()

	time.Sleep(2 * time.Second)

	//lockOriginalSessionId := <-lockOriginalSessionIdChan
```

and set the parameter
--max-load=Threads_running=10 \
--critical-load=Threads_running=10 \

to reproduce the problem. And I think we should cleanup if we meet a panic abort.

shlomi-noach:
This issue has been considered in https://github.com/github/gh-ost/issues/216, and solved by https://github.com/github/gh-ost/pull/266

[...]

PR #330, opened 2016-12-09
don't check the critical-load at cut-over section

rj03hou:
Related issue: https://github.com/github/gh-ost/issues/329

### Description

don't check the critical-load at cut-over section, details in the issue.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -219,7 +219,11 @@ func (this *Migrator) onChangelogStateEvent(dmlEvent *binlog.BinlogDMLEvent) (er
 // listenOnPanicAbort aborts on abort request
 func (this *Migrator) listenOnPanicAbort() {
 	err := <-this.migrationContext.PanicAbort
+
+	//should cleanup before exit.
+	this.finalCleanup()
 	log.Fatale(err)
+
 }
 
 // validateStatement validates the `alter` statement meets criteria.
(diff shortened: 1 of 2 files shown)


go/logic/throttler.go
```diff
@@ -163,10 +163,15 @@ func (this *Throttler) collectGeneralThrottleMetrics() error {
 	if err != nil {
 		return setThrottle(true, fmt.Sprintf("%s %s", variableName, err), base.NoThrottleReasonHint)
 	}
-	if criticalLoadMet && this.migrationContext.CriticalLoadIntervalMilliseconds == 0 {
+
+	// we don't throttle even critical-load when cutting over.
+	isThrottle, _, _ := this.migrationContext.IsThrottled()
+	isThrottle = true
```
shlomi-noach: ?
This is probably a left over of some experiment of yours

go/logic/migrator.go
```diff
@@ -219,7 +219,11 @@ func (this *Migrator) onChangelogStateEvent(dmlEvent *binlog.BinlogDMLEvent) (er
 // listenOnPanicAbort aborts on abort request
 func (this *Migrator) listenOnPanicAbort() {
 	err := <-this.migrationContext.PanicAbort
+
+	//should cleanup before exit.
+	this.finalCleanup()
```
shlomi-noach: I disagree. This is what panic is all about; getting here means something went wrong; attempting to clean up may make it worse.

rj03hou: How do you do in github, when fail, the dba check the reason and clean up? critical error may b often.

shlomi-noach: there's not much to clean up, really, just the remains of the `_gho` table. We don't have often critical errors, or at least we work hard not to have often critical errors.

Issue #332, opened 2016-12-09
Runtime error when trying to ALTER a table without PRIMARY KEY / UNIQUE KEY


ramkireddy:
Running into a following run time error when trying to ALTER a table.

```
#./gh-ost --max-load=Threads_running=25 --critical-load=Threads_running=1000 --chunk-size=1000 --max-lag-millis=1500 --user="ghuser" --password="xxxxxxxx" --host=sand1mysql-01 --database="czen" --table="ZIP_DIST_TEST" --verbose --alter="ADD COLUMN DESCRIPTION VARCHAR(255) NULL" --allow-master-master --cut-over=default --exact-rowcount --concurrent-rowcount --default-retries=120 --initially-drop-old-table --initially-drop-ghost-table --panic-flag-file=/tmp/ghost.panic.flag --postpone-cut-over-flag-file=/tmp/ghost.postpone.flag --execute
2016-12-09 16:12:31 INFO starting gh-ost 1.0.30
2016-12-09 16:12:31 INFO Migrating `czen`.`ZIP_DIST_TEST`
2016-12-09 16:12:31 INFO connection validated on dom-use-sand1mysql-04:3306
2016-12-09 16:12:31 INFO User has SUPER, REPLICATION SLAVE privileges, and has ALL privileges on `czen`.*
2016-12-09 16:12:31 INFO binary logs validated on dom-use-sand1mysql-04:3306
2016-12-09 16:12:31 INFO Restarting replication on dom-use-sand1mysql-04:3306 to make sure binlog settings apply to replication thread
2016-12-09 16:12:31 INFO Table found. Engine=InnoDB
2016-12-09 16:12:31 INFO Estimated number of rows via EXPLAIN: 10134006
2016-12-09 16:12:31 INFO Recursively searching for replication master
2016-12-09 16:12:31 INFO Master found to be sand1-mysqlmaster:3306
2016-12-09 16:12:31 INFO log_slave_updates validated on dom-use-sand1mysql-04:3306
2016-12-09 16:12:31 INFO connection validated on dom-use-sand1mysql-04:3306
2016-12-09 16:12:31 INFO Registering replica at dom-use-sand1mysql-04:3306
2016-12-09 16:12:31 INFO Connecting binlog streamer at bin-log.000002:220592054
2016-12-09 16:12:31 INFO rotate to next log name: bin-log.000002
2016-12-09 16:12:31 INFO connection validated on sand1-mysqlmaster:3306
[...]

PR #333, opened 2016-12-11
fix: bailing out on no PRIMARY/UNIQUE KEY

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/332

`gh-ost` must bail out when no `PRIMARY KEY` or otherwise a `UNIQUE KEY` can be found on the migrated table.
A bug prevented that error from presenting itself, causing `gh-ost` to panic on a later stage.


shlomi-noach:
passes testing in prod

Issue #205, opened 2016-08-27
Allow a migration to pause and resume under a new gh-ost process


jonahberquist:
We've [previously discussed in passing](https://github.com/github/gh-ost/issues/99#issuecomment-234613334) the potential to do this. Our conclusion at the time was that it would be possible, though not particularly useful. The use case that I had thought of at the time was resuming if a `gh-ost` process died.

I have a more interesting use case now, which is the desire to change `gh-ost` itself during a long migration.

As an example, if during the course of a 25 hour migration, I decide I want to change the value of a configuration option to something less than a hard coded minimum, I am out of luck. There are few of these hard coded configuration options, but I can imagine other times where a behavior change that would require a change in `gh-ost` itself could be beneficial to an ongoing migration.

@shlomi-noach, imagine having this in conjunction with branch deploys.

I would be happy to work on the development of this feature.

SuperQ:
I guess the easiest thing for this would be to write out a checkpoint file containing the necessary information.  It could be written to a `gh-ost` checkpoint table in mysql, but I think a local file would be better.

pbitty:
This would be a very nice feature to have.  We've been running migrations as jobs on our compute cluster.  We could treat this as a resumable job and wouldn't have to allocate a specific long-running host to run the job - if a host is set to be terminated and the job gets moved to a different host, it would just resume automagically.

I'd be happy to help develop/test this.

I would vote for having the checkpoint data stored in a table instead of a file, so the job itself doesn't need to keep any state.

shlomi-noach:
Terminology wise let's find a different name for this since pause/resume are already used to describe throttling/unthrottling.

SuperQ:
I would call it shutdown checkpoint and restart from checkpoint.

shlomi-noach:
In order to make that happen, a new invocation needs to have the context to a previously interrupted migration. On the grander scale, this must include:

- The last processed copy-range
- The last applied bin log position

On the finer scale we are also interested in:

- reading the unique-key by which migration chooses to iterate
- "last processed copy-range" must be **typed** values
- last applied bin log position -- not too very important. Turns out RBR is idempotent. OK to resume from an already applied position and re-apply it.
- For that matter, same goes for rowcopy-range. This means we don't strictly need fsync or two-phase commit for everything. Periodic flush of status is good enough.
- all configuration
- all calculated values (number of rows? applied rows? time estimate?)

[...]

PR #343, opened 2016-12-20
WIP: Resurrect

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/205

WORK IN PROGRESS: resurrecting a migration after failure.
The idea is that `gh-ost` would routinely dump migration status/context. It would be ossible for one `gh-ost` process to fail (e.g. having met `critical-load`) and for another `gh-ost` process to pick up from where the first left off.

Initial commits present exporting of migration context, with some shuffling & cleanup.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -6,7 +6,9 @@
 package base
 
 import (
+	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"os"
 	"regexp"
 	"strings"
@@ -158,7 +160,6 @@ type MigrationContext struct {
 	UserCommandedUnpostponeFlag            int64
 	CutOverCompleteFlag                    int64
 	InCutOverCriticalSectionFlag           int64
-	PanicAbort                             chan error
 
 	OriginalTableColumnsOnApplier    *sql.ColumnList
 	OriginalTableColumns             *sql.ColumnList
@@ -174,8 +175,6 @@ type MigrationContext struct {
 	Iteration                        int64
 	MigrationIterationRangeMinValues *sql.ColumnValues
 	MigrationIterationRangeMaxValues *sql.ColumnValues
-
-	CanStopStreaming func() bool
 }
 
 type ContextConfig struct {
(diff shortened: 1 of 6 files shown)


shlomi-noach:
TODO:

- [ ] must not export passwords

shlomi-noach:
Export is to _changelog_ table. This ensures atomicy and durability of write, assuming _changelog_ table is `InnoDB`. notable is that if migrated table is `MyISAM`, so is the changelog table.

I'm :fine: stating that resurrection does not work on `MyISAM`, because `MyISAM`.

shlomi-noach:
https://github.com/github/gh-ost/pull/343/commits/5f25f741ad22e0d234376333e8c7688b071f16b1 makes for something that works! I'll need to iterate to see what has been overlooked, but basically we're getting there fast.

shlomi-noach:
A concern is to not rely on the streamer's last known position, because streamer writes to a buffer (currently hard coded to `100` events). Those events would be lost upon resurrection.

- [ ] Instead, we should have the migration report the last _applied_ event's coordinates.

tomkrouper:
Off issue, you were mentioning gh-ost was having checksum issues with resurrections. When you mentioned that I was thinking, could it be related to the fact that we have two things going on: the backlog and the iteration of inserts? I hope that makes sense. None the less, it was something that popped in my head that I hoped might help when you get back to this. (Not fully understanding the code changes, this might already be something you're handling.)

shlomi-noach:
@tomkrouper the conjecture is as follows:

- assuming `gh-ost` breaks while copying rows `5,000`-`5,100`
- and while reading `mysql-bin.000120` at position `123456`

it should be OK to resume execution

- start with copying rows `4,300`-`4,400` (way before the point of breakage)
- start with reading binary log `mysql-bin.000120` at position `121234` (way before the point of breakage)

this is the conjecture's logic:

- re-copying same rows just overwrites existing rows (or adds rows that weren't there before!)
- re-applying binary logs is an idempotent action
I find it a bit difficult right now to substantiate these claims, but I believe them to be true.

But then, of course, tests are failing...

Xopherus:
@shlomi-noach are there any plans to revisit this feature? I'm looking at gh-ost again and one of the concerns my team has is that we have some very large tables that can take days, if not a week to copy. If the process were to crash in the middle we'd have a lot of wasted effort, especially when we have to slowly drain the _gho table to prevent the dreaded global metadata lock when dropping it. This would be extremely helpful for us!

shlomi-noach:
> I've found that tuning gh-ost is one of the challenges because the feedback cycles are so long

@Xopherus  Could you please elaborate on that? I'm not sure I understand.

Xopherus:
Oh I just mean that if your migrations can take multiple hours / days, it can be tricky to tune parameters (e.g. critical load threshold or lock cutover timeouts, etc) because it takes longer to experiment. Fortunately we've gotten solid advice from you and others here to help guide us in the right direction.

daniel-nichter:
Hi @shlomi-noach :-) I think "resurrect" is not the best term. It's not a standard technical term. Even doc/command-line-flags.md has to clarify: "It is possible to resurrect/resume a failed migration". When people think, "Can I resume an osc?", they'll look for and Google with that term. Imho, "resurrect" will never cross people's minds. By contrast, everyone knows what "resume" (and its reciprocals "suspend" or "pause") mean. I'd also argue that it's not technically descriptive or intention-revealing. A dead body can be resurrected, and I get the joke with the app being called "ghost", but it begs the question: What does it mean to resurrect a program? My last argument is: for non-native English speakers/readers, these issues are compounded by uncommon words in a technical context. --  I'd vote for pause/resume or start/stop.

shlomi-noach:
Thank you @daniel-nichter

sfgghh:
0507744451

sfgghh:
0507744451

rakhibs:
bumping this feature request to check on if there were any changes to make this feature available?

tomkrouper:
> bumping this feature request to check on if there were any changes to make this feature available?

This code is fairly old and there are a bunch of conflicting files at this point. We don't have any immediate plans to work on this, but I do agree this would be a good feature to have and if anyone would like to continue the work, we'd love the community contribution.

meiji163:
superseded by https://github.com/github/gh-ost/pull/1595

Issue #230, opened 2016-09-12
Support `binlog_row_image=MINIMAL`


shlomi-noach:
Right now `gh-ost` only supports `binlog_row_image=FULL`

ggunson:
Given the issues we saw with https://github.com/github/gh-ost/issues/3, is that possible? We were seeing some weird behaviour with tables that had primary keys as well as unique keys.

shlomi-noach:
I believe `gh-ost` should be fine updating a `PK` value reading it from the binary logs. [local tests](https://github.com/github/gh-ost/blob/master/doc/local-tests.md) now give us a good platform to confirm that.

shlomi-noach:
Unsure yet how to differentiate minimal-NULLs and real-NULLs. Looking at:
```
create table tnull(id int auto_increment primary key, c1null int default null, c2null int default null, c3 int not null default 0);
insert into tnull values (null, 4, 4, 4);
update tnull set c1null=44 where c3=4;
```

`go-mysql` output:
```
=== XIDEvent ===
Date: 2016-09-13 07:55:11
Log position: 33349
Event size: 31
XID: 482593

=== QueryEvent ===
Date: 2016-09-13 07:55:11
Log position: 33421
Event size: 72
Slave proxy ID: 199978
Execution time: 0
Error code: 0
Schema: test
Query: BEGIN

=== TableMapEvent ===
Date: 2016-09-13 07:55:11
Log position: 33472
Event size: 51
TableID: 927
TableID size: 6
Flags: 1
Schema: test
Table: tnull
Column count: 4
Column type:
00000000  03 03 03 03                                       |....|
NULL bitmap:
00000000  06                                                |.|

=== UpdateRowsEventV2 ===
Date: 2016-09-13 07:55:11
Log position: 33518
Event size: 46
TableID: 927
Flags: 1
Column count: 4
Values:
--
0:4
1:<nil>
2:<nil>
3:<nil>
--
0:<nil>
1:44
2:<nil>
3:<nil>
```
what in the above suggests that `c2null` was kept unchanged as opposed to being set as `null`?

Followed by 

```
update tnull set c2null=46 where c3=4;
```

I see:

```
=== UpdateRowsEventV2 ===
Date: 2016-09-13 08:10:40
Log position: 33918
Event size: 46
TableID: 927
Flags: 1
Column count: 4
Values:
--
0:4
1:<nil>
2:<nil>
3:<nil>
--
0:<nil>
1:<nil>
2:46
[...]

PR #345, opened 2016-12-25
WIP: support for MINIMAL row image

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/230

This PR and ongoing commits will address the `MINIMAL` row image format.
Next comments will clarify how this works.

This is work in progress, with a few implications still to tackle.

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -876,33 +876,48 @@ func (this *Applier) ShowStatusVariable(variableName string) (result int64, err 
 
 // buildDMLEventQuery creates a query to operate on the ghost table, based on an intercepted binlog
 // event entry on the original table.
-func (this *Applier) buildDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) (query string, args []interface{}, rowsDelta int64, err error) {
+func (this *Applier) buildDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) (query string, args []interface{}, dataViaBinlog bool, rowsDelta int64, err error) {
 	switch dmlEvent.DML {
 	case binlog.DeleteDML:
 		{
 			query, uniqueKeyArgs, err := sql.BuildDMLDeleteQuery(dmlEvent.DatabaseName, this.migrationContext.GetGhostTableName(), this.migrationContext.OriginalTableColumns, &this.migrationContext.UniqueKey.Columns, dmlEvent.WhereColumnValues.AbstractValues())
-			return query, uniqueKeyArgs, -1, err
+			return query, uniqueKeyArgs, true, -1, err
 		}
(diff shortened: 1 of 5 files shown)


shlomi-noach:
Here's the implementation, and the problem with the implementation:

First, with `MINIMAL` image, the chunking key must be the `PRIMARY KEY` -- because `PK` columns ar the only ones guaranteed to be found in the binary logs. This is a reasonable constraint.

Now, assuming our chunking key is indeed `PK`: we cannot rely on any other column existing in the binary log. Some statements may update some column `c1`, others may update some other column `c2`, and only affected columns are present in the binary logs.

So instead of reading entire row content from the binary log, we only read the `PK` values, and then issue a `INSERT INTO _ghost_ SELECT * FROM _original_table_ WHERE (<pk columns=pk values>)`

> Side note, to become relevant shortly: this re-couples original and ghost tables

This unfortunately re-introduces the problem of charsets/timezones. Remember binarylogs convert all texts to `utf8` and all timestamps to `utc`. Now, what if we have, for example, a `id (int), ts (timestamp)` as `PRIMARY KEY` (example of [failing test](https://github.com/github/gh-ost/blob/41013901b1c8316cb2f59740a3a446ac5126b8e7/localtests/compound-pk-ts/create.sql)?

When we read the `ts` in the binary log it's `utc`. We cannot _find_ the value via `WHERE (<pk columns=pk values>)` unless we `SET time_zone='+00:00'`. Sound good?

Bad. What if there's additional `ts2 (timestamp), dt2 (datetime)` columns on the table? Now _they_ get corrupted by the `SET time_zone='+00:00'`.

Our problem is that we mix time stamps that are _natively_ there (`ts2`, `dt2`), where a simple `insert into...select` should just work, with timestamps that we imports from the outside (`ts`, retrieved from the binary log), for which we must perform conversion.

Such conversion may not happen in the app, since the app doesn't have all timezone info that MySQL has.

I suspect the `insert into...select` solution would not work.

#### Suggested alternative 

Instead, we may need to _even further_ decouple _original table_ from _ghost table_. What we'd need to do is a read-then-write:

- `select * from _original_table_ where <pk_columns=pk_values>` -- load all columns into the app
- `insert/replace into _ghost_table_ values (...)`

The first would use `utf8`/`utc` conversions to _find_ the matching row
The second would just apply all read values without conversion.

#### Implication of suggested alternative

For each event we find in the binary log we would need to execute a `SELECT` and then an `INSERT`, where previously we only issued an `INSERT`.

This makes for an overhead.

However, this also gets us one step closer to "phase 2", where we don't even use `insert ... select` for row copy, which would allows us to migrate tables remotely. It's literally the same algorithm.

So I'm going to pursue that.

PR #349, opened 2016-12-28
Add possibility to set serverId as gh-ost option

cypiszzz:
Signed-off-by: Cyprian Nosek <cypiszzz@gmail.com>

In a case where I want to run several gh-ost processes in parallel it seems to be not possible due to all of the processes show up with the same serverId: 99999 (it's const in the code).

Possibility to pass serverId as an option with command line (with default set to 99999) would solve the problem.

The change adds optional `-replica-server-id` flag where it's possible to set serverId.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -123,6 +123,7 @@ type MigrationContext struct {
 	InitiallyDropOldTable        bool
 	InitiallyDropGhostTable      bool
 	CutOverType                  CutOver
+	ReplicaServerId              uint
 
 	Hostname                               string
 	AssumeMasterHostname                   string
--- a/go/binlog/gomysql_reader.go
+++ b/go/binlog/gomysql_reader.go
@@ -9,16 +9,13 @@ import (
 	"fmt"
 	"sync"
 
+	"github.com/github/gh-ost/go/base"
 	"github.com/github/gh-ost/go/mysql"
 	"github.com/github/gh-ost/go/sql"
 
 	"github.com/outbrain/golib/log"
 	gomysql "github.com/siddontang/go-mysql/mysql"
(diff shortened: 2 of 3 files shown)


shlomi-noach:
Thank you. This should ideally be submitted on https://github.com/siddontang/go-mysql, and of course we can diverge from that library a bit.

Having said that, we've been running multiple concurrent migrations without problem. What is the error you saw?

cypiszzz:
Hi, thanks for the quick reply. The error I meet is as follows:

> INFO StreamEvents encountered unexpected error: ERROR 1236 (HY000): A slave with the same server_uuid/server_id as this slave has connected to the master; the first event 'mysql-bin.000004' at 4, the last event read from '/var/log/mysql/mysql-bin.000004' at 246679, the last byte read from '/var/log/mysql/mysql-bin.000004' at 246679

shlomi-noach:
What's your `select version()`? I'm curious since I never got this error.

cypiszzz:
```
mysql> SELECT VERSION();
+------------+
| VERSION()  |
+------------+
| 5.6.34-log |
+------------+
1 row in set (0.00 sec)
```

I've noticed the same error with mysql 5.6.19. 
I tested a little since yesterday and so far migrations were able to successfully done despite the error. 

Some time ago I've tested under heavy traffic and I remember there were some problems with migration, unfortunately I'm not 100% they were connected to this specific error, because I had also another problems I solved on the way.  Now I can't reproduce failed migration (even though the error is present), but I don't have possibility to test under heavy traffic for now. 

As an additional information, when I see the error I also see a lot of messages like:

> DEBUG Skipping handled query at mysql-bin.000004:35549

Setting unique serverId solves this.

shlomi-noach:
Thank you

Issue #348, opened 2016-12-27
gh-ost seems blocked when Backlog get 100/100


sumonsun:
#### when Backlog get 100/100,gh-ost didn't work,can we increase this parameter?

```
gh-ost --allow-on-master --max-load=Threads_running=30 --critical-load=Threads_running=45 --chunk-size=1000 --max-lag-millis=1500 --initially-drop-old-table --initially-drop-ghost-table --initially-drop-socket-file --ok-to-drop-table --host="10.0.xxx" --port=3306 --user="xxx" --password="xxx" --database="test" --table="comm_go_ost" --verbose --alter="add column testa varchar(50)" --switch-to-rbr --cut-over=default --default-retries=120 --panic-flag-file=/tmp/ghost.panic.flag --postpone-cut-over-flag-file=/tmp/ghost.postpone.flag  --throttle-query="select 0" --execute
	
	
Copy: 4979000/22690750 21.9%; Applied: 612609; Backlog: 100/100; Time: 11m30s(total), 11m30s(copy); streamer: mysql-bin.000187:386543989; State: migrating; ETA: 40m54s
Copy: 4979000/22690750 21.9%; Applied: 641517; Backlog: 100/100; Time: 12m0s(total), 12m0s(copy); streamer: mysql-bin.000187:390623293; State: migrating; ETA: 42m41s
Copy: 4979000/22690750 21.9%; Applied: 670192; Backlog: 100/100; Time: 12m30s(total), 12m30s(copy); streamer: mysql-bin.000187:394667040; State: migrating; ETA: 44m27s
Copy: 4979000/22690750 21.9%; Applied: 698549; Backlog: 100/100; Time: 13m0s(total), 13m0s(copy); streamer: mysql-bin.000187:398671937; State: migrating; ETA: 46m14s
Copy: 4979000/22690750 21.9%; Applied: 727394; Backlog: 100/100; Time: 13m30s(total), 13m30s(copy); streamer: mysql-bin.000187:402753149; State: migrating; ETA: 48m1s
```

shlomi-noach:
If the backlog is `100/100` that means `gh-ost` is unable to apply the binlog events it reads from the binary logs onto the _ghost_ table.

Typically this is because `gh-ost` is throttling, however this isn't your case. In you case, there is no throttling, and `gh-ost` advances in the binary log: `mysql-bin.000187:386543989`, `mysql-bin.000187:390623293`, ...

[...]

PR #352, opened 2017-01-03
Batch apply dml events

shlomi-noach:
This PR applies DML events (events read from the binary logs) in batches, i.e. it aggregates events from the binary log and writes multiple DMLs (`insert`, `delete`, `update`) in a single transaction.

This has the potential to increase `gh-ost`'s write capacity. This would _potentially_ help https://github.com/github/gh-ost/issues/348, though we have not yet diagnosed https://github.com/github/gh-ost/issues/348 enough.

The PR introduces a `--dml-batch-size` flag, default `10`.

Noteworthy about the implementation: `gh-ost` does not _wait_ for events to arrive in order to batch. The logic is as follows:

- If there's nothing in the event queue (coming from the binary log), copy some rows
- If there's an event to handle in the events queue, read it. Further determine how many more _available_ events are in the queue. Of these, read _at most_ `dml-batch-size`. Of these, only handle the prefix that is true DML events (distinguished from `gh-ost`'s internal events).

So the logic doesn't block on waiting for "more events to arrive", but is greedy, and takes as many available events as there are, up to some limit, and only in suh case that _there are events to apply_.


shlomi-noach:
This was subjected to prod testing for the past days and behaves as expected; merging.

Issue #353, opened 2017-01-08
Gh-ost breaks when using unsigned MEDIUMINT


serdarsahin:
When using unsigned integer data types, binary log behavior changes if the inserted number is higher than the max signed integer value.

The original table
```
CREATE TABLE `test_table` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `column1` int(11) NOT NULL,
  `column2` int(11) NOT NULL,
  `column3` mediumint(8) unsigned NOT NULL,
  `column4` tinyint(3) unsigned NOT NULL,
  `column5` int(11) NOT NULL,
  `column6` int(11) NOT NULL,
  PRIMARY KEY (`id`),
  KEY `user_id_level_id` (`column1`,`column2`)
) ENGINE=InnoDB AUTO_INCREMENT=4647930180 DEFAULT CHARSET=utf8
```

The new schema;

```
CREATE TABLE `test_table` (
  `id` bigint(20) NOT NULL AUTO_INCREMENT,
  `column1` int(11) NOT NULL,
  `column2` smallint(5) unsigned NOT NULL,
  `column3` mediumint(8) unsigned NOT NULL,
  `column4` tinyint(3) unsigned NOT NULL,
  `column6` int(11) NOT NULL,
  PRIMARY KEY (`column1`,`column2`),
  UNIQUE KEY `id` (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=4632662231 DEFAULT CHARSET=utf8
```

We were experiencing the same issue explained on the issue #348 . Alter was never starting, even for hours. The table has more than 4 billion records and receiving 200 million records per week. So we tested the version that has included in this comment: https://github.com/github/gh-ost/issues/348#issuecomment-270592112 and it was quite fast. It was estimating 35 hours to complete which is perfectly acceptable for us.

We then executed the following statement

```
[...]

PR #355, opened 2017-01-09
localtests and fix for unsigned medium int problem

shlomi-noach:
Trying to reproduce https://github.com/github/gh-ost/issues/353, ~~insofar without success.~~, with success and fix


shlomi-noach:
The test now reproduced the problem for which it was created:

```
2017-01-10 09:16:23 FATAL Error 1264: Out of range value for column 'column3' at row 1; query=
       			replace /* gh-ost `test`.`_gh_ost_test_gho` */ into
       				`test`.`_gh_ost_test_gho`
       					(`id`, `column1`, `column2`, `column3`, `column4`, `column6`)
       				values
       					(?, ?, ?, ?, ?, ?)
       		; args=[6 13382498 536 4288190080 3 1483892218]
```

shlomi-noach:
The test is now simplified to a standard `engine=innodb`. Adding/dropping indexes/columns are not needed to reproduce the problem.

shlomi-noach:
This is under testing, and then to be merged.

shlomi-noach:
this passes tests. merging

Issue #357, opened 2017-01-10
Identical PK or unique key between old and new tables


ggunson:
This is coming out of the issue https://github.com/github/gh-ost/issues/353 but unrelated to the bug it reported. 

In that issue, there's an alter statement within the gh-ost command:

```
--alter="ADD UNIQUE INDEX (id), 
DROP COLUMN column5, DROP INDEX column1_column2, 
DROP PRIMARY KEY, ADD PRIMARY KEY(column1,column2), 
MODIFY column2 SMALLINT UNSIGNED NOT NULL, MODIFY column3 INT NOT NULL"
```

In that case, the primary key becomes a unique key and a non-unique index becomes the new primary key. All columns involved in these indexes are not-null. 

The docs, however, state (https://github.com/github/gh-ost/blob/master/doc/requirements-and-limitations.md):

> The two before & after tables must share some UNIQUE KEY. Such key would be used by gh-ost to iterate the table.

The issue is, is this particular alter statement still okay, given there's technically no shared unique key between the two tables (since it's "renamed" from PK to unique)?

If this is determined to be okay, it might be worth clarifying the docs, because they give an example of a unique key becoming a PK (requiring two migrations) but not the other direction:

> As an example, if your table has a single UNIQUE KEY and no PRIMARY KEY, and you wish to replace it with a PRIMARY KEY, you will need two migrations: one to add the PRIMARY KEY (this migration will use the existing UNIQUE KEY), another to drop the now redundant UNIQUE KEY (this migration will use the PRIMARY KEY).

shlomi-noach:
Fixed by #358

PR #358, opened 2017-01-10
Testing and clarifying shared unique key requirement

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/357

This PR adds the tests to confirm expected behavior of `gh-ost`, when requiring shared unique key between original/ghost tables.

To be followed up by a documentation update.

cc @ggunson


doc/requirements-and-limitations.md
```diff
@@ -31,7 +31,9 @@ The `SUPER` privilege is required for `STOP SLAVE`, `START SLAVE` operations. Th
 - MySQL 5.7 `JSON` columns are not supported. They are likely to be supported shortly.
 
 - The two _before_ & _after_ tables must share some `UNIQUE KEY`. Such key would be used by `gh-ost` to iterate the table.
-  - As an example, if your table has a single `UNIQUE KEY` and no `PRIMARY KEY`, and you wish to replace it with a `PRIMARY KEY`, you will need two migrations: one to add the `PRIMARY KEY` (this migration will use the existing `UNIQUE KEY`), another to drop the now redundant `UNIQUE KEY` (this migration will use the `PRIMARY KEY`).
+  - What matters is not the key's _name_, but the columns covered by such key.
+  - As an example, if your table has a single `UNIQUE KEY my_unique_key(list, of, columns)` and no `PRIMARY KEY`, and you wish to replace the unique key with a `PRIMARY KEY`, you are good to go with a single migration: `--alter='DROP KEY my_unique_key, ADD PRIMARY KEY (list, of, columns)'`
```
ggunson: Depending on the outcome of testing, you should add what happens if you replace a PK (since the issue that sparked this got the PK demoted to unique and replaced with a different key). E.g. is it ok to switch out the PK if the original PK becomes a unique key? (Not necessarily obvious).

doc/requirements-and-limitations.md
```diff
@@ -30,8 +30,7 @@ The `SUPER` privilege is required for `STOP SLAVE`, `START SLAVE` operations. Th
 
 - MySQL 5.7 `JSON` columns are not supported. They are likely to be supported shortly.
 
-- The two _before_ & _after_ tables must share some `UNIQUE KEY`. Such key would be used by `gh-ost` to iterate the table.
-  - As an example, if your table has a single `UNIQUE KEY` and no `PRIMARY KEY`, and you wish to replace it with a `PRIMARY KEY`, you will need two migrations: one to add the `PRIMARY KEY` (this migration will use the existing `UNIQUE KEY`), another to drop the now redundant `UNIQUE KEY` (this migration will use the `PRIMARY KEY`).
+- The two _before_ & _after_ tables must share some `UNIQUE KEY`. Such key would be used by `gh-ost` to iterate the table. See [Read more](shared-key.md)
 
 - The chosen migration key must not include columns with `NULL` values.
```
ggunson: I feel this is a sub-note of the previous note... It is about the migration key.

I might even want to say "the chosen migration key must include only `not null` columns". "columns with null values" could be more vague than referring to the column description being `NOT NULL`?

shlomi-noach: > I might even want to say "the chosen migration key must include only not null columns".

However that is not the case. By default `gh-ost` does indeed only look for Uinique Keys with `not null` columns, but supports `null`-able columns when the user takes responsibility and provides with `--allow-nullable-unique-key`.

For the migration to run correctly there must be no `null` values in the chosen key. `gh-ost` utilizes table definition when possible to _know_ that is not the case, or trusts the user when this is not the case.

ggunson: Okay. I think this is really wordy, though, because it confused me. Since we're doing things in note form, maybe something like:

- The migration key must not include columns with `NULL` values. This means either:
  1. The columns are `NOT NULL`, or
  2. The columns are nullable but don't contain any NULL values. 

- `gh-ost` will try to pick a migration key with non-nullable columns. It will not run if the only `UNIQUE KEY` includes nullable columns. 
  - You may override this via `--allow-nullable-unique-key` but **make sure** there are no actual `NULL` values in those columns. Existing `NULL` values can't guarantee data integrity on the migrated table.

shlomi-noach: This is great wording and better than I can express myself. Would you mind pushing this into this branch?

Also, I don't know if you've seen the new file, [shared-key.md](https://github.com/github/gh-ost/pull/358/files#diff-a3f29a9d676d9f2703ccd894b444ae48)? It really elaborates on the topic. I'm thinking we can just drop the entire paragraph from the `requirements` page altogether, given that it links to that page, and add your description in the `shared-key.md` file?

ggunson: Yeah, I saw shared-key.md too... The shared unique key and non-null values should be in the main page, I think. Can look at a moving around of text when it's not my bedtime.

ggunson: I made an edit but maybe it's still too wordy and I should have edited shared-key.md.

ggunson:
Ok to merge for now.

Issue #361, opened 2017-01-28
Throttler sticks when changing throttle-control-replicas during lag


jonahberquist:
Today we changed `throttle-control-replicas` during a migration to remove a host from service.

This worked fine. Then I commanded the socket `throttle-control-replicas` with no arguments, because I couldn't remember if that would return the current list of throttle control replicas. (For the record, it does not. It clears the list, unless you hit this bug.)

After doing that, I issued a `status`, and everything seemed fine. The migration was throttled, seeing one replica had `0.700882s` of lag, which I interpreted as perfectly normal behavior, and also figured that with no arguments `throttle-control-replicas` just didn't do anything.

The next several migration status outputs in chat displayed exactly the same throttled state. The same replica, the same lag, no progress was being made. Then we noticed. We reset throttle-control-replicas to its previous list (minus the one we removed) and everything proceeded normally.

I was able to reproduce this behavior by manually stopping replication on a control replica and then issuing a `throttle-control-replicas` command. The migration gets stuck in a throttled state. It does not reproduce when there is not lag.

shlomi-noach:
Addressed by https://github.com/github/gh-ost/pull/364. Thank you @jonahberquist and @tomkrouper for analyzing the problem and actively reproducing it (as seen in internal logs)!

PR #364, opened 2017-01-29
fixing phantom throttle-control-replicas lag result

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/361

A bug was introduced via https://github.com/github/gh-ost/commit/e900dae2e99cb7150cf669058da5b6060c2f03cf and
https://github.com/github/gh-ost/commit/fc831b0548b33f62d34b6b6aadff0614507117c8, when moving into the new more-async throttling logic.

With this bug, the throttling check for throttle-control-replicas lag would consistently return some phantom value even if the throttle control list is empty. More specifically, it would retain the result of the last known throttle-control-replicas list before changing to an empty list.

With this fix an empty throttle control list always returns a "no lag" result.


shlomi-noach:
This will only be tested in production towards end of this week or beginning of next week

PR #347, opened 2016-12-26
Reading replication lag via _changelog_ table, also on control replicas

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/346

This PR removes the use of `--replication-lag-query` and uniformly reads replication lag via _changelog_ table, base on `gh-ost`'s internal heartbeat mechanism -- also on the throttle control replicas.

PR #365, opened 2017-02-02
testing/running on replica: gets lags via SHOW SLAVE STATUS

shlomi-noach:
Followup on https://github.com/github/gh-ost/pull/347

on `test-on-replica` or `migrate-on-replica`, relying on the internal changelog heartbeat is incorrect behavior and leads to lags. This is because is those cases the heartbeat is injected onto the replica in the first place. So clearly as we read it it will be happy and up-to-date. But this means we're overloading the replica with our migration, causing it to lag.

With this PR, on `test-on-replica` or `migrate-on-replica`, we read `SHOW SLAVE STATUS` of the replica. This will have 1-second granularity but this is fine for running on replica.

### DIFF
--- a/go/logic/throttler.go
+++ b/go/logic/throttler.go
@@ -84,18 +84,30 @@ func (this *Throttler) parseChangelogHeartbeat(heartbeatValue string) (err error
 	}
 }
 
-// collectHeartbeat reads the latest changelog heartbeat value
-func (this *Throttler) collectHeartbeat() {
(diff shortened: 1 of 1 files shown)


shlomi-noach:
Note that it is still possible to use `throttle-control-replicas` on `test-on-replica` and `migrate-on-replica`, and when these are provided, we expect to read internal heartbeat. These would be descendants of the migrated replica.

PR #366, opened 2017-02-05
Documentation: common questions

shlomi-noach:
Adding doc page for common questions: "how?", "why?", "Is it possible to...?"

cc @github/database-infrastructure @ikewalker

### DIFF
--- a/README.md
+++ b/README.md
@@ -58,9 +58,10 @@ More tips:
 Also see:
 
 - [requirements and limitations](doc/requirements-and-limitations.md)
+- [common questions](doc/questions.md)
 - [what if?](doc/what-if.md)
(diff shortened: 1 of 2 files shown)


ggunson:
>  This behavior may be supported in the future.

From the MySQL 5.7 docs:

> As of MySQL 5.7.4, the IGNORE clause for ALTER TABLE is removed and its use produces an error.

Given it's deprecated, it might not be worth saying it may be supported later (or pointing out that it's deprecated, at least).

shlomi-noach:
> Given it's deprecated, it might not be worth saying it may be supported later (or pointing out that it's deprecated, at least).

Have quoted the deprecation clause and wrote we're unlikely to support it.

PR #370, opened 2017-02-07
Migration only starting after first replication lag metric collected

shlomi-noach:
This PR postpones beginning of actual migration until replication lag for the inspected server (the `--host` param) has been determined.
Migration already waits today until first throttle metrics are present, such as `--max-load` values etc. This PR merely adds waiting for the replication lag info.


shlomi-noach:
sent for testing

shlomi-noach:
Still can return untrue results because `SHOW SLAVE STATUS` can return `Seconds_behind_master: 0` even when lagging (because life).
But this would only affect tiny migrations. It's good enough to go.

Issue #280, opened 2016-10-19
improve output without --verbose option


groenewold-pythian:
If gh-ost is run without the `--verbose` option like the example below, we don't get any information about the process completion. Would be nice to have a one-liner saying process complete or something like that

```
[vagrant@mysqlsandbox master]$ gh-ost --user=msandbox --password=msandbox --host=localhost --table=t --database=test --alter='engine=innodb' --port=22896 --allow-on-master
# Migrating `test`.`t`; Ghost table is `test`.`_t_gho`
# Migrating mysqlsandbox.dev:22896; inspecting mysqlsandbox.dev:22896; executing on mysqlsandbox.dev
# Migration started at Wed Oct 19 14:19:08 +0000 2016
# chunk-size: 1000; max-lag-millis: 1500ms; max-load: ; critical-load: ; nice-ratio: 0.000000
# throttle-additional-flag-file: /tmp/gh-ost.throttle
# Serving on unix socket: /tmp/gh-ost.test.t.sock
# Migrating `test`.`t`; Ghost table is `test`.`_t_gho`
# Migrating mysqlsandbox.dev:22896; inspecting mysqlsandbox.dev:22896; executing on mysqlsandbox.dev
# Migration started at Wed Oct 19 14:19:08 +0000 2016
# chunk-size: 1000; max-lag-millis: 1500ms; max-load: ; critical-load: ; nice-ratio: 0.000000
# throttle-additional-flag-file: /tmp/gh-ost.throttle
# Serving on unix socket: /tmp/gh-ost.test.t.sock
Copy: 0/0 100.0%; Applied: 0; Backlog: 0/100; Time: 0s(total), 0s(copy); streamer: mysql-bin.000002:1651; State: migrating; ETA: due
CREATE TABLE `_t_gho` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `created` datetime DEFAULT NULL,
  `job_id` bigint(20) NOT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB DEFAULT CHARSET=latin1
Copy: 0/552 0.0%; Applied: 0; Backlog: 0/100; Time: 0s(total), 0s(copy); streamer: mysql-bin.000002:1651; State: migrating; ETA: N/A
```

For comparison here is the output of the same command with the `--verbose` option

```
[...]

PR #373, opened 2017-02-07
improve output without --verbose option

groenewold-pythian:
Related issue: #280 

### Description

This PR tries to improve visibility of the tool when running without the `--verbose` option by printing the "# Done" message to stdout rather than through the logging facility.


shlomi-noach:
Thank you!

shlomi-noach:
For next PR, please review again that "- [ ] code is formatted via `gofmt` (please avoid `goimports`)
"
otherwise this makes tests fail. Thank you!

groenewold-pythian:
Thanks! will do

PR #376, opened 2017-02-21
supporting --timestamp-old-table

shlomi-noach:
A new flag: `--timestamp-old-table` makes the _old_ table name include a timestamp. This makes the _old_ table unique and non-repeating.

This could either be nice for operations (non conflicting names), or not nice for operations (name changes all the time). This why it's flaggable.

Example:

Assuming table is `gh_ost_test`,
- old table name with `--timestamp-old-table=false` would be `_gh_ost_test_del`
- old table name with `--timestamp-old-table=true`would be `_gh_ost_test_20170221103147_del`


shlomi-noach:
The timestamp format is `YYYYMMDDHHmmSS` (year, month, day, hour, minute, seconds)

tomkrouper:
You'll want to be careful on max table_name size (I believe 64 bytes)

shlomi-noach:
@tomkrouper code now verifies resulting table name is within allowed length.

PR #386, opened 2017-03-26
supporting throttle-http

shlomi-noach:
A new flag, `--throttle-http=<url>` provides with new means for throttling.

If given, `gh-ost` will routinely poll given URL via `HEAD` request. Any response code that is not `200` is interpreted as a reason for throttling.

If given, the server behind the URL should provide with good response times.

This is compatible with https://github.com/github/freno. At this time https://github.com/github/freno is a private repo, but will turn public in the future.

cc @github/database-infrastructure @miguelff


shlomi-noach:
- [ ] todo: add documentation for `throttle-http`

PR #387, opened 2017-03-26
compound pk tests

shlomi-noach:
Adding a couple interesting tests:

- a compound `PRIMARY KEY` on `int auto_increment`, `varchar`
- a compound `PRIMARY KEY` on `int auto_increment`, `timestamp`

test script improves visibility on checksum failure.


demianstack6 (approved):

localtests/test.sh
```diff
@@ -224,6 +224,8 @@ test_single() {
   ghost_checksum=$(cat $ghost_content_output_file | md5sum)
 
   if [ "$orig_checksum" != "$ghost_checksum" ] ; then
+    gh-ost-test-mysql-replica --default-character-set=utf8mb4 test -e "select ${orig_columns} from gh_ost_test" -ss > $orig_content_output_file
```
timvaillancourt, line 227: @shlomi-noach there are almost identical lines above these 2 x [at lines  221 and 222](https://github.com/github/gh-ost/blob/a2dbb3fb1f06e6cfa42fffcbac0646dbebce7559/localtests/test.sh#L221-L222)

Are these 2 x new lines necessary? 🙇

shlomi-noach: The former two lines are used to diff the two tables according to explicitly specified order, which is test dependent. The checksum is taken according to that order.
The latter two lines are used to present the table "as is" to the engineer, in the event there's a checksum failure.

I think it makes sense to keep it this way.

timvaillancourt: @shlomi-noach 👋 ahh, thanks for the context! Makes sense 👍

Issue #388, opened 2017-03-27
gh-ost migration: [error] close sync with err: sync is been closing...


ChinaVogel:
Hello gh-ost team,
I get this error when running a migration with gh-ost:
binlogstreamer.go:47: [error] close sync with err: sync is been closing...

the migration seems to run fine (at least the table got changed ;) and all the content is there)
But this such an error make me feel uncomfortable using it on production.

System: macOS Sierra 10.12.3
Installed: just running the downloaded executable: 
```
> /gh-ost --version
1.0.35
```

DB System:
```
> mysql --version
mysql Ver 15.1 Distrib 10.2.4-MariaDB, for osx10.12 (x86_64) using readline 5.1
```

Is this more a warning? I did not look into source code of gh-ost so far. I got this message also for other fields and other migrations.

regards

Here the whole verbose output (it is a test run with root on a dev machine, that is why I use high secure no pwd root... )
```
./gh-ost --database="excelsior_development" --table="attachments" --user="root" --alter="MODIFY filesize BIGINT DEFAULT 0" --switch-to-rbr --allow-on-master --initially-drop-ghost-table --verbose --execute
2017-03-27 11:15:24 INFO starting gh-ost 1.0.35
2017-03-27 11:15:24 INFO Migrating `excelsior_development`.`attachments`
2017-03-27 11:15:24 INFO connection validated on 127.0.0.1:3306
2017-03-27 11:15:24 INFO User has ALL privileges
2017-03-27 11:15:24 INFO binary logs validated on 127.0.0.1:3306
2017-03-27 11:15:24 INFO Restarting replication on 127.0.0.1:3306 to make sure binlog settings apply to replication thread
2017-03-27 11:15:24 INFO Inspector initiated on PVGM50860507A:3306, version 10.2.4-MariaDB-log
2017-03-27 11:15:24 INFO Table found. Engine=InnoDB
2017-03-27 11:15:24 INFO Estimated number of rows via EXPLAIN: 17
2017-03-27 11:15:24 INFO Recursively searching for replication master
2017-03-27 11:15:24 INFO Master found to be PVGM50860507A:3306
2017-03-27 11:15:24 WARNING log_slave_updates not found on 127.0.0.1:3306, but executing directly on master, so I'm proceeeding
2017-03-27 11:15:24 INFO connection validated on 127.0.0.1:3306
[...]

Issue #383, opened 2017-03-15
binlogstreamer.go "close sync with err" on Tungsten cluster


johnz1:
gh-ost 1.0.35
MySQL 5.6.23
Tungsten 4.0.7

Hello

I'm running gh-ost against the master of a Tungsten 4.0.7 cluster.  I'm trying to drop a column, but it's giving the following error for binlogstreamer:

`binlogstreamer.go:47: [error] close sync with err: sync is been closing...`

Here's the full output:

```
2017-03-15 16:11:20 INFO starting gh-ost 1.0.35
2017-03-15 16:11:20 INFO Migrating `db`.`tbl`
2017-03-15 16:11:20 INFO connection validated on testhost:13306
2017-03-15 16:11:20 INFO User has ALL privileges
2017-03-15 16:11:20 INFO binary logs validated on testhost:13306
2017-03-15 16:11:20 INFO Restarting replication on testhost:13306 to make sure binlog settings apply to replication thread
2017-03-15 16:11:20 INFO Inspector initiated on testhost:13306, version 5.6.23-enterprise-commercial-advanced-log
2017-03-15 16:11:20 INFO Table found. Engine=InnoDB
2017-03-15 16:11:20 INFO Estimated number of rows via EXPLAIN: 10
2017-03-15 16:11:20 INFO Master forced to be testhost:13306
2017-03-15 16:11:20 WARNING log_slave_updates not found on testhost:13306, but --tungsten provided, so I'm proceeding
2017-03-15 16:11:20 INFO connection validated on testhost:13306
2017/03/15 16:11:20 binlogsyncer.go:75: [info] create BinlogSyncer with config &{99999 mysql testhost 13306 mwadmin passwd  false false <nil>}
2017-03-15 16:11:20 INFO Connecting binlog streamer at mysql-bin.000044:13138
2017/03/15 16:11:20 binlogsyncer.go:241: [info] begin to sync binlog from position (mysql-bin.000044, 13138)
2017/03/15 16:11:20 binlogsyncer.go:134: [info] register slave for master server testhost:13306
2017/03/15 16:11:20 binlogsyncer.go:568: [info] rotate to (mysql-bin.000044, 13138)
2017-03-15 16:11:20 INFO rotate to next log name: mysql-bin.000044
2017-03-15 16:11:20 INFO connection validated on testhost:13306
2017-03-15 16:11:20 INFO connection validated on testhost:13306
[...]

PR #389, opened 2017-03-27
fixing binlog syncer double-close()

shlomi-noach:
Storyline:

- https://github.com/github/gh-ost/issues/383
- https://github.com/github/gh-ost/issues/388

This PR avoids closing `go-mysql`'s binlog syncer `close()` twice. Well, `gh-ost` only called it _once_, but following an upgrade binlog-syncer takes care of itself and `gh-ost` is no longer required to `close()`


shlomi-noach:
Deploys and tests well in production

PR #407, opened 2017-04-24
Add `onTestReplicaComplete` hook

### LINKED ISSUE #406: Add `ghost-on-test-on-replica-complete` hook
As I've been setting up continuously running migrations, I've managed to accumulate a few scripts that I have been running following  the completion of a `--test-on-replica` run that I think are worth automating. Here are some of the actions I've been taking:

- Run `pt-table-checksum` against the two tables, post the row and diff results to our metrics collector
- Run MySQL checksums against the two tables, post the results checksum results to our metrics collector
- Update the originating pull request with the results (tables, checksum output, run time)
- Snapshotting replica lag during the migration from our metrics collector
- Post results to our chat to inform everyone that everything is 👌  and isn't on 🔥 

Being the fan of automation that I am, I checked the [hook definitions](https://github.com/github/gh-ost/blob/master/go/logic/hooks.go#L20-L34) however I couldn't find anything that was appropriate for this use case. The [gh-ost-on-before-cut-over](https://github.com/github/gh-ost/blob/master/go/logic/hooks.go#L27) is close however I'm only interested in doing this for test on replica runs at the moment - not all of them.

Is this something you'd be looking to add to `gh-ost` hooks? I'm happy to open a PR with the changes but just wanted to check and make sure I wasn't missing anything obvious before spending time on it.

jacobbednarz:
When doing `--test-on-replica` runs, it's common to want to perform some
sort of integrity checks against the data and schema of your tables. To
cater for this in a more automated means, this adds a new hook
`onTestReplicaComplete` which is fired once the migration is completed
and cutover but before replication is restarted to ensure consistency
with rows.

Closes #406


go/logic/migrator.go
```diff
@@ -396,6 +396,10 @@ func (this *Migrator) handleCutOverResult(cutOverError error) (err error) {
 	// Only on error:
 
 	if this.migrationContext.TestOnReplica {
+		if err := this.hooksExecutor.onTestReplicaComplete(); err != nil {
```
jacobbednarz: Is this the correct spot for it? I am aiming to have this run after replication has stopped, cutover is performed but before the replication is started again. I _think_ this is where it should be however I'd like some more 👀  in case I've buggered something up here.

jacobbednarz:
👉 @shlomi-noach

go/logic/migrator.go
shlomi-noach: Whoops missed this one. Is this the correct spot? It greatly depends on what you want to achieve! 
The name `onTestReplicaComplete` is certainly not fitting to this place; the migration is not complete.

In fact, when the test-on-replica is complete, `gh-ost` just bails out without restarting the replication. So I do believe the normal "migration complete" hook should just work well for you. We use it that way, and just pass completely different hooks to the test-on-replica invocation.

jacobbednarz: My apologies! I thought #406 and the description conveyed what I was after however it doesn't appear I did 😞  The intention was to have a hook that could be used with `--test-on-replica` runs and would run before the cut over would happen (and not on all migrations, just the test ones) to do things like integrity checks + checksumming whilst the replication was stopped and not yet complete.

> We use it that way, and just pass completely different hooks to the test-on-replica invocation.

I did explore this but could have been doing it wrong. What do you use to differentiate between the two types of migrations? Would you mind sharing a simple example of how you're achieving this?

shlomi-noach:
@jacobbednarz we use different `--hooks-path` for production migrations and for test-migrations.

Here are the hooks we have:

screen shot 2017-06-13 at 14 50 36

for test-on-replica, we're less in need of hooks because once `gh-ost` terminates the two tables are in place, replication is stopped. We don't need a hook, we just wait for process to complete. Then our script drops the _ghost_ table and starts replication.

jacobbednarz:
Ah, I remember why this was differing from the `cut-over` hook. I was looking to only run some actions when `--test-on-replica` was used. Your idea of different hook paths might also be enough for this.

shlomi-noach:
I suspect the different hook path will indeed solve your problem; but please let me know if you hit something different.

jacobbednarz:
Going with your option of multiple paths, thanks for the idea!

PR #409, opened 2017-04-26
[WIP] Add `--cleanup` flag

### LINKED ISSUE #96: Support --cleanup flag
The `--cleanup` flag is mutually exclusive to `--alter`. It indicates we wish to clean up _ghost_ and _changelog_ tables.

Like `--alter`, it would default to _noop_ (in which case it should print the `drop` statements), and would support `--execute`

jacobbednarz:
If a migration is interrupted or told to exit via a panic, the `gh-ost`
migration and changelog tables will be left behind and no cleanup will be
performed.

This updates `gh-ost` to support a `--cleanup` option which will go ahead
and remove these safely.

Fixes #96


jacobbednarz:
Swapped the title to include WIP since I can't add a label to do it.

jacobbednarz:
I've come back to this (finally, I know!) and I'm a bit torn on where this cleanup would take place. Would you agree this should live in [`finalCleanup()`](https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L1190`) just with some redefined logic and the existing cleanup tasks?

shlomi-noach:
> Would you agree this should live in finalCleanup() just with some redefined logic and the existing cleanup tasks?

I'm ambivalent. Because `finalCleanup()` assumes a migration has been executed. This, it prints out the `show create table` and has some considerations for `TestOnReplica`.

However in our case what we really want is to just forcibly drop the ghost and changelog table.

I'd say create a new function to do that.

jacobbednarz:
Closing this for now since I don't have the time to focus on it.

Issue #414, opened 2017-05-05
postpone-cut-over-flag-file is useless without a touched file


et:
Hi all, thank you for building such a great tool.

I just wanted to provide some feedback on the `postpone-cut-over-flag-file` where I was simply copying a call to `gh-ost` with this option set as `--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag`.
I ran a 36 hour migration and `gh-ost` tried to do the cut over before timing out due too much traffic.

Unfortunately, I neglected to `touch /tmp/ghost.postpone.flag` before starting `gh-ost`. Hence, I need to re-run the migration from scratch -- of course remembering to `touch` that file first so I can manually handle the cut over during low traffic.

Is there a particular reason why if this option is set to something, `gh-ost` can't just `touch` the file automatically?

shlomi-noach:
@et thank you. There is a (potentially minor?) consideration for what happens if `gh-ost` is unable to touch the file due to permission limitations. Should it bail out?

Otherwise I think this is some legacy consideration. I think we should be good to have `gh-ost` touch the file upon startup. It makes sense.

shlomi-noach:
Addressed by #415 
I'm not sure how I feel about this at this time.

Just to share how we do it: we always indicate a `--postpone-cut-over-flag-file=/etc/some-constant-file.flag`. Our `puppet` setup ensures the file exists. That is, we _always_ postpone the cutover. We don't need `gh-ost` to create the file because it's always there. We cut-over not by deleting the file, but by issuing a `cut-over` command.

I can now see that #415 can affect tests and automation built around `gh-ost`. I'll think about it some more.

et:
Thanks a ton for the feedback @shlomi-noach.
It definitely was a small "gotcha" that I failed to notice.

> I can now see that #415 can affect tests and automation built around gh-ost. I'll think about it some more.

If this seems unreasonable, how about bailing out if that option is specified but the file doesn't exist? Seems better to bail now rather than days later 😉

et:
@shlomi-noach - one more suggestion. What about a new flag called `--postpone-cut-over` that postpones the cut over regardless of the presence of a file?

shlomi-noach:
@et thank you and sorry for the delay; had a lot on my plate. `--postpone-cut-over` is a good suggestion. There's other options, too; let me pick one and will make most sense.

groenewold-pythian:
We were also bit by this one. Wondering if you reached out any conclusion yet?

[...]

PR #415, opened 2017-05-07
-postpone-cut-over-flag-file implies touching indicated file

shlomi-noach:
fixes https://github.com/github/gh-ost/issues/414

With this PR, when `--postpone-cut-over-flag-file=...` is given, `gh-ost` auto-creates the file (or, if the file already exists, does nothing).


shlomi-noach:
This will be sent to testing ; there's a queue right now; I'd expect to roll this out in two weeks or so. Sorry for this late submission, cc @et @groenewold-pythian

shlomi-noach:
Sent to testing

Issue #418, opened 2017-05-17
Demote --critical-load from exiting process to pausing it


ggunson:
I discussed this with @shlomi-noach on Monday: 

`--critical-load` is currently a copy of a flag from pt-online-schema-change: in the event that that value is hit (e.g. high threads_running), gh-ost panics and dies. 

This behaviour makes sense in pt-online-schema-change because it has triggers on the master/host that copy writes from the old to new table, so on things going badly, you want those triggers removed since they impede performance.

With gh-ost, however, if its copy thread is paused, there's no extra load on the master (the gh-ost tables are there, but no write activity is going to them). So pausing gh-ost for a period of time looks the same to the master as gh-ost not running at all.

On that note, I suggest a new feature, where `--critical-load` is treated as a more severe `--max-load`, where gh-ost is paused for, say, an hour (or some other default value). This is on the idea that whatever bad thing just happened is temporary and/or fixable and we'd rather not lose hours/days of table copying because of it. And if things were fixed before that hour is up, gh-ost could be unpaused.

shlomi-noach:
We can introduce this new behavior without breaking backwards compatibility by adding some `--critical-load-sleep-seconds`

- would default `0`
- if nonzero, then it overrides the `panic` executed today on `critical-load`, and instead go to sleep. During that sleep, `gh-ost` will not issue _any queries_. At the end of the sleep, `gh-ost` resumes work as usual (and of course may hit `critical-load` again and go to sleep again).

tomkrouper:
When I first read the title of this, I was thinking... I don't like it. Then I read the explanation, and I'm like :+1: :+1:. I like the idea and the implementation idea.

vjanelle:
Instead of a specific amount of time, you could also implement an exponential backoff.

shlomi-noach:
> Instead of a specific amount of time, you could also implement an exponential backoff.

that could work well, too.

shlomi-noach:
For now, #424 is merged with a CLI flag `--critical-load-hibernate-seconds`, and without exponential backoff. We will be running this in production for some time before re-considering whether we need a more elaborate backoff or whether the pre-set hibernation time is good enough.

PR #424, opened 2017-05-24
hibernate on critical-load

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/418

This PR introduces `--critical-load-hibernate-seconds`. This flag defaults `0` (in which case the new feature introduced in this PR is disabled)

when `critical-load-hibernate-seconds > 0`, then the behavior of `critical-load` changes. Instead of panicking and bailing out with error, `gh-ost` enters hibernation.

As opposed to throttling, hibernation implies absolutely no queries on neither master not migrated server. No heartbeats; no throttle checks; no replication lag checks; nothing.

When `gh-ost` awakes from hibernation, it's all back to normal, and throttling checks take place as usual, and of course it may potentially re-enter hibernation if `critical-load` is still met.

cc @ggunson @github/database-infrastructure

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -40,8 +40,9 @@ const (
 type ThrottleReasonHint string
 
 const (
-	NoThrottleReasonHint          ThrottleReasonHint = "NoThrottleReasonHint"
-	UserCommandThrottleReasonHint                    = "UserCommandThrottleReasonHint"
+	NoThrottleReasonHint                 ThrottleReasonHint = "NoThrottleReasonHint"
+	UserCommandThrottleReasonHint                           = "UserCommandThrottleReasonHint"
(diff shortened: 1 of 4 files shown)


shlomi-noach:
TODO: on leaving hibernation, `gh-ost` has no up-to-date metrics. It shouldn't just resume work; it must first run a collection of throttle-metrics, inject heartbeat etc. Only then should it resume work.

This is somewhat complicated to achieve and supervise, since most of these processes are independent.

shlomi-noach:
I believe https://github.com/github/gh-ost/pull/424/files/3955a6d67f80404d1f2611b82023ba4daeda423c..ad47f7c1477780020631627514d3ef168e8f71ff#diff-6298c3b05013fa1f1cc736d61051b34eR292 should do well.

PR #426, opened 2017-06-05
Update to state RDS works, but has it's own limitations

groodt:
clarifying RDS & Google cloud support


shlomi-noach:
thank you!

groodt:
No problem.

PR #427, opened 2017-06-05
[WIP] Localtests to run in CI and possibly locally

shuhaowu:
This is supposed to be a PR that makes it possible to run localtests in CI and locally. However, I'm tagging it a WIP until it actually works, as I have encountered several issues that I don't quite know how to resolve. 

### How it works

The previous CI script simply installs Go, builds the gh-ost binary, and then runs `go test`, which AFAICT does not cover a large portion of the code base. 

The new CI setup works by performing several steps:

1. Installs Percona server 5.7 (we can change the flavour and version here if we want, but travis uses 5.5 and that's quite old).
2. Sets up two instances of MySQL running from the travis user and set replication up between them.
3. Builds the gh-ost binary with the Go provided by travis so we don't need to download it again.
4. Run the Go unittests.
5. Run the localtests.
6. Shutdown mysql as otherwise the build will hang as Travis detect that we still have processes running.

The build step (`02-install-ghost.sh`) and the localtest step (`04-run-localtests.sh`) are currently a modified version `script/build` and `localtests/test.sh`. 

### Issues

I noticed that when running the localtests, I see in the logs that gh-ost produces in stdout that seems to indicate it has errored:

```
2017-06-05 13:57:22 ERROR replication not running; Slave_IO_Running=No, Slave_SQL_Running=No
goroutine 132 [running]:
runtime/debug.Stack(0x5c, 0x42, 0x0)
        /home/travis/.gimme/versions/go1.7.linux.amd64/src/runtime/debug/stack.go:24 +0x79
runtime/debug.PrintStack()
        /home/travis/.gimme/versions/go1.7.linux.amd64/src/runtime/debug/stack.go:16 +0x22
github.com/github/gh-ost/vendor/github.com/outbrain/golib/log.logErrorEntry(0x2, 0x8cac60, 0xc4201e7730, 0x5015a0, 0xc4203b2f58)
        /home/travis/gh-ost/.gopath/src/github.com/github/gh-ost/vendor/github.com/outbrain/golib/log/log.go:178 +0x10b
github.com/github/gh-ost/vendor/github.com/outbrain/golib/log.Errore(0x8cac60, 0xc4201e7730, 0x8cac60, 0xc4201e7730)
        /home/travis/gh-ost/.gopath/src/github.com/github/gh-ost/vendor/github.com/outbrain/golib/log/log.go:224 +0x3e
github.com/github/gh-ost/go/logic.(*Throttler).collectReplicationLag.func1(0xc4201d2f78, 0x1)
        /home/travis/gh-ost/.gopath/src/github.com/github/gh-ost/go/logic/throttler.go:105 +0xd1
created by github.com/github/gh-ost/go/logic.(*Throttler).collectReplicationLag
        /home/travis/gh-ost/.gopath/src/github.com/github/gh-ost/go/logic/throttler.go:124 +0x129
``` 

In fact, you can see this error yourself at: https://travis-ci.org/shuhaowu/gh-ost/builds/238507173

However, when I examined the actual mysql databases, I can clearly see that replication is setup:

```
mysql> show slave status\G;
*************************** 1. row ***************************
               Slave_IO_State: Waiting for master to send event
                  Master_Host: 127.0.0.1
                  Master_User: repl
                  Master_Port: 20001
                Connect_Retry: 60
              Master_Log_File: mysqld-bin.000002
          Read_Master_Log_Pos: 1651
               Relay_Log_File: ghostci-relay-bin.000010
                Relay_Log_Pos: 457
        Relay_Master_Log_File: mysqld-bin.000002
             Slave_IO_Running: Yes
            Slave_SQL_Running: Yes
       .....
```

Despite the error in the log, the test is marked as passed. I'm not quite sure what is causing this at the moment, perhaps someone can chime in here?

-----------------------

The other issue that I'm having is that this script setup is not very usable for the local environment without some more manual steps. I can confirm that you can spin up a ubuntu precise box and manually run these scripts in the order (with some additional commands with respect to installing mysql and go). 

In fact, I have a set of scripts that spins up a ubuntu precise LXC and run the travis tests. However, I'm not sure if I should include these scripts to provide a way to run these tests locally as everyone's environment is so different and the scripts are fairly hacky.


.travis.yml
```diff
@@ -1,20 +1,35 @@
 # http://docs.travis-ci.com/user/languages/go/
 language: go
 
-go: 1.8
+go: 1.7
```
shuhaowu: I noticed that we only use go1.7 as we fetch go ourselves, so updated this to match.

travis/04-run-localtests.sh
```diff
@@ -0,0 +1,137 @@
...
+    --verbose \
+    --debug \
+    --stack \
+    --execute ${extra_args[@]}"
+  echo_dot
+  echo $cmd > $exec_command_file
+  echo_dot
+  bash $exec_command_file 1> $test_logfile 2>&1
+
+  execution_result=$?
```
shuhaowu: This is 0 even tho we have the error from the replication setup in another goroutine.

travis/04-run-localtests.sh
```diff
@@ -0,0 +1,137 @@
...
+      echo "+ FAIL"
+      return 1
+    else
+      echo
+      echo "+ pass"
+    fi
+    $SLAVE_MYSQL -e "start slave"
+    echo "LOG FOR $test_name"
+    cat $test_logfile
+    break
```
shuhaowu: I'm breaking this for now because every test case contains the error saying replication is not setup. Otherwise we get spammed. Alternatively if I hide the logs it will just say everything passed.

.travis.yml
shlomi-noach: I might experiment with go1.8 plugin, in which case this would _need_ to be `1.8` and not `1.7`. This work is not imminent, so this is a heads up.

laceypm: What is it for?

shlomi-noach: @laceypm please explain your question, since otherwise I expect to mark you as a pammy user.

timvaillancourt:
@shuhaowu since this PR was created we discontinued the use of Travis CI and moved to GitHub Actions for CI tasks

Unfortunately this PR cannot be merged because this would undo that progress. We are happy to accept a PR that would adapt these changes with GitHub Actions, however. I apologize for the late/frustrating reply! 😞

PR #440, opened 2017-07-05
freno-aware http-throttler

shlomi-noach:
For convenience purposes, `gh-ost` heuristically guesses whether the `throttle-http` service is in fact [freno](https://github.com/github/freno)

If so, the throttle messages turn to be more user friendly. Instead of `throttled: http=417`, we print `throttled: freno: threshold exceeded`

Even if it isn't a `freno` throttler, we print the textual representation of status codes. E.g. `404` is `Not found` etc.

cc @github/database-infrastructure

### DIFF
--- a/go/logic/throttler.go
+++ b/go/logic/throttler.go
@@ -8,6 +8,7 @@ package logic
 import (
 	"fmt"
 	"net/http"
+	"strings"
 	"sync/atomic"
 	"time"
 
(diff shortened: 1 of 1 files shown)


shlomi-noach:
in prod:

```
Copy: 306300/7948470 3.9%; Applied: 52; Backlog: 100/100; Time: 26s(total), 25s(copy); streamer: mysql-bin.012580:763286125; State: throttled, freno: access forbidden (http=417); ETA: 10m26s
```

shlomi-noach:
in prod:
```
Copy: 584500/7948475 7.4%; Applied: 423; Backlog: 0/100; Time: 54s(total), 53s(copy); streamer: mysql-bin.012580:786210546; State: throttled, freno: threshold exceeded (http=429); ETA: 11m7s
```

Issue #444, opened 2017-07-11
the data loss,  a scene


luoxiaocai91:
the  machine that work gh-ost is not ok,always timeout。the gh-ost appear olny copy one half of  the old table , then cut-over. it cause loss data,please analyze the reason
this is the log:
![image](https://user-images.githubusercontent.com/26221539/28049784-2b074850-662c-11e7-93ca-1b6994fc6c0f.png)

 Thank you!

shlomi-noach:
Hi, can you please help me out by:

- attaching the log as text rather than image, and have as much of the log as possible
- `show create table`
- the `gh-ost` command you were running (including all flags)
- MySQL versions

thank you!

shlomi-noach:
In particular, I'm not sure `gh-ost` only copied half of the table because I don't know if you used `--exact-rowcount`. If not, then it relied on the `InnoDB` estimation of row count, where a `50%` off is actually common.

cenkore:
@shlomi-noach I reproduce it , If kill the gh-ost session on the master while gh-ost is running , It maybe occur some unexpected phenomenon . In my test ,at the firstly kill ,it just showed `[MySQL] 2017/07/11 13:45:49 packets.go:32: unexpected EOF` and proceed . Secondly , it showed as follows, and cut-over immediately . Go to cut-over stage directly after `bad connection`, it is so terrible .. 
```
[MySQL] 2017/07/11 13:46:33 packets.go:32: unexpected EOF
[MySQL] 2017/07/11 13:46:33 statement.go:27: Invalid Connection
2017-07-11 13:46:33 ERROR driver: bad connection
2017-07-11 13:46:33 INFO Row copy complete
```  
**Command**
```
gh-ost --host=10.28.77.43 --database=ghostdb --table=test50g --replica-server-id=7777 --port=55944 --alter="add column addcol1s int default 0" --concurrent-rowcount --conf=/usr/local/gh-ost.cnf --throttle-flag-file="/tmp/gh-ost.ghostdb.test50g.throttle" --postpone-cut-over-flag-file="/tmp/gh-ost.ghostdb.test50g.postpone" --hooks-path="/tmp" --max-load="Threads_connected=5000,Threads_running=5000" --initially-drop-old-table --initially-drop-ghost-table --initially-drop-socket-file --nice-ratio=0.2 --verbose --execute
```

**FULL LOG**
```
2017-07-11 13:45:04 INFO starting gh-ost 1.0.36
[...]

PR #445, opened 2017-07-11
responding to errors in rowcopy

shlomi-noach:
context: https://github.com/github/gh-ost/issues/444

not all rowcopy errors were handled. While errors in row copying had retries, the entire process was not reporting back (as it is being executed in a goroutine).

As result, significant errors such as killed connections were wrongly interpreted as "copy complete" leading to missing rows, i.e. **data corruption**.

This PR identifies such errors and bails out the migration altogether (having executed `default-retires` times).


shlomi-noach:
tested in production.

shlomi-noach:

Issue #446, opened 2017-07-12
the gh-ost is case-sensitive,a columns data loss....


luoxiaocai91:
test table ：
```
Create Table: CREATE TABLE `sbtest` (
  `id` int(10) unsigned NOT NULL,
  `k` int(10) unsigned NOT NULL DEFAULT '0',
  `c` char(120) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  `pad` char(60) COLLATE utf8mb4_unicode_ci NOT NULL DEFAULT '',
  PRIMARY KEY (`id`),
  KEY `k` (`k`),
  KEY `idx_c` (`c`(10))
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

```sql
alter table sbtest modify K bigint unsigned NOT NULL DEFAULT '0';
```
the data:
```
select k,count(*)  from sbtest group by k;
+-----+----------+
| k   | count(*) |
+-----+----------+
|   0 |     9900 |
|  10 |      207 |
|  15 |        1 |
|  20 |        5 |
| 100 |        1 |
+-----+----------+
after alter,the data:
select k,count(*)  from sbtest group by k;
+---+----------+
| k | count(*) |
+---+----------+
| 0 |    10114 |
+---+----------+
```
the log:
![image](https://user-images.githubusercontent.com/26221539/28106444-6c3db530-6716-11e7-80c8-fc05e4acb101.png)

it not choice the columns of k,the mysql always ingore case-sensitive.so i want consider this condition.this is a example. we loss data in the product,because of the other people‘s carelessness. pt-online-schema-change is ingore case-sensitive.

the version is "5.6.28-76.1-log Percona Server (GPL), Release 76.1, Revision 5759e76"
the table is common.
command :
```
gh-ost_v1.0.36 
--max-load=Threads_running=32 
--critical-load=Threads_running=32 
--chunk-size=1000 
--throttle-control-replicas="$throttle_control_replicas" 
--max-lag-millis=1000 
--initially-drop-old-table 
--initially-drop-ghost-table 
--ok-to-drop-table 
--conf="***" 
--host="$binlog_slave" 
--port="$port" 
--database="$database" 
--table="$table" 
--debug 
--alter="$sql" 
--switch-to-rbr 
--allow-master-master 
--cut-over=default 
--default-retries=3 
--critical-load-interval-millis=3000 
[...]

PR #447, opened 2017-07-12
handling column name case change

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/446

This PR will fix the scenario where a column is changes its case. Example:
```
create table gh_ost_test (
  id int auto_increment,
  c1 int not null default 0,
  c2 int not null default 0,
  primary key (id)
) auto_increment=1;
```
and
```
--alter="modify C2 int not null default 0"
```

leading to a `C2` column in altered table.

Right now `gh-ost` does not detect `c2` and `C2` are the same.

initial commit: an integration test to catch the error (confirmed)

### DIFF
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -656,18 +656,21 @@ func (this *Inspector) getSharedUniqueKeys(originalUniqueKeys, ghostUniqueKeys [
 
 // getSharedColumns returns the intersection of two lists of columns in same order as the first list
 func (this *Inspector) getSharedColumns(originalColumns, ghostColumns *sql.ColumnList, columnRenameMap map[string]string) (*sql.ColumnList, *sql.ColumnList) {
-	columnsInGhost := make(map[string]bool)
-	for _, ghostColumn := range ghostColumns.Names() {
-		columnsInGhost[ghostColumn] = true
-	}
 	sharedColumnNames := []string{}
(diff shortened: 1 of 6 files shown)


shlomi-noach:
added tests confirms and reproduces the issue; test result is:
```
shlomi-gh ~/dev/go/src/github.com/github/gh-ost$ ./localtests/test.sh ^modify
Building
Testing: modify-change-case......ERROR modify-change-case: checksum mismatch
---
3	20	26
4	21	27
5	22	28
6	23	29
7	24	30
8	25	31
9	26	32
---
3	20	0
4	21	0
5	22	0
6	23	0
7	24	0
8	25	0
9	26	0
+ FAIL
```

shlomi-noach:
https://github.com/github/gh-ost/pull/447/commits/effae08fbf54025111cf9627c322cc77575cf295 fixes the issue as follows:

- if a column's name only changes upper/lower case, `gh-ost` identifies it to be same column in _original_ and _ghost_ tables
- `gh-ost` will not pick a shared key where such a name change takes place. As an example, if your table has an `id int primary key`, you won't be able to `modify ID bigint` unless there's another unqiue key to be used.

We can potentially rediuce this shared key limitation, at the cost of more complexity. I think it's not worth it, for now, unless a strong case comes by.

shlomi-noach:
This is being tested right now.

shlomi-noach:

PR #451, opened 2017-07-19
ghost/changelog/del table names support max characters limit

shlomi-noach:
With this PR it's OK for the original table name to be very long, and up to maxlen (`64` characters).

The _old_, _ghost_ and _changelog_ names are truncated as needed to accommodate.

Furthermore, the truncation happens on the table name part, rather than on the suffix. To illustrate, this table name:

```
a12345678901234567890123456789012345678901234567890123456789012
```

is `63` characters long. The names `gh-ost` will pick are:
```
_a123456789012345678901234567890123456789012345678901234567_del
_a123456789012345678901234567890123456789012345678901234567_gho
_a123456789012345678901234567890123456789012345678901234567_ghc
```
Furthermore, if using `--timestamp-old-table` the suffix is even longer, and this truncates even more of the original name, e.g:
```
_a1234567890123456789012345678901234567890123_20130203195400_del
```

### DIFF
--- a/go/base/context_test.go
+++ b/go/base/context_test.go
@@ -1,0 +1,38 @@
+/*
+   Copyright 2016 GitHub Inc.
+	 See https://github.com/github/gh-ost/blob/master/LICENSE
+*/
+
+package base
+
+import (
+	"testing"
+
+	"github.com/outbrain/golib/log"
+	test "github.com/outbrain/golib/tests"
+)
+
+func init() {
+	log.SetLevel(log.ERROR)
+}
+
+func TestGetTableNames(t *testing.T) {
+	context = newMigrationContext()
+	{
+		context.OriginalTableName = "some_table"
+		test.S(t).ExpectEquals(context.GetOldTableName(), "_some_table_del")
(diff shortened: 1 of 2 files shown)


go/base/context_test.go
```diff
@@ -0,0 +1,47 @@
...
+		test.S(t).ExpectEquals(oldTableName, "_a1234567890123456789012345678901234567890123456789012345678_del")
+	}
+	{
+		context.OriginalTableName = "a123456789012345678901234567890123456789012345678901234567890123"
+		context.TimestampOldTable = true
+		longForm := "Jan 2, 2006 at 3:04pm (MST)"
+		context.StartTime, _ = time.Parse(longForm, "Feb 3, 2013 at 7:54pm (PST)")
+		oldTableName := context.GetOldTableName()
+		test.S(t).ExpectEquals(oldTableName, "_a1234567890123456789012345678901234567890123_20130203195400_del")
+	}
```
arthurnn: I am wondering in splitting this tests into different methods. When adding new tests here I got some mixed results because the context was mutated already.
WDYT?

shlomi-noach: @arthurnn sure!

Issue #450, opened 2017-07-19
Issue with DATETIME(6) on MariaDB


aadant:
Here are the steps to reproduce : 

1. on the master

```
create database ghost;
use ghost;
drop table if exists t;
create table t(id bigint, d datetime(6),primary key(id)) engine=InnoDB;
```

2. on the test replica 

Using gh-ost-binary-linux-20170403125842.tar.gz

```
touch /tmp/ghost.postpone.t.flag 
./gh-ost \
--max-load=Threads_running=32 \
--critical-load=Threads_running=1000 \
--initially-drop-ghost-table \
--initially-drop-old-table \
--chunk-size=10000 \
--dml-batch-size=100 \
--max-lag-millis=1500 \
--user="" \
--password="" \
--host= \
--database="ghost" \
--table="t" \
--verbose \
--skip-foreign-key-checks \
--alter="engine=InnoDB" \
--switch-to-rbr \
--cut-over=default \
--test-on-replica \
--default-retries=120 \
--panic-flag-file=/tmp/ghost.panic.flag \
--postpone-cut-over-flag-file=/tmp/ghost.postpone.t.flag \
--approve-renamed-columns \
--replica-server-id=83 \
--execute
```

3. on the master

> insert into t(id,d) values(1,'2017-07-14 00:09:51.073974')
> commit;

4. on the test replica

`rm /tmp/ghost.postpone.t.flag`

5. on the test replica database, microseconds are multiplied by 10

```
MariaDB [ghost]> select * from t;
+----+----------------------------+
| id | d                          |
+----+----------------------------+
|  1 | 2017-07-14 00:09:51.073974 |
+----+----------------------------+
1 row in set (0.02 sec)

MariaDB [ghost]> select * from _t_gho;
+----+----------------------------+
| id | d                          |
+----+----------------------------+
|  1 | 2017-07-14 00:09:51.739740 |
+----+----------------------------+
1 row in set (0.02 sec)
```
This looks like a major issue with DATETIME(6). The database is MariaDB 10.1.20.
Not sure if it is an issue with Oracle MySQL.

shlomi-noach:
Thank you, I'll investigate.

shlomi-noach:
I have a guess as for why this is happening. It's a bit outside my code but I should be able to fix it.

shlomi-noach:
Reproduced

shlomi-noach:
[...]

PR #452, opened 2017-07-19
timestamp/datetime submillis corruption fix

shlomi-noach:
fixes https://github.com/github/gh-ost/issues/450

This PR will fix the following issue: when applying events from binlog and where column type is `TIMESTAMP(6)` or `DATETIME(6)`, submillis can get corrupted. In particular this happens when submillis are `0*`. For example, the value `2016-10-31 11:22:33.012300` turns to `2016-10-31 11:22:33.123000`.

Initial commit adds a test to reproduce & validate the problem.


shlomi-noach:
Sent to testing

shlomi-noach:

PR #457, opened 2017-07-21
Add a force-table-names command flag

### LINKED ISSUE #399: gh-ost temp table
Hi  

Can we add one parameter to control the three ghost table name ,we want to control the table name

arthurnn:
usage example:
```
gh-ost  ...  --force-table-names=migration_tbl
```

The force-table-names flag will receive a table name and use that name
when creating the _del, _gho, _ghc tables.
For instance, for that example above, the tables would be:

```
_migration_tbl_del
_migration_tbl_gho
_migration_tbl_ghc
```

[fixes #399]


shlomi-noach:
3 years later, and this PR comes super handy for something I'm working on. Thought I'd drop a note.

Issue #454, opened 2017-07-19
FR Speed up the binary log application


aadant:
I noticed that gh-ost runtime is sensitive to its physical location. For example, if you use --test-on-replica, the gh-ost executable needs to be located on the slave for best performance. On a WAN, it can make a 1 to 100 difference. It can be the reason gh-ost will never catch up and the migration will last forever.

The reason is simple :

when looking at the general log, there are a lot of repeated statements like this on the same table :

`             
259010 Prepare  replace /* gh-ost `ghost`.`_t_gho` */ into
                                `_ghost`.`_t_gho`
                                values
                                        (?,, ?, ?)
259010 Execute  replace /* gh-ost `ghost`.`t_gho` */ into ...
259010 Close stmt
`

A single event created 3 network round trips.
This can obviously optimized.

If 2 consecutive events are hitting the same table and are of the same kind (insert, update), then they can be added at the end of the replace, up to a certain number in mysqldump fashion. The most efficient would be not to use prepared statement (that could be an option) and just generate a large statement ( 1 network round trip) or execute several statements per prepare if the next event is compatible ...

Using bulk statements or repeated execute would tremendously improve performance if the table is being written to when doing the schema change.

aadant:
@shlomi-noach : I think this should address https://www.percona.com/blog/2017/07/12/gh-ost-benchmark-against-pt-online-schema-change-performance/

shlomi-noach:
@aadant thank you for this suggestion!

A few thoughts:

- it is true we can merge multiple `INSERT` or multiple `REPLACE` statement.
- this would work in `dml-batch-size > 1` ; requires some app logic changes
- another approach would be to use [interpolateParams=false](https://github.com/go-sql-driver/mysql#interpolateparams) and this is a low hanging fruit, very easy to set up. This will cancel the extra round trips, and there'll only be a single trip per query!
  - I'll start testing this.

With regard https://www.percona.com/blog/2017/07/12/gh-ost-benchmark-against-pt-online-schema-change-performance/ it's worth noting that the test was executed all-on-the-same-box ; single master node with `gh-ost` running from same box. So network is unlikely to be a big issue.
I'd still be very curious to see the impact of non-prepared statements on this benchmark; but first a few days of testing are in order.

PR #459, opened 2017-07-25
interpolateParams=true: client side prepared statements

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/454

In an attempt to speed up applying of binlog events, this is a low hanging fruit: set `interpolateParams=true` ([doc](https://github.com/go-sql-driver/mysql#interpolateparams))

This will set client-side prepared statements. The benefit to that is reduced round trips to the MySQL server: only one per query instead of three per query.

### DIFF
--- a/go/mysql/connection.go
+++ b/go/mysql/connection.go
@@ -56,5 +56,5 @@ func (this *ConnectionConfig) GetDBUri(databaseName string) string {
 		// Wrap IPv6 literals in square brackets
 		hostname = fmt.Sprintf("[%s]", hostname)
 	}
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?charset=utf8mb4,utf8,latin1", this.User, this.Password, hostname, this.Key.Port, databaseName)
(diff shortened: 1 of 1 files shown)


go/mysql/connection.go
```diff
@@ -56,5 +56,5 @@ func (this *ConnectionConfig) GetDBUri(databaseName string) string {
 		// Wrap IPv6 literals in square brackets
 		hostname = fmt.Sprintf("[%s]", hostname)
 	}
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?charset=utf8mb4,utf8,latin1", this.User, this.Password, hostname, this.Key.Port, databaseName)
```
arthurnn: Do we want to add unit tests to the db uri method?

shlomi-noach: @arthurnn do you have something in mind? I'm happy for any and every test addition.

arthurnn: Let's merge that, I can work on these tests in a separate PR.

shlomi-noach: This will only be merged after a few exhaustive days of testing in production (which is happening right now 😄 ).

shlomi-noach:
This is being tested for `8` days now in production with no issues. Will be merged shortly.

shlomi-noach:
ok, tested for `15` days straight, no problems.

Issue #302, opened 2016-11-01
Optional feature: regret cut-over


shlomi-noach:
This feature was suggested by someone at the audience at PerconaLive Amsterdam 2016: when we cut-over, revert syncing and keep the _old_ table in sync with incoming traffic onto the new table, and support the ability to cut-back.

Consider the case of a `drop column` or a `drop index` that were incorrectly planned and cause outage. We would be able to revert such an operation by cutting-back onto the original table and completely canceling the migration.

There's a lot involved here, but it is a possible task.

zytek:
Yeah, it was me ;-) and I just came back here with the intention to post that feature request as we started evaluating gh-ost in my company. 

Another usage scenarios: `row_format=compressed` and ability to rollback in case of performance degradation.

shlomi-noach:
Updating that this is not on our immediate roadmap and though I've taken a quick look at it (and there's actually a beginning of a PR), I'm likely to not work on this in the coming month or maybe two.

morgo:
For `drop index` - this is one of the use cases for invisible indexes (MySQL 8.0).  Perhaps there is an opportunity to use similar nomenclature, or (when using 8.0) recommend this feature only for columns?

There exists a sister to invisible indexes called invisible columns.  But there are no current plans to implement such.  So I concur with the FR here that emulation is useful.

jessbreckenridge:
At Percona Live 2017, we brainstormed a possible solution for this, that would be perhaps less intensive.

The idea is as follows:

- When a gh-ost migration is run, a `gh-ost` migration would be run as normal, from the *old* table to table the *new* table.
- When the cutover happens, we start a new `gh-ost` migration... from *new* table to the *old* table.
- This cutover continues for X amount of time, for example 24 hours.
- If you regret what you have done, we can cutback to the original table as normal.

Pros:

- Uses capabilities that gh-ost is already capable of. This is simply another migration.
- Parity in load, disk space used, etc. The migration should be no more intensive than the migration that just stopped.
- Provides a method to recovery from potentially catastrophic migration mistakes that slip through CI, review, etc.

Cons:
- Tooling to properly synchronously run the cutover, and start the new migration into the old table would be needed.
- Please add and discuss.

tomkrouper:
@jonahberquist we discussed an alternate to this, but I didn't want to get your details wrong, can you post your thoughts here?

leeparayno:
[...]

PR #479, opened 2017-08-28
Implement teardown for gh-ost library

### LINKED ISSUE #465: Fixing leaked goroutines
Hey all!

Thanks for building such an awesome library!

We're getting ready to start using this as a part of our production workflow, and we wanted to build a wrapper around the `gh-ost` utility to manage some configuration on our end and to integrate it into our development/deployment workflows. We were hoping to build this wrapper in Go and consume the library directly.

When building out the proof of concept, we noticed that `migrator.go` and a few other places are starting infinite goroutines (for example the `initiateStatus`) that seem to be relying on the fact that it is invoked from a CLI and that the program will exit when `Migrate()` is complete.

The wrapper we are building will be a long running process (http server) that will manage the invocations of the gh-ost library. Would you be willing to accept a PR that cleans up these infinite goroutines, allowing the core logic to be used outside of the single CLI run?

Happy to hear any feedback on this approach if not. Thanks again!

nikhilmat:
Fixes #465 

This PR adds explicit cleanup to gh-ost for when the library is consumed by another application and might be executed multiple times (for example, by a webserver). This stops any infinite goroutines that have started once the migration has finished, closes any DB connections it has opened, and ensures that the migration context that is passed around is not being globally accessed - instead it instantiates a new one for each migration run and passes it around.

Please let me know how this looks to you, and happy to make any changes to improve it!

It also pulls in the change from https://github.com/outbrain/golib/pull/4.


shlomi-noach:
Thank you! just as heads up it'll take me a few days to review this, sorry for the delay.

nikhilmat:
@shlomi-noach do you think you will get a chance to review this this week?

shlomi-noach:
this is the week!

vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
```diff
@@ -144,6 +144,17 @@ func GetDB(mysql_uri string) (*sql.DB, bool, error) {
 	return knownDBs[mysql_uri], exists, nil
 }
 
+// Resets the knownDBs cache, used when the DB connections have been closed,
+//   and new connections are needed to access the DB
+func ResetDBCache() {
+	knownDBsMutex.Lock()
+	defer func() {
+		knownDBsMutex.Unlock()
+	}()
```
shlomi-noach: simpler:
```
defer knownDBsMutex.Unlock()
```

go/logic/migrator.go
```diff
@@ -1211,3 +1230,24 @@ func (this *Migrator) finalCleanup() error {
...
+		log.Infof("Tearing down applier")
+		this.applier.Teardown()
+	}
+
+	if this.eventsStreamer != nil {
+		log.Infof("Tearing down streamer")
+		this.eventsStreamer.Teardown()
+	}
+
+	sqlutils.ResetDBCache()
```
shlomi-noach: :+1:

go/logic/applier.go
```diff
@@ -30,16 +30,18 @@ const (
...
 type Applier struct {
-	connectionConfig *mysql.ConnectionConfig
-	db               *gosql.DB
-	singletonDB      *gosql.DB
-	migrationContext *base.MigrationContext
+	connectionConfig  *mysql.ConnectionConfig
+	db                *gosql.DB
+	singletonDB       *gosql.DB
+	migrationContext  *base.MigrationContext
+	finishedMigrating bool
```
shlomi-noach: This variable is accesses read/write by different goroutines. You'd need o change it to `int64` and use `atomic.StoreInt64` and `atomic.LoadInt64` ; see various such examples throughout the code.

go/logic/applier.go
```diff
@@ -310,6 +312,9 @@ func (this *Applier) InitiateHeartbeat() {
 
 	heartbeatTick := time.Tick(time.Duration(this.migrationContext.HeartbeatIntervalMilliseconds) * time.Millisecond)
 	for range heartbeatTick {
+		if this.finishedMigrating {
```
shlomi-noach: unsafe operation ; concurrency

go/logic/applier.go
```diff
@@ -1039,3 +1044,10 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
 	log.Debugf("ApplyDMLEventQueries() applied %d events in one transaction", len(dmlEvents))
 	return nil
 }
+
+func (this *Applier) Teardown() {
+	log.Debugf("Tearing down...")
+	this.db.Close()
+	this.singletonDB.Close()
+	this.finishedMigrating = true
```
shlomi-noach: unsafe operation ; concurrency ; use `atomic.StoreInt64`

go/logic/migrator.go
```diff
@@ -83,11 +84,13 @@ type Migrator struct {
 	applyEventsQueue chan *applyEventStruct
 
 	handledChangelogStates map[string]bool
+
+	finishedMigrating bool
```
shlomi-noach: see concurrency concerns above

go/logic/migrator.go
```diff
@@ -718,6 +727,9 @@ func (this *Migrator) initiateStatus() error {
 	this.printStatus(ForcePrintStatusAndHintRule)
 	statusTick := time.Tick(1 * time.Second)
 	for range statusTick {
+		if this.finishedMigrating {
```
shlomi-noach: unsafe operation ; concurrency

go/logic/migrator.go
```diff
@@ -942,6 +954,9 @@ func (this *Migrator) initiateStreaming() error {
 	go func() {
 		ticker := time.Tick(1 * time.Second)
 		for range ticker {
+			if this.finishedMigrating {
```
shlomi-noach: unsafe operation ; concurrency

go/logic/migrator.go
```diff
@@ -1132,6 +1147,10 @@ func (this *Migrator) executeWriteFuncs() error {
 		return nil
 	}
 	for {
+		if this.finishedMigrating {
```
shlomi-noach: unsafe operation ; concurrency

go/logic/migrator.go
```diff
@@ -1211,3 +1230,24 @@ func (this *Migrator) finalCleanup() error {
 
 	return nil
 }
+
+func (this *Migrator) teardown() {
+	this.finishedMigrating = true
```
shlomi-noach: unsafe operation ; concurrency

vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
nikhilmat: of course, thanks!

nikhilmat:
@shlomi-noach thanks for the notes on the concurrency concerns - those have been fixed!

i've been doing some additional testing over the past week and came across an issue with the DB cache in `sqlutils` when running concurrent migrations. since we are now closing our DB connections at the end of each migration, if another migration is running when one ends, the connection gets pulled out from underneath them because the cache has them share a connection pool. 

Do you have any objections to attaching the db connections pools that connect to `migrationContext.DatabaseName` and the `information_schema` to the migration context, and removing the database cache in `sqlutils`? that way, we avoid creating too many pools by always referencing it on the context, and it will be scoped to each migration so they don't interfere if one migration finishes and closes its pool.

nikhilmat:
@shlomi-noach, I implemented that proposed changed here https://github.com/github/gh-ost/pull/479/commits/84bdfdb1ada1669bb27458aaf9c0750b6f276de6 - let me know what you think of the approach! i believe with these changes, each migration will maintain its own connections before cleaning them up at the end - allowing for migrations to run concurrently.

Thanks again for your review, and sorry for the delay with the above changes! Will be sure to get to any feedback you have shortly :)

shlomi-noach:
@nikhilmat I'll be in a conference this week so will only take a look next week. But meanwhile, Re: thread pool, please see https://github.com/github/gh-ost/pull/491 and https://github.com/github/gh-ost/issues/482, work in progress by @zmoazeni to address that.

zmoazeni:
@shlomi-noach @nikhilmat hey folks. I'm not sure how the thread pool changes affect this PR, but by all means if you need any help just ask.

As far as I understand the codebase and my threadpool PR, I don't explicitly create a new connection. It just reuses the existing ones. So I _think_ they will both go hand-in-hand. But if they don't, just let me know.

shlomi-noach:
@nikhilmat 

> Do you have any objections to attaching the db connections pools that connect to migrationContext.DatabaseName and the information_schema to the migration context, and removing the database cache in sqlutils? 
>  I implemented that proposed changed here 84bdfdb - let me know what you think of the approach! i believe with these changes, each migration will maintain its own connections before cleaning them up at the end - allowing for migrations to run concurrently.

Thank you for working on this. 
I do not have objections to this, and can see the reasoning.

One thing I'd like not to change is `sqlutils.go`, which, as you notice, is under `vendor`. I'd like to keep it as it is (though I confess I've tweaked it here and there in other projects, and it's a technical debt I have to consolidate all changes).

Also, whatever happened to the `finishedMigrating` flag? The code looks so different now that I feel this is a completely new review.

nikhilmat:
@shlomi-noach great, I believe all the changes on moving the DB connections are covered in this PR - I was able to successfully run concurrent migrations locally.

Totally understand about not wanting to change the source code in `vendor/`. Are you okay with submitting this change to https://github.com/outbrain/golib and then updating the dependencies here? If not, I can stop using `sqlutils.GetDB` altogether, and just rely on that method exposed on the `migrator`. Let me know which you prefer, and I'll make the change. (Relatedly, I had a question about some of the vendor dependencies here: https://github.com/github/gh-ost/issues/501)

Sorry for the confusion with the layered changes. The `finishedMigrating` flag is still present on the `applier` and `migrator`. I updated them to use `atomic. LoadInt64` and `atomic. StoreInt64` in this commit https://github.com/github/gh-ost/pull/479/commits/5788ab534726f40a555db04861862c28d0ecf125. Besides addressing the concurrency concerns, there were no changes in the last two commits since your previous review (unless I missed something).

Thanks again! Let me know how you'd like to proceed with the `sqlutils` changes.

nikhilmat:
@shlomi-noach any thoughts on the above with regards to the `vendor/` changes for the DB cache? Let me know if there's anything else here that needs revising :)

nikhilmat:
@shlomi-noach friendly ping :) anything else i can do to help this along?

shlomi-noach:
@nikhilmat sorry for the delay! I was out for a while and then busy catching up. Will get to it.

nikhilmat:
@shlomi-noach no problem, thanks!

shlomi-noach:
Again, apologies for the delay. Still on it, just have some serious backlog.

nikhilmat:
@shlomi-noach thanks for checking in on this. i know there's still a decision to be made on how to apply the changes to the vendor directory:

> Totally understand about not wanting to change the source code in vendor/. Are you okay with submitting this change to https://github.com/outbrain/golib and then updating the dependencies here? If not, I can stop using sqlutils.GetDB altogether, and just rely on that method exposed on the migrator. Let me know which you prefer, and I'll make the change.

Feel free to chime in on your preference on this and I can make the change before taking another pass. Otherwise, I can go ahead remove the usage of `sqlutils.GetDB` and just rely on the migration context specific db cache.

Thanks!

shlomi-noach:
Apologies for keeping you waiting on such a small issue. 
For now, I'd like you to restore `sqlutils.go` to its original state.

What I also see you've done is https://github.com/github/gh-ost/pull/479/files#diff-b9f3c8a49a606c2654d3c9c33727e879R253, which just duplicates the code in `sqlutils.go`? But also undermines your desire to avoid caching the DB object.

Furthermore, `MigrationContext` isn't the right place for generating the DB instance ; consider it more as a data object that can be serialized.

I suggest the following:
- Restore `sqlutils.go` to its original state
- Remove the `GetDB()` code from `MigrationContext`
- Create `GetDB()` in `go/mysql/utils.go`
- Remove reference to `sqlutils.GetDB` and refactor as `mysql.GetDB`

Thank you!

nikhilmat:
@shlomi-noach the goal with moving the caching to the `MigrationContext` was that the caching would be done a per migration level, and so if migrations were running concurrently, closing a connection in one migration would not affect another. However, happy to follow your suggestion and revert the sqlutils changes and move that method to the mysql utils.

Thanks!

shlomi-noach:
> the goal with moving the caching to the MigrationContext was that the caching would be done a per migration level, and so if migrations were running concurrently, closing a connection in one migration would not affect another

I understand better now, thank you for explaining. To elaborate abit, a futuristic idea is to have the `MigrationContext` instance serialized and persisted. To that effect, it would not be possible to assume database driver/connections would exist when the instance loads. The specific use case right now is https://github.com/github/gh-ost/issues/302.

nikhilmat:
Ah, cool! Thanks for sharing, that's good to know. I had a separate question about adding a custom, migration-specific logger (we're sending our logs to cloudwatch), and this is really helpful. Will open up an issue to discuss.

This is ready for another pass when you get a chance. Incorporated your feedback and updated with the latest master!

nikhilmat:
👋 @shlomi-noach - do you think you'll get a chance to take another look at this soon?

shlomi-noach:
@nikhilmat apologies yet again for the delays and I appreciate your patience, thank you for pinging.
This looks good to me!

What I'll do is merge this onto a local `nm-refactor-migration-context` branch, and send to testing.
You will see this PR as "Merged", but it won't be merged to master. I'll @mention you on the merged branch.
This complexity is required because of the decoupling we make between this public repo and our internal infrastructure.

nikhilmat:
wonderful - thank you!!!

Issue #480, opened 2017-08-30
gh-ost hangs if global variable autcommit is OFF 


renecannao:
If global variable `autocommit` is OFF , gh-ost makes no progress after it creates the tables, and hangs forever.
Here is an example:
```
2017-08-30 10:17:56 INFO will use time_zone='SYSTEM' on applier
2017-08-30 10:17:56 INFO Examining table structure on applier
2017-08-30 10:17:56 INFO Applier initiated on mtdtbtest01:3306, version 5.6.36-82.1-log
2017-08-30 10:17:56 INFO Droppping table `schemaname`.`_table_tx_gho`
2017-08-30 10:17:56 INFO Table dropped
2017-08-30 10:17:56 INFO Droppping table `schemaname`.`_table_tx_ghc`
2017-08-30 10:17:56 INFO Table dropped
2017-08-30 10:17:56 INFO Creating changelog table `schemaname`.`_table_tx_ghc`
2017-08-30 10:17:56 INFO Changelog table created
2017-08-30 10:17:56 INFO Creating ghost table `schemaname`.`_table_tx_gho`
2017-08-30 10:17:56 INFO Ghost table created
2017-08-30 10:17:56 INFO Altering ghost table `schemaname`.`_table_tx_gho`
2017-08-30 10:17:56 INFO Ghost table altered
2017-08-30 10:17:56 INFO Waiting for ghost table to be migrated. Current lag is 0s
```

shlomi-noach:
Thank you!
Solution will be to add `&autocommit=true` to the connection string on applier connection (and preferably also on the inspector connection)

PR #481, opened 2017-08-30
gh-ost uses autocommit=1 for all connections

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/480

All connections use `autocommit=1` overriding whatever global value is set for `autocommit`.

cc @renecannao


shlomi-noach:
bug easily reproduced and confirmed to be fixed by this PR. Sent to testing in production.

Issue #482, opened 2017-08-30
Possible problems with gh-ost, pool-of-threads, and hot tables


zmoazeni:
I could use some help with a downtime we encountered using gh-ost today.

First we had no problem with the replica only tests and testing minor tables but we just tried it out on what we thought could be a meaty table. Sitting at 500k rows, it is much closer to one of our smaller tables than our largest. And it's not actively/heavily written to, but it very read heavily. And I think that's what got us into trouble.

gh-ost was able to clone the table and waited for me to drop the postpone flag. Once I did the writer-server's MySQL became unresponsive. Even when I logged into it and tried to connect directly my MySQL connection would hang. Now this is a very beefy machine, and the load average didn't budge. So it wasn't a matter of the hardware stalling, but definitely the way MySQL was processing.

We are using Percona MySQL 5.7 with the `thread_handling=pool-of-threads` turned on (a result of our upgrade from 5.6), and I think what may have happened according to the description outlined in https://github.com/github/gh-ost/issues/82 is that we must have starved the thread pool once gh-ost tried to lock this hot table for renaming.

I don't think the statements prior to gh-ost applying for the lock stalled us, but perhaps all of the queries immediately after gh-ost applied for the lock did. From the outside they began collecting up waiting for gh-ost to execute the atomic rename. That we can see happening by a pile up of our unicorn processes. gh-ost didnt seem to timeout applying for the lock ("INFO Tables locked") but it never renamed the table. It seemed to restart the sync process once more (restarted the "# Migrating ..." block) and that's when I killed it manually.

But I wonder if the prioritization of the actual rename got lost within MySQL and most queries were waiting for that to finish so gh-ost would give up the lock. That is my initial diagnosis anyway.

I need to do some digging to try and figure out if this is perhaps a configuration tuning issue. Perhaps we need to reconsider moving back from pool-of-threads to one-thread-per-connection. But I was hoping I could share that experience and see if you folks had some insight to help point me in the right direction. Maybe someone else has already run into this exact issue and could share their solution.

The threadpool size is certainly one configuration to re-evaluate but I think we may need to look into the thread pool high priority options in https://www.percona.com/doc/percona-server/LATEST/performance/threadpool.html

One direct question I wanted to ask: if we decided to try the two step rename with `--cut-over=two-step` does it require the same locking that is currently happening in the atomic rename? I know we could error queries between those two steps executing, but I wonder if that is another option worth exploring, or if I should be focusing all of my energy on MySQL's configuration.

[...]

PR #491, opened 2017-09-06
Allow gh-ost to modify the server using extra port

zmoazeni:
### Description

@shlomi-noach this PR is to extend gh-ost to allow it to connect to the master via the extra port per https://github.com/github/gh-ost/issues/482

- [x] contributed code is using same conventions as original code (I believe so, please let me know if I goofed up here)

We're going to test out this modified binary this week/next, so I can report back production behavior if you're interested.

~I did setup my stuff locally and verified ./test.sh works and ./localtests/test.sh mostly works. Everything except the last test `enum-pk` passes, but that test fails on master too, so this isn't the cause.~ _Edit: that must just be an inconsistent test. Re-running both ./test.sh works and ./localtests/test.sh passes 100% now._

---

Both Percona and Maria allow MySQL to be configured to listen on an extra port when their thread pool is enabled.

* https://www.percona.com/doc/percona-server/5.7/performance/threadpool.html
* https://mariadb.com/kb/en/the-mariadb-library/thread-pool-in-mariadb-51-53/

This is valuable because if the table has a lot of traffic (read or write load), gh-ost can end up starving the thread pool as incoming connections are immediately locked.

By using gh-ost on the extra port, MySQL locking will still behave the same, but MySQL will keep a dedicated thread for each gh-ost connection.

When doing this, it's important to inspect the extra-max-connections variable. Both Percona and Maria default to 1, so gh-ost may easily exceed with its threads.

An example local run using this

```
$ mysql -S /tmp/mysql_sandbox20393.sock -e "select @@global.port, @@global.extra_port"
+---------------+---------------------+
| @@global.port | @@global.extra_port |
+---------------+---------------------+
|         20393 |               30393 |
+---------------+---------------------+

./bin/gh-ost \
--initially-drop-ghost-table \
--initially-drop-old-table \
--assume-rbr \
--port="20395" \
--assume-master-host="127.0.0.1:30393" \
--max-load=Threads_running=25 \
--critical-load=Threads_running=1000 \
--chunk-size=1000 \
--max-lag-millis=1500 \
--user="gh-ost" \
--password="gh-ost" \
--database="test" \
--table="mytable" \
--verbose \
--alter="ADD mynewcol decimal(11,2) DEFAULT 0.0 NOT NULL" \
--exact-rowcount \
--concurrent-rowcount \
--default-retries=120 \
--panic-flag-file=/tmp/ghost.panic.flag \
--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag \
--execute
```

### DIFF
--- a/go/base/utils.go
+++ b/go/base/utils.go
@@ -11,6 +11,10 @@ import (
 	"regexp"
 	"strings"
 	"time"
+
+	gosql "database/sql"
+	"github.com/github/gh-ost/go/mysql"
+	"github.com/outbrain/golib/log"
 )
 
 var (
@@ -49,4 +53,26 @@ func StringContainsAll(s string, substrings ...string) bool {
 		}
 	}
 	return nonEmptyStringsFound
+}
+
+func ValidateConnection(db *gosql.DB, connectionConfig *mysql.ConnectionConfig) (string, error) {
+	query := `select @@global.port, @@global.version`
+	var port, extraPort int
+	var version string
+	if err := db.QueryRow(query).Scan(&port, &version); err != nil {
+		return "", err
+	}
+	extraPortQuery := `select @@global.extra_port`
+	if err := db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
+		// swallow this error. not all servers support extra_port
+	}
+
+	if connectionConfig.Key.Port == port || (extraPort > 0 && connectionConfig.Key.Port == extraPort) {
+		log.Infof("connection validated on %+v", connectionConfig.Key)
+		return version, nil
+	} else if extraPort == 0 {
+		return "", fmt.Errorf("Unexpected database port reported: %+v", port)
+	} else {
+		return "", fmt.Errorf("Unexpected database port reported: %+v / extra_port: %+v", port, extraPort)
+	}
 }
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -53,12 +53,14 @@ func (this *Applier) InitDBConnections() (err error) {
 		return err
 	}
 	this.singletonDB.SetMaxOpenConns(1)
-	if err := this.validateConnection(this.db); err != nil {
+	version, err := base.ValidateConnection(this.db, this.connectionConfig)
+	if err != nil {
 		return err
 	}
-	if err := this.validateConnection(this.singletonDB); err != nil {
+	if _, err := base.ValidateConnection(this.singletonDB, this.connectionConfig); err != nil {
 		return err
 	}
+	this.migrationContext.ApplierMySQLVersion = version
 	if err := this.validateAndReadTimeZone(); err != nil {
 		return err
 	}
@@ -71,20 +73,6 @@ func (this *Applier) InitDBConnections() (err error) {
 		return err
 	}
 	log.Infof("Applier initiated on %+v, version %+v", this.connectionConfig.ImpliedKey, this.migrationContext.ApplierMySQLVersion)
-	return nil
-}
-
-// validateConnection issues a simple can-connect to MySQL
-func (this *Applier) validateConnection(db *gosql.DB) error {
-	query := `select @@global.port, @@global.version`
-	var port int
-	if err := db.QueryRow(query).Scan(&port, &this.migrationContext.ApplierMySQLVersion); err != nil {
-		return err
-	}
-	if port != this.connectionConfig.Key.Port {
-		return fmt.Errorf("Unexpected database port reported: %+v", port)
-	}
-	log.Infof("connection validated on %+v", this.connectionConfig.Key)
 	return nil
 }
 
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -195,16 +195,10 @@ func (this *Inspector) validateConnection() error {
 	if len(this.connectionConfig.Password) > mysql.MaxReplicationPasswordLength {
 		return fmt.Errorf("MySQL replication length limited to 32 characters. See https://dev.mysql.com/doc/refman/5.7/en/assigning-passwords.html")
 	}
-	query := `select @@global.port, @@global.version`
-	var port int
-	if err := this.db.QueryRow(query).Scan(&port, &this.migrationContext.InspectorMySQLVersion); err != nil {
-		return err
-	}
-	if port != this.connectionConfig.Key.Port {
-		return fmt.Errorf("Unexpected database port reported: %+v", port)
-	}
-	log.Infof("connection validated on %+v", this.connectionConfig.Key)
-	return nil
+
+	version, err := base.ValidateConnection(this.db, this.connectionConfig)
+	this.migrationContext.InspectorMySQLVersion = version
+	return err
 }
 
 // validateGrants verifies the user by which we're executing has necessary grants
--- a/go/logic/streamer.go
+++ b/go/logic/streamer.go
@@ -107,7 +107,7 @@ func (this *EventsStreamer) InitDBConnections() (err error) {
 	if this.db, _, err = sqlutils.GetDB(EventsStreamerUri); err != nil {
 		return err
 	}
-	if err := this.validateConnection(); err != nil {
+	if _, err := base.ValidateConnection(this.db, this.connectionConfig); err != nil {
 		return err
 	}
 	if err := this.readCurrentBinlogCoordinates(); err != nil {
@@ -130,20 +130,6 @@ func (this *EventsStreamer) initBinlogReader(binlogCoordinates *mysql.BinlogCoor
 		return err
 	}
 	this.binlogReader = goMySQLReader
-	return nil
-}
-
-// validateConnection issues a simple can-connect to MySQL
-func (this *EventsStreamer) validateConnection() error {
-	query := `select @@global.port`
-	var port int
-	if err := this.db.QueryRow(query).Scan(&port); err != nil {
-		return err
-	}
-	if port != this.connectionConfig.Key.Port {
-		return fmt.Errorf("Unexpected database port reported: %+v", port)
-	}
-	log.Infof("connection validated on %+v", this.connectionConfig.Key)
 	return nil
 }
 


zmoazeni:
We did a run against a lesser read table this morning against the extra port and this branch. Worked great.

go/logic/inspect.go
```diff
@@ -195,12 +196,22 @@ func (this *Inspector) validateConnection() error {
...
 	}
-	query := `select @@global.port, @@global.version`
+
+	extraPortRegexp := regexp.MustCompile("Percona|Maria")
+	query := `select @@global.port, @@global.version, @@global.version_comment`
 	var port int
-	if err := this.db.QueryRow(query).Scan(&port, &this.migrationContext.InspectorMySQLVersion); err != nil {
+	var extraPort int
+	var versionComment string
+	if err := this.db.QueryRow(query).Scan(&port, &this.migrationContext.InspectorMySQLVersion, &versionComment); err != nil {
```
zmoazeni: This bit could be DRY'd up, but I wanted to get something in front of ya'll first. I can iterate to clean up some of this if you all want.

go/logic/inspect.go
```diff
@@ -195,12 +196,22 @@ func (this *Inspector) validateConnection() error {
...
 	}
-	if port != this.connectionConfig.Key.Port {
+	if extraPortRegexp.MatchString(versionComment) {
+		extraPortQuery := `select @@global.extra_port`
+		if err := this.db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
+			return err
+		}
+	}
+	if port != this.connectionConfig.Key.Port && extraPort != this.connectionConfig.Key.Port {
 		return fmt.Errorf("Unexpected database port reported: %+v", port)
```
zmoazeni: I do wonder if this error message should include the extra port (if it's available). Not sure.

go/logic/inspect.go
```diff
@@ -195,12 +196,22 @@ func (this *Inspector) validateConnection() error {
 	if len(this.connectionConfig.Password) > mysql.MaxReplicationPasswordLength {
 		return fmt.Errorf("MySQL replication length limited to 32 characters. See https://dev.mysql.com/doc/refman/5.7/en/assigning-passwords.html")
 	}
-	query := `select @@global.port, @@global.version`
+
+	extraPortRegexp := regexp.MustCompile("Percona|Maria")
```
zmoazeni: I just realized yesterday that Enterprise MySQL offers a thread pool, and presumably an extra port as well. I don't have access to a running version of that though (only Community). But this could be extended to include that as well if someone does have access or knows a valid way to differentiate.

go/logic/applier.go
```diff
@@ -76,12 +77,21 @@ func (this *Applier) InitDBConnections() (err error) {
...
+	if err := db.QueryRow(query).Scan(&port, &this.migrationContext.ApplierMySQLVersion, &versionComment); err != nil {
 		return err
 	}
-	if port != this.connectionConfig.Key.Port {
+	if extraPortRegexp.MatchString(versionComment) {
+		extraPortQuery := `select @@global.extra_port`
+		if err := db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
+			return err
+		}
+	}
```
shlomi-noach: Are all Percona/MariaDB servers guaranteed to have this variable? If so, since which version?

go/logic/inspect.go
shlomi-noach: This definition for extraPortRegexp repeats itself; can we make a constant in `context.go`, at least for the pattern?

go/logic/inspect.go
shlomi-noach: yes please :)

go/logic/inspect.go
shlomi-noach: Yes. Do let's `return log.Errorf("problem reading @@global.extra_port: %+v", err)`

zmoazeni:
> I'm first wondering whether -assume-master-host=my.master.host.com:30393 would solve the problem in the first place. Con: you'd need to specify master host name (not auto-detected) ; Pro: you specify the port explicitly. Or... I guess the calidateConnection() function then fails, because port != this.connectionConfig.Key.Port. I'm thinking aloud here.

Yeah I used `-assume-master-host=my.master.host.com:30393` and that bit works swell. I don't expect gh-ost to just _know_ I want it to connect on the extra port. But these `validateConnection()` functions fail hard because they're trying to protect the user. That's understandable.

One other option I thought we could is instead of querying to see if the extra port is enabled and if it matches the args passed, we could offer a new switch to just ignore validating the connection. That's a bit dangerous but it is lower tech (e.g. easier to maintain) and it would work. Let me know if you prefer that option instead.

> I'm suggesting a small safety net. I don't like that we test for Percona|Maria ; I'm thinking we should instead go brute force:
> 
> Attempt to SELECT @@global.extra_port AS gh_ost_attempt_read_extra_port. If error, then we silently keep extraPort == 0 and do not exit with error. If successful, our win.

I have zero objection to this. I'll make it so.

go/logic/applier.go
zmoazeni: I'll do some digging. I believe Percona ported Maria's code and Maria has had it a while, but I'll try and get some concrete dates/version numbers.

Regardless, I do like your thoughts on brute forcing it. Then we don't even have to check for Percona or Maria (or the early versions without the functionality).

I'll update once I get more info.

zmoazeni: Looks like this was introduced in MariaDB  5.5.21 on 16 Mar 2012:

* https://mariadb.com/kb/en/library/mariadb-5521-release-notes/ (this was the beta release for the 5.5 line, but the functionality stuck around for all 5.5+ versions)
* https://mariadb.com/kb/en/library/changes-improvements-in-mariadb-55/
* https://mariadb.com/kb/en/library/thread-pool-in-mariadb/

I found this bit interesting:

> Significantly more efficient thread pool, comparable in functionality to the closed source feature in MySQL Enterprise.

So I assume it MySQL enterprise has the same extra_port semantics. I don't have access to it, so I don't know. I also don't see any extra port information in their public documentation https://dev.mysql.com/doc/refman/5.5/en/thread-pool.html

---

For Percona it was introduced at the same time with version 5.5.29-30.0 on 26 Feb 2012 as beta functionality:

* https://www.percona.com/doc/percona-server/5.5/release-notes/Percona-Server-5.5.29-30.0.html#5.5.29-30.0
* Percona: https://www.percona.com/doc/percona-server/5.5/performance/threadpool.html

But has persisted in 5.6 and 5.7 as a stable feature.

Percona explicitly mentions that it was ported from Maria's code:

> Ported the Thread Pool patch from MariaDB. This feature enables the server to keep the top performance even with the increased number of client connections.

I found the timeline interesting (Percona's release/announcement before Maria). But I assume some folks on Percona's side were keeping up to date on the Maria alpha/pre-merged improvements and brought it over to Percona early.

---

Completely off topic, but has GitHub evaluated using the Thread Pool for their own workload? I'm just assuming you all don't since you haven't run into this problem. I also imagine you all deal with MANY more connections than we do (total connections that is. you may have better partitioning so per-server connections are around the same).

I'm genuinely curious if you all evaluated it and decided against it for one reason or another. Your experience might influence us to switch back to one-connection-per-thread. You also might be running MySQL Community so this isn't even an option.

If you prefer not to answer that publicly, but still want to share your reasoning, by all means you can also shoot me an email at zach.moazeni@gmail.com

I have been meaning to write up a blog post but we noticed issues in our upgrade from Percona 5.6 to Percona 5.7. I worked with Percona's team but it's really hard to narrow down what exactly changed behind the scenes in 5.7. My assumption is that something that was blocking concurrency internally in 5.6 was unblocked in 5.7 and the more threads began running concurrently causing us load issues. Particularly on deploys when we cycle thousands of application connections (close down old ones for dying unicorns and spin up new ones for new unicorns).

We could have spent more time on the connection cycling (more sleep/pause between a unicorn dying and a unicorn taking its place) but moving to the thread pool was a decent solution and so far it has worked well for us. Rails 5's schema caching has also helped prevent a thundering herd of queries as our Rails instances boot up, which helped reduce the query load from connections as they startup.

shlomi-noach: > Completely off topic, but has GitHub evaluated using the Thread Pool for their own workload? I'm just assuming you all don't since you haven't run into this problem. I also imagine you all deal with MANY more connections than we do (total connections that is. you may have better partitioning so per-server connections are around the same).

> I'm genuinely curious if you all evaluated it and decided against it for one reason or another. Your experience might influence us to switch back to one-connection-per-thread. You also might be running MySQL Community so this isn't even an option.

Nothing to hide. No, we haven't, and the reason is just backlog. There's a lot of MySQL features we just haven't had time to poke _yet_. Sometime...

shlomi-noach:
I'm also wondering whether `gh-ost` should issue a warning if it starts up and notices there's a thread pool; perhaps it should indicate that the user should try and use the `extra_port`, or actually just go ahead and use `extra_port` implicitly?

zmoazeni:
Hrm. I can't answer that one. I'm really not a thread pool expert. 

One thing that needs to be set is the extra_port_max_connections though. That defaults to 1 and gh-ost quickly exhausts that limit. So I am wary about automatically choosing that.

I'm not so concerned about the warning however it could be annoying if someone knows what they're doing and can't turn it off.

I was thinking of putting together a doc all about using gh-ost + thread pool, so perhaps that can help educate. 

I'm planning on spending time on these updates next week.

shlomi-noach:
> One thing that needs to be set is the extra_port_max_connections though. That defaults to 1 and gh-ost quickly exhausts that limit. So I am wary about automatically choosing that.

Good point. 

> I'm not so concerned about the warning however it could be annoying if someone knows what they're doing and can't turn it off.

I'd say `log.Warn("hey watch out....")` is a responsible thing to do.

> I was thinking of putting together a doc all about using gh-ost + thread pool, so perhaps that can help educate.
> I'm planning on spending time on these updates next week.
Merge state

Thank you so much!

go/base/utils.go
```diff
@@ -50,3 +54,25 @@ func StringContainsAll(s string, substrings ...string) bool {
...
+	var version string
+	if err := db.QueryRow(query).Scan(&port, &version); err != nil {
+		return "", err
+	}
+	extraPortQuery := `select @@global.extra_port`
+	if err := db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
+		// swallow this error. not all servers support extra_port
+	}
+
+	if connectionConfig.Key.Port == port || (extraPort > 0 && connectionConfig.Key.Port == extraPort) {
```
zmoazeni: I reworked this conditional into something that felt a bit better. This is easier for me to grok. But let me know if you see something else that I can change.

go/base/utils.go
```diff
@@ -50,3 +54,25 @@ func StringContainsAll(s string, substrings ...string) bool {
...
+		log.Infof("connection validated on %+v", connectionConfig.Key)
+		return version, nil
+	} else if extraPort == 0 {
+		return "", fmt.Errorf("Unexpected database port reported: %+v", port)
+	} else {
+		return "", fmt.Errorf("Unexpected database port reported: %+v / extra_port: %+v", port, extraPort)
```
zmoazeni: I wanted to call out this error message. Let me know if you want me to tweak the format.

go/logic/applier.go
```diff
@@ -53,12 +53,14 @@ func (this *Applier) InitDBConnections() (err error) {
 		return err
 	}
 	this.singletonDB.SetMaxOpenConns(1)
-	if err := this.validateConnection(this.db); err != nil {
+	version, err := base.ValidateConnection(this.db, this.connectionConfig)
+	if err != nil {
 		return err
 	}
-	if err := this.validateConnection(this.singletonDB); err != nil {
+	if _, err := base.ValidateConnection(this.singletonDB, this.connectionConfig); err != nil {
```
zmoazeni: This exposed kind of a problem with the original code. We would overwrite the ApplierMySQLVersion with this second call. I almost wrote it that way, but I wasn't entirely sure about `this.db` vs `this.singletonDB`. So let know how you want me to handle it.

I chose the first connection `this.db` to set the version.

go/logic/inspect.go
```diff
@@ -195,16 +195,10 @@ func (this *Inspector) validateConnection() error {
...
-	if err := this.db.QueryRow(query).Scan(&port, &this.migrationContext.InspectorMySQLVersion); err != nil {
-		return err
-	}
-	if port != this.connectionConfig.Key.Port {
-		return fmt.Errorf("Unexpected database port reported: %+v", port)
-	}
-	log.Infof("connection validated on %+v", this.connectionConfig.Key)
-	return nil
+
+	version, err := base.ValidateConnection(this.db, this.connectionConfig)
```
zmoazeni: I decided to keep around this private `validateConnection` function since you added the password length check.

go/base/utils.go
shlomi-noach: It's fine.

go/logic/applier.go
shlomi-noach: I'll look into it.

go/logic/inspect.go
shlomi-noach: I'll consolidate and add a password-check flag

zmoazeni:
Thank YOU! And the rest of the gh-ost contributors!

Issue #496, opened 2017-09-18
Version 1.0.40 writes the password in the output


plapythian:
We upgraded to 1.0.40 and we're seeing the database password is written in the log:

XXXX/XX/XX XX:XX:XX binlogsyncer.go:75: [info] create BinlogSyncer with config &{99999 mysql db.somedomain.com 3306 user **password**  false false <nil>}

shlomi-noach:
This is done as part of the `go-mysql` underlying library, [here](https://github.com/siddontang/go-mysql/blob/89fc82130ac2531f20c672e80d5283c0b0c15290/replication/binlogsyncer.go#L106). 

I'll edit to remove the password as I'm including this library inline.

@plapythian would you care to likewise submit this issue in https://github.com/siddontang/go-mysql/, so that @siddontang can consider doing the same upstream?

siddontang:
@shlomi-noach 

I will ignore the password output.

siddontang:
PTAL @shlomi-noach https://github.com/siddontang/go-mysql/pull/163

plapythian:
Thanks

shlomi-noach:
I'll incorporate the change either tomorrow or after PerconaLive.

morgado-pythian:
Wondering if this has been fixed in any release?
I'm still seeing it on latest:

```
$ sh test-g.sh
2018-01-11 11:31:20 INFO starting gh-ost 1.0.42
2018-01-11 11:31:20 INFO Migrating `foo`.`tbl1`
2018-01-11 11:31:20 INFO connection validated on localhost:20795
2018-01-11 11:31:20 INFO User has ALL privileges
2018-01-11 11:31:20 INFO binary logs validated on localhost:20795
2018-01-11 11:31:20 INFO Restarting replication on localhost:20795 to make sure binlog settings apply to replication thread
2018-01-11 11:31:20 INFO Inspector initiated on LMCI:20795, version 5.7.20-log
2018-01-11 11:31:20 INFO Table found. Engine=InnoDB
2018-01-11 11:31:20 INFO Estimated number of rows via EXPLAIN: 1
2018-01-11 11:31:20 INFO Recursively searching for replication master
2018-01-11 11:31:20 INFO Master found to be LMCI:20795
2018-01-11 11:31:20 INFO log_slave_updates validated on localhost:20795
2018-01-11 11:31:20 INFO connection validated on localhost:20795
[...]

PR #499, opened 2017-10-02
updated go-mysql library

shlomi-noach:
Periodic update of the `go-mysql` library. Among other things, tackling https://github.com/github/gh-ost/issues/496


shlomi-noach:
This one is failing local tests in the meantime.

shlomi-noach:
There's an inconsistency of timzone handling between `time2` and `datetime2`, and `gh-ost` is unable to treat the two in the same way. For now, I'm not updating the `go-mysql` library.

druud62:
In MySQL, the timestamp-type is physically stored as a UNIX-epoch-based numeric value,
but the datetime-type is much like a string. (And there is still no datetimezone-type.)

So datetime-values should always be copied as-is, as they have no inherent timezone.

This of course makes alters that switch between these types problematic, 
for example because a datetime-value can be non-existent in the current timezone.
https://bugs.mysql.com/bug.php?id=85641
And much more so if not all involved database servers run with the same timezone config.

With timestamp-values, one should simply always copy the raw UNIX-epoch-based numerical value, but there is no good functionality for that on the SQL-level.

shlomi-noach:
> In MySQL, the timestamp-type is physically stored as a UNIX-epoch-based numeric value,
but the datetime-type is much like a string. (And there is still no datetimezone-type.)

> So datetime-values should always be copied as-is, as they have no inherent timezone.

I am indeed very much aware of both. Interestingly, B.com delivered the first non trivial TZ use case where the two did not agree with each other.

Also of interest is that binary logs always store timestamps as UTC, so they convert them, while `datetime`s are of course never converted.
This is also much about the library I use to read events from the binary logs.

druud:
There is an exception on the "with timestamp-values, one should simply always copy the raw UNIX-epoch-based numerical value", because 0 is a special value, that should not be adjusted per timezone.

shlomi-noach:
`go-mysql` has been updated in https://github.com/github/gh-ost/pull/684. This PR is redundant.

PR #504, opened 2017-10-21
Simplify validation of cut-over-lock-timeout-seconds flag

### LINKED ISSUE #474: Why limit CutOverLockTimeoutSeconds to 10 seconds?
There's a command line flag (`-cut-over-lock-timeout-seconds`) which allows you to control how much time to wait for a lock when performing a cutover.

Why is this [limited to 10 seconds](https://github.com/github/gh-ost/blob/5294ab6bd10c36bba7cd37e5739e6cab70b2cfda/go/base/context.go#L341)? Seems like [MySQL defaults to 50 sec](https://dev.mysql.com/doc/refman/5.7/en/innodb-parameters.html#sysvar_innodb_lock_wait_timeout) and allows much larger values.

It would be nice for potentially high-load tables to increase this lock wait timeout. We've had to limit database load temporarily in order to perform cut-overs due to the short lock wait timeout.

Xopherus:
- If user provides an invalid value for the flag, gh-ost
  will log + exit instead of keeping a default

- If user doesn't provide value for flag, default
  is still set to 3sec

Closes #474


go/base/context.go
```diff
@@ -340,10 +339,10 @@ func (this *MigrationContext) HasMigrationRange() bool {
 
 func (this *MigrationContext) SetCutOverLockTimeoutSeconds(timeoutSeconds int64) error {
 	if timeoutSeconds < 1 {
-		return fmt.Errorf("Minimal timeout is 1sec. Timeout remains at %d", this.CutOverLockTimeoutSeconds)
+		return fmt.Errorf("Minimal timeout is 1sec.")
```
shlomi-noach: Let's clarify in this message _which timeout_, namely, specify `-cut-over-lock-timeout-seconds`

go/base/context.go
```diff
@@ -340,10 +339,10 @@ func (this *MigrationContext) HasMigrationRange() bool {
...
 	}
 	if timeoutSeconds > 10 {
-		return fmt.Errorf("Maximal timeout is 10sec. Timeout remains at %d", this.CutOverLockTimeoutSeconds)
+		return fmt.Errorf("Maximal timeout is 10sec.")
```
shlomi-noach: same as above

go/base/context.go
```diff
@@ -225,7 +225,6 @@ func newMigrationContext() *MigrationContext {
 		InspectorConnectionConfig:           mysql.NewConnectionConfig(),
 		ApplierConnectionConfig:             mysql.NewConnectionConfig(),
 		MaxLagMillisecondsThrottleThreshold: 1500,
-		CutOverLockTimeoutSeconds:           3,
```
shlomi-noach: Why is this removed? The default should remain at `3`

Xopherus: Ah, I figured it would be best to have one place where the default value is defined - that would be in [main.go](https://github.com/github/gh-ost/blob/master/go/cmd/gh-ost/main.go#L89) with the rest of the flags.

Issue #519, opened 2017-11-08
Update primary key value, lost data


rudy-gao:
Hello, I use gh-ost version 1.0.42 to do ddl, when I update primary key value, I may lost date.
For example my table t2 has 1000000 rows, table  struct as following:
```
CREATE TABLE `t2` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `name` varchar(20) DEFAULT NULL,
  `password` varchar(20) DEFAULT NULL,
  `drc_check_time` timestamp(3) NOT NULL DEFAULT CURRENT_TIMESTAMP(3) ON UPDATE CURRENT_TIMESTAMP(3),
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=1000000 DEFAULT CHARSET=utf8mb4
```
1. First gh-ost use prmary key order to copy data to new table
2. When gh-ost copy to 100000, I execute a update sql
update t2 set name='rudytest',id=-2 where id=888887;
3. Gh-ost read binlog event and do event on new table
```
2017-11-08 14:04:46 INFO dml event query args; query=
 			update /* gh-ost `test`.`_t2_gho` */
 					`test`.`_t2_gho`
				set
					`id`=?, `name`=?, `password`=?, `drc_check_time`=?
				where
 					((`id` = ?))
 		; args=[-2 rudytest <nil> 2017-11-08 14:04:24.948 888887],dmlEvent=[Update on test:t2]
```
4. Gh-ost update 0 row on new table, because id=888887 is not copy to new table, and old table lost id=888887 data, row copy can not copy this row to new table, so I lost this row data

shlomi-noach:
Confirmed.

While updating a `PRIMARY KEY` is not generally considered good practice, it should be supported.

`oak-online-alter-table` originally addressed this very issue by converting any `update` into two queries:
- `delete`
- `replace into`

I'm not sure how it came to be I skipped this scenario in `gh-ost`. But I would be wary of reproducing `delete`+`replace`.

shlomi-noach:
This bug is true not only for `PRIMARY KEY` but more specifically to the chosen migration's `UNIQUE KEY`. The migration _typically_ uses `PK` for iteration, but not necessarily so. It can utilize another `UNIQUE KEY` in the event `PK` does not exist, or `PK` is being modified.

shlomi-noach:
https://github.com/github/gh-ost/pull/520 follows up on this issue.

[...]

PR #520, opened 2017-11-08
UPDATE on migration's unique key column data loss: fixing

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/519

This work in progress PR is to tackle https://github.com/github/gh-ost/issues/519, where an update to the migration's iterating unique key column can result with data loss.

We will either:
- Solve it completely, or
- Recognize the problem in runtime and throw an error, failing the migration, or
- Update the documentation to note such an `update` is unsupported.

We'll see how goes.

Initial commit: a test that reproduces the data loss.


shlomi-noach:
At this stage `gh-ost` will fail a migration when an `UPDATE` is made to values of the migration key.
This is not to be merged yet. I want to see if `gh-ost` can further resolve the issue gracefully.

shlomi-noach:
https://github.com/github/gh-ost/pull/520/commits/3898d49f6c362a06225aa35fad70fe1e16825e8a turns an `UPDATE` which modifies the migration key into `DELETE`+`INSERT` (to clarify, only when columns of migration key are actually changed).

Sending to testing.

shlomi-noach:
Tests are good in production.

PR #536, opened 2018-01-11
go-mysql library to avoid printing password in cleartext

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/496

This PR mimics https://github.com/siddontang/go-mysql/pull/163/ without importing the updated `go-mysql` library.


shlomi-noach:
Deploys and works well.

Issue #538, opened 2018-01-22
Incorrect datetime value: '1970-01-01 00:00:00' for column 'create_time'


liuguoshun:
**https://dev.mysql.com/doc/refman/5.7/en/datetime.html
The TIMESTAMP data type is used for values that contain both date and time parts. TIMESTAMP has a range of '1970-01-01 00:00:01' UTC to '2038-01-19 03:14:07' UTC. 

Caused by :1970-01-01 00:00:00 is not in the range of timestamp. 
**

How to repeat:

1.  
mysql version:5.5.44-log

2.
```
create database mysqlslap charset utf8;

CREATE TABLE test11 (
  id bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT '',
  create_time timestamp NULL DEFAULT '0000-00-00 00:00:00',
  update_time timestamp NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  counter int(10) unsigned DEFAULT NULL,
  PRIMARY KEY (id)
) ENGINE=InnoDB AUTO_INCREMENT=3 DEFAULT CHARSET=utf8;

insert into test11 values (1,'0000-00-00 00:00:00',now(),0);
```
  3.create sql for test.
more gh-ost_test.sql
`update test11 set counter=counter+1 where id=1;`

4. use mysqlslap for update
`mysqlslap -uroot -proot123 -h127.0.0.1 --port=3309  test -c 10 --query=gh-ost_test.sql --number-of-queries=1000000;`

5.while mysqlslap  is runing, in new window execute gh-ost .
`/apps/sh/gh-ost  -user="root" -password=root123  -host=127.0.0.1 -port=3309 -allow-on-master  --initially-drop-old-table -database="mysqlslap" -table="test11"  -alter="add column name varchar(1)" -execute`

6.logs:
```
2018/01/22 18:06:07 binlogsyncer.go:75: [info] create BinlogSyncer with config &{99999 mysql 127.0.0.1 3309 root root123  false false <nil>}
2018/01/22 18:06:07 binlogsyncer.go:241: [info] begin to sync binlog from position (mysql-bin.000003, 16900979)
2018/01/22 18:06:07 binlogsyncer.go:134: [info] register slave for master server 127.0.0.1:3309
2018/01/22 18:06:07 binlogsyncer.go:568: [info] rotate to (mysql-bin.000003, 16900979)
[...]

PR #539, opened 2018-01-28
Incorrect datetime 1970

shlomi-noach:
Investigating https://github.com/github/gh-ost/issues/538, this PR adds tests to reproduce #538 's error.


shlomi-noach:
5 days into testing, so far so good.

Issue #470, opened 2017-08-15
enhancement:does it support aliyun rds?


dikang123:
aliyun rds is a version of amazon rds in China:[aliyun rds](https://www.alibabacloud.com/product/apsaradb-for-rds?spm=a3c0i.7939108.675768.dnavproductsd1.4f0fe7d4puBL7f).

I have a test,but return some sort of error:
`2017-08-15 16:40:38 FATAL sql: Scan error on column index 0: converting driver.Value type <nil> ("<nil>") to a int: invalid syntax`

shlomi-noach:
@dikang123 any more input is appreciated. What is the table you're trying to convert? Which `mysql` version? Which `gh-ost` version? What's the `alter` statement?

dikang123:
origin table structure:
CREATE TABLE `vote_record` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT COMMENT '主键id',
  `user_id` varchar(22) DEFAULT NULL,
  `vote_num` int(10) unsigned NOT NULL DEFAULT '0' COMMENT '投票数',
  `group_id` int(10) unsigned NOT NULL DEFAULT '0' COMMENT '用户组id 0-未激活用户 1-普通用户 2-vip用户 3-管理员用户',
  `status` tinyint(2) unsigned NOT NULL DEFAULT '1' COMMENT '状态 1-正常 2-已删除',
  `create_time` datetime NOT NULL DEFAULT '0000-00-00 00:00:00' COMMENT '创建时间',
  `modify_time` datetime NOT NULL DEFAULT CURRENT_TIMESTAMP ON UPDATE CURRENT_TIMESTAMP,
  `col2` bigint(20) DEFAULT NULL COMMENT 'test',
  PRIMARY KEY (`id`),
  UNIQUE KEY `udx_user_id` (`user_id`),
  KEY `idx_usr_id` (`user_id`)
) ENGINE=InnoDB AUTO_INCREMENT=200000 DEFAULT CHARSET=utf8 COMMENT='投票记录表'；

mysql version:5.6.16-log
gh-ost:1.0.40


alter statement:./gh-ost --host=xxx --user=xxx --password=xxx --database=appversionsdb --table=vote_record --alter="add column tag int not null default 0" --chunk-size=2000 --max-load=Threads_running=20 --allow-on-master --execute

shlomi-noach:
@dikang123 can I get the log preceeding that fatal message please?

[...]

PR #541, opened 2018-01-30
Support Aliyun RDS

zhangxiaojian:
Related issue: https://github.com/github/gh-ost/issues/470
### Description

This PR makes gh-ost to support Aliyun RDS.  

Aliyun RDS is the largest cloud database in China, and the third of the world. For the reason of security and architecture , we need to hide some mysql variables so gh-ost will get ilegal values. This PR add --aliyun-rds flag to avoid getting them.

### Usage
For now, users can't connect to the slave node of rds, so gh-ost may working directly on master with flags:
* --allow-on-master
* --assume-rbr
* --assume-master-host
* --aliyun-rds

Thanks for gh-ost team developed an excellent tools for online DDL, and hope more people can use it on cloud database.  

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -91,6 +91,7 @@ type MigrationContext struct {
 	SkipRenamedColumns       bool
 	IsTungsten               bool
 	DiscardForeignKeys       bool
+	AliyunRDS                bool
 
 	config            ContextConfig
 	configMutex       *sync.Mutex
@@ -213,6 +214,8 @@ type ContextConfig struct {
 		Max_Load              string
 	}
 }
+
+var Context *MigrationContext
 
 func NewMigrationContext() *MigrationContext {
 	return &MigrationContext{
--- a/go/base/utils.go
+++ b/go/base/utils.go
@@ -65,15 +65,24 @@ func StringContainsAll(s string, substrings ...string) bool {
 }
 
 func ValidateConnection(db *gosql.DB, connectionConfig *mysql.ConnectionConfig) (string, error) {
-	query := `select @@global.port, @@global.version`
+	versionQuery := `select @@global.version`
 	var port, extraPort int
 	var version string
-	if err := db.QueryRow(query).Scan(&port, &version); err != nil {
+	if err := db.QueryRow(versionQuery).Scan(&version); err != nil {
 		return "", err
 	}
 	extraPortQuery := `select @@global.extra_port`
 	if err := db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
 		// swallow this error. not all servers support extra_port
+	}
+	// AliyunRDS set users port to "NULL", replace it by gh-ost param
+	if Context.AliyunRDS {
+		port = connectionConfig.Key.Port
+	} else {
+		portQuery := `select @@global.port`
+		if err := db.QueryRow(portQuery).Scan(&port); err != nil {
+			return "", err
+		}
 	}
 
 	if connectionConfig.Key.Port == port || (extraPort > 0 && connectionConfig.Key.Port == extraPort) {
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -44,6 +44,7 @@ func acceptSignals(migrationContext *base.MigrationContext) {
 // main is the application's entry point. It will either spawn a CLI or HTTP interfaces.
 func main() {
 	migrationContext := base.NewMigrationContext()
+	base.Context = migrationContext
 
 	flag.StringVar(&migrationContext.InspectorConnectionConfig.Key.Hostname, "host", "127.0.0.1", "MySQL hostname (preferably a replica, not the master)")
 	flag.StringVar(&migrationContext.AssumeMasterHostname, "assume-master-host", "", "(optional) explicitly tell gh-ost the identity of the master. Format: some.host.com[:port] This is useful in master-master setups where you wish to pick an explicit master, or in a tungsten-replicator where gh-ost is unable to determine the master")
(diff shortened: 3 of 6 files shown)


dikang123:
Great,hope this PR could be merged into the trunk ASAP!:)

shlomi-noach:
I'm very sorry for the delay. I'll be looking into this in more depth shortly.

shlomi-noach:
Sorry for the delay, I'll be reviewing this shortly.

zhangxiaojian:
@shlomi-noach  Any progress ?

go/mysql/utils.go
```diff
@@ -171,9 +171,11 @@ func GetSelfBinlogCoordinates(db *gosql.DB) (selfBinlogCoordinates *BinlogCoordi
 }
 
 // GetInstanceKey reads hostname and port on given DB
-func GetInstanceKey(db *gosql.DB) (instanceKey *InstanceKey, err error) {
+func GetInstanceKey(db *gosql.DB, isAliyunRDS bool) (instanceKey *InstanceKey, err error) {
 	instanceKey = &InstanceKey{}
-	err = db.QueryRow(`select @@global.hostname, @@global.port`).Scan(&instanceKey.Hostname, &instanceKey.Port)
+	if isAliyunRDS != true {
```
shlomi-noach: Please change to `if !isAliyunRDS {`

go/logic/inspect.go
```diff
@@ -53,10 +53,12 @@ func (this *Inspector) InitDBConnections() (err error) {
 	if err := this.validateConnection(); err != nil {
 		return err
 	}
-	if impliedKey, err := mysql.GetInstanceKey(this.db); err != nil {
+	if impliedKey, err := mysql.GetInstanceKey(this.db, this.migrationContext.AliyunRDS); err != nil {
 		return err
 	} else {
-		this.connectionConfig.ImpliedKey = impliedKey
+		if this.migrationContext.AliyunRDS != true {
```
shlomi-noach: Please change to `if !isAliyunRDS {`

go/base/context.go
```diff
@@ -214,6 +215,8 @@ type ContextConfig struct {
 	}
 }
 
+var Context *MigrationContext
```
shlomi-noach: I'd like to get rid of this global variable.

go/base/utils.go
```diff
@@ -65,16 +65,25 @@ func StringContainsAll(s string, substrings ...string) bool {
 }
 
 func ValidateConnection(db *gosql.DB, connectionConfig *mysql.ConnectionConfig) (string, error) {
-	query := `select @@global.port, @@global.version`
+	versionQuery := `select @@global.version`
 	var port, extraPort int
 	var version string
-	if err := db.QueryRow(query).Scan(&port, &version); err != nil {
+	if err := db.QueryRow(versionQuery).Scan(&version); err != nil {
```
shlomi-noach: I don't see where `version` is used?

go/cmd/gh-ost/main.go
```diff
@@ -44,6 +44,8 @@ func acceptSignals(migrationContext *base.MigrationContext) {
 // main is the application's entry point. It will either spawn a CLI or HTTP interfaces.
 func main() {
 	migrationContext := base.NewMigrationContext()
+	base.Context = migrationContext
```
shlomi-noach: please remove the `base.Context` global variable.

shlomi-noach: I actually don't see where it is used.

go/logic/applier.go
```diff
@@ -89,10 +89,12 @@ func (this *Applier) InitDBConnections() (err error) {
 	if err := this.validateAndReadTimeZone(); err != nil {
 		return err
 	}
-	if impliedKey, err := mysql.GetInstanceKey(this.db); err != nil {
+	if impliedKey, err := mysql.GetInstanceKey(this.db, this.migrationContext.AliyunRDS); err != nil {
 		return err
 	} else {
-		this.connectionConfig.ImpliedKey = impliedKey
+		if this.migrationContext.AliyunRDS != true {
```
shlomi-noach: Please change to `if !isAliyunRDS {`

go/logic/inspect.go
shlomi-noach: Or also see my suggestion to replace the entire block.

go/mysql/utils.go
shlomi-noach: Or also see my suggestion to replace the entire block.

go/logic/applier.go
shlomi-noach: Or also see my suggestion to replace the entire block.

go/base/utils.go
zhangxiaojian: The version is used in the function Inspector.validateConnection like before, this patch just prevent to get port from database.

go/logic/applier.go
zhangxiaojian: Thanks for your suggestion ! replace the entire block makes the code cleaner.

go/base/context.go
zhangxiaojian: The global variable is used in base/utils.go ValidateConnection. I need to get the aliyunRDS flag to avoid get the port from database. If don't use the global variable, add a param ( migrationContext may be ) to the function? or any other suggestions ?

shlomi-noach: Indeed it is used there. Sorry, I've missed this.

I still wish to get rid of the global variable. It will not play well with testing and embedding of `gh-ost` as a library. Please add a `migrationContext *MigrationContext` argument to base/utils.go `ValidateConnection()` function and use it.

go/base/utils.go
shlomi-noach: Thank you, you are right.

go/base/context.go
zhangxiaojian: Branch is updated, thanks for the review.

go/base/utils.go
```diff
@@ -64,17 +64,26 @@ func StringContainsAll(s string, substrings ...string) bool {
...
 		return "", err
 	}
 	extraPortQuery := `select @@global.extra_port`
 	if err := db.QueryRow(extraPortQuery).Scan(&extraPort); err != nil {
 		// swallow this error. not all servers support extra_port
 	}
+	// AliyunRDS set users port to "NULL", replace it by gh-ost param
```
everpcpc: Maybe this can be changed to a more general option like `assume port` ?
Since mysql running in docker with NAT would also report different port.

zhangxiaojian: That's a good point, --aliyun-rds is prevent getting port and hostname now, maybe --assume-check or something, @shlomi-noach we may need a product manager to deicide this.  : )

shlomi-noach: `--assume-port` sounds good.
Please feel free to choose whether you want to change this in this PR or in a different PR, and let me know.

zhangxiaojian: I would like to keep --aliyun-rds flags because we can do more things in the future. For example, the slave node is invisible for users and gh-ost cann't request binlog from one of them, maybe we can open a path for gh-ost.  --assume-port also make sense, but in a different PR I think.

shlomi-noach: Makes sense to me to keep `--aliyun-rds`.

wps353:
@zhangxiaojian 
If database name including special character,such as '_'.
this tool still not working.

Issue #542, opened 2018-02-05
'lock in share mode' is for what? usefull?


coco-b1107:
> This is the place to report a bug, ask a question, or suggest an enhancement.

> This is also the place to make a discussion before creating a PR.

> If this is a bug report, please provide a test case (e.g., your table definition and gh-ost command) and the error output.

> Please use markdown to format code or SQL: https://guides.github.com/features/mastering-markdown/

> Please label the issue on the right (bug, enhancement, question, etc.).

> And please understand if this issue is not addressed immediately or in a timeframe you were expecting.

> Thank you!

shlomi-noach:
Please be more specific, what is your concern, where is the relevant code?

coco-b1107:
an enhancement
1. rowcopy query>> insert ignore into ghosttable (sharedcolumn) (select sharedcolumn from originaltable ... lock in share mode)
2. lock in share mode>> prevent from updating the row data when select a row data, get row data uniformity
3. Online DML operation (take off  'lock in share mode')
         * insert; no effect
         * delete; when select,  however, the row data has been deleted, no effect
         * update; when select, get the row data updating, but, binlog prefer(data uniformity), no effect
4. conclusion
     'lock in share mode' has no effect, would be took off

shlomi-noach:
Thank you. In `pt-online-schema-change` there's a reference to https://bugs.mysql.com/bug.php?id=45694 for the reason to use `lock in share mode`. this bug is long since closed.

I _think_ `lock in share mode` can be removed, especially since we're not using triggers and the copy is not synchronous. cc @ggunson for additional thoughts.

ggunson:
tl;dr I say test it out, but I'm not sure. 

Here are my additional thoughts.

1. My understanding has been that `insert...select` does a shared read lock on the selected rows already, making `LOCK IN SHARE MODE` redundant. The "High Performance MySQL" book first agrees with me on this (assuming you keep `innodb_locks_unsafe_for_binlog=OFF`). 

2. Do note, `insert...select` behaviour is affected by binlog_format:

> INSERT ... SELECT requires a greater number of row-level locks [in statement-based replication] than with row-based replication.
https://dev.mysql.com/doc/refman/5.7/en/replication-sbr-rbr.html

3. And later in the "High Perf MySQL" book they state that

> In particular, `INSERT ... SELECT` locks all the rows it reads from the source table by default when using statement-based replication
High Perf MySQL 3rd edition p 503

[...]

PR #547, opened 2018-02-11
Experiment: removing LOCK IN SHARE MODE

shlomi-noach:
Storyline: https://github.com/github/gh-ost/issues/542

This PR experiments with removing `LOCK IN SHARE MODE` while running rowcopy.

To be deployed onto testing servers in prod.


dikang123:
should this pr merge to master branch:) ?

shlomi-noach:
I didn't get enough insight to conclude that this won't damage migrations. testing-on-replica will not test this particular change well. I'm inclined to keep `lock in share mode`.

PR #559, opened 2018-03-12
GBK charset tests

shlomi-noach:
Context: https://github.com/github/gh-ost/issues/532

Some time after #533 was submitted by @ceshihao we [found](https://github.com/github/gh-ost/pull/533#issuecomment-371382744) that `gkb` charset seems to work out of the box. It may relate to an update to the `go-sql-driver`.

This PR adds tests to confirm `gbk` works.

PR #533, opened 2018-01-08
Support gbk Encoding To Fix #532 

ceshihao:
To fix #532 

### Description

This PR adds `gbk` encoding support.

Issue #532, opened 2018-01-03
error 1366 on invalid character


cenkore:
I believe i have run into the same issue #290 ,But my table is gbk-encoded and contains utf8 characters.

For reproduce it , I add postpone file ,and update a row after row-copy complete .

Actually  the column of name's value is `涪陵北`. 

v1.0.40

**table structure**
```
CREATE TABLE `recur_gbk` (
  `id` int(11) NOT NULL AUTO_INCREMENT,
  `Name` varchar(512) DEFAULT NULL,
  `addcol` varchar(255) DEFAULT NULL COMMENT '添加普通列测试',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=32 DEFAULT CHARSET=gbk
```

**gh-ost command**
```
gh-ost --host=10.28.77.43 --database="ghostdb" --table="recur_gbk" --debug --alter="ADD COLUMN addcol1 VARCHAR(255) DEFAULT NULL COMMENT '添加普通列测试'" --concurrent-rowcount --conf="/usr/local/gh-ost.cnf" --port=55944 --execute --postpone-cut-over-flag-file="/tmp/r.postpone" --initially-drop-old-table
```

**log**
```
2018-01-02 19:03:44 INFO starting gh-ost 1.0.40
2018-01-02 19:03:44 INFO Migrating `ghostdb`.`recur_gbk`
2018-01-02 19:03:44 INFO connection validated on 10.28.77.43:55944
2018-01-02 19:03:44 INFO User has ALL privileges
2018-01-02 19:03:44 INFO binary logs validated on 10.28.77.43:55944
2018-01-02 19:03:44 INFO Restarting replication on 10.28.77.43:55944 to make sure binlog settings apply to replication thread
2018-01-02 19:03:44 DEBUG Replication restarted
2018-01-02 19:03:44 INFO Inspector initiated on SVR14883DE630:55944, version 5.7.18-ctrip-log
2018-01-02 19:03:44 INFO Table found. Engine=InnoDB
2018-01-02 19:03:44 DEBUG Estimated number of rows via STATUS: 19
2018-01-02 19:03:44 DEBUG Validated no foreign keys exist on table
2018-01-02 19:03:44 DEBUG Validated no triggers exist on table
2018-01-02 19:03:44 INFO Estimated number of rows via EXPLAIN: 19
[...]

PR #560, opened 2018-03-12
Gbk encoding simple

shlomi-noach:
Context: https://github.com/github/gh-ost/issues/532, see also https://github.com/github/gh-ost/pull/559

Some time after #533 was submitted by @ceshihao we [found](https://github.com/github/gh-ost/pull/533#issuecomment-371382744) that `gkb` charset seems to work without specifying risky-charset argument. This PR is a stripped down version of #533 in an attempt to find the minimal set of changes where `gbk` is still supported.


shlomi-noach:
Interestingly tests pass even though we use `interpolateParams=true`:

https://github.com/github/gh-ost/blob/12424e94fb2cffceefb44f9d3ab20c2266a1ef2b/vendor/github.com/go-sql-driver/mysql/utils.go#L149-L151

Seems like as long as not specified via DSN, `gbk` is good to go.

ceshihao:
@shlomi-noach 

I think it is because `gh-ost` DSN only contains `interpolateParams ` and `charset`, but NOT `collation`.

From the code you linked, the validation is on `interpolateParams ` and `collation`.

shlomi-noach:
@ceshihao makes sense. Can you possibly test this branch on your side and confirm it works as expected?

ceshihao:
@shlomi-noach It works on my side.

BTW
I think this change is not needed either.

https://github.com/github/gh-ost/pull/560/files#diff-0ec6a3deacd5e2577724a82d8347226fR59

shlomi-noach:
Great, thank you for testing! And thank you for your efforts.

> BTW I think this change is not needed either.

Yep. I thought I might leave that as a strict reminder about `interpolateParams`.

PR #562, opened 2018-03-13
Documenting limitation: encrypted binlogs

### LINKED ISSUE #561: binlog encryption?
I saw that Percona Server 5.7 has a binlog encryption feature -- https://www.percona.com/blog/2018/03/08/binlog-encryption-percona-server-mysql/

It's currently in beta, but if it gains traction we'll need to either say we don't support it and/or get the go mysql parser to support it. But temporarily it might be useful to add it to the docs somewhere?

shlomi-noach:
Closes #561 

Doc update to clarify that Percona Server binlog encryption is not supported.


zmoazeni:
You have to be intentional about turning on binlog encryption for a replica too right? 

If so, should we treat this like we do with the RBR? gh-ost can work against a replica with unencrypted binlogs (doesn't matter if upstream encrypts them).

_I could be wrong and it might be encrypted all the way through, but that doesn't seem likely._

shlomi-noach:
@zmoazeni good point! Makes sense to me. 

> I could be wrong and it might be encrypted all the way through, but that doesn't seem likely.

I'll ask around.

timvaillancourt (approved):

Issue #558, opened 2018-03-06
Support exponential backoff for cutover attempts


kurtzur:
Hello, we're using this library at Gusto. My teammate @nikhilmat has contributed here previously.

I'd like `gh-ost` to have a configuration option that enables exponential backoff for cutover attempts. From my reading, it seems the current behavior is that [`retryOperation`](https://github.com/github/gh-ost/blob/f0c3028f27d9fafc1f511a29d3105e9ada2798be/go/logic/migrator.go#L133) waits a constant 1 second in between cutover attempts. In our production environment, we've observed that waiting exponentially longer times between each attempt would result in a better DB load profile. One reason for this is that we have periodic background jobs that hammer certain tables at certain times. I expect that's a relatively common scenario for others as well.

I'd be happy to implement this change myself and submit a PR for it. I wanted to first verify with this issue that it's functionality you're open to adding in.

shlomi-noach:
@kurtzur thank you, this makes sense and we're happy to receive a PR!

PR #563, opened 2018-03-16
Support exponential backoff for cutover attempts

kurtzur:
Related issue: https://github.com/github/gh-ost/issues/558

### Description

This PR introduces the flags `cut-over-exponential-backoff` and `cut-over-exponential-backoff-max-interval`. `cut-over-exponential-backoff` is a boolean flag which, if set to `true`, spaces out cutover attempts using an exponential backoff algorithm. This flag defaults to `false`.`cut-over-exponential-backoff-max-interval` is a numeric flag which dictates the upper bound that the exponentially increasing attempt spacing should obey. This flag defaults to 64 seconds, or 8 total attempts. I enforced a minimum of 2 seconds for this value, which would correspond to 3 total attempts. I added this validation because enabling exponential backoff with fewer than 3 attempts seems both pointless and likely accidental.

The exponential backoff is implemented as a separate function `retryOperationWithExponentialBackoff`. Like `retryOperation`, this function would be reusable in other parts of the project if desired.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -100,30 +100,30 @@ type MigrationContext struct {
 	CliMasterUser     string
 	CliMasterPassword string
 
-	HeartbeatIntervalMilliseconds        int64
-	defaultNumRetries                    int64
-	ChunkSize                            int64
-	niceRatio                            float64
-	MaxLagMillisecondsThrottleThreshold  int64
-	throttleControlReplicaKeys           *mysql.InstanceKeyMap
-	ThrottleFlagFile                     string
-	ThrottleAdditionalFlagFile           string
-	throttleQuery                        string
-	throttleHTTP                         string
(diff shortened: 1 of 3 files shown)


kurtzur:
@shlomi-noach I made the updates you suggested, with one exception: I don't think we should rename the flag `cut-over-exponential-backoff` to `exponential-backoff`. That's because my changeset only respects this boolean flag for the cut-over operation specifically. It would be misleading to give it a generic name. It might be nice to have a globally obeyed `exponential-backoff` flag, but I'd have to find and update each place where it should be obeyed.

shlomi-noach:
Sent to testing.

shlomi-noach:
@kurtzur ah, my mistake! Will merge to `master`. Thank you for pointing this out.

shlomi-noach:
Merged into `master`

Issue #573, opened 2018-04-18
silent table corruption with nullable blob/text columns and binlog_row_image=NOBLOB


Roguelazer:
I have identified a table corruption bug in 1.0.42 (haven't had a chance to test 1.0.45 yet) when binlog_row_image is set to NOBLOB (which we've had as our default for multiple years). gh-ost will sometimes (unpredictably) synthesize NULL values for nullable BLOB/TEXT columns when processing a binlog UPDATE which does not alter that column (and which thus does not contain a field for that column in the binlog).

Reproduction steps:

 - start a local mysql instance (I used 5.6.37) and set binlog_row_image to NOBLOB
 - create a mysql database named "ghtest"
 - start the script from https://gist.github.com/Roguelazer/79e0eac7d3361cca25d00b12dee9932a
 - once it prints out 'created', run `gh-ost -alter 'DROP COLUMN to_drop' -table ghtest -database ghtest -allow-on-master -chunk-size 100 -nice-ratio 0.5 -execute` (the chunk size and nice ratio are just to ensure that gh-ost runs slowly enough to reproduce this)

When gh-ost finishes, the rows that were updated during the gh-ost run will have had their blob columns nulled out silently.

I see #230 ; it should probably be clarified that this tool also causes table corruption with NOBLOB and gh-ost should probably check `@@binlog_row_image` before running... or, better yet, bail if it a column isn't set in the bitmap in the UPDATE_ROWS_EVENT structure.

shlomi-noach:
Thank you for the bug report.

At this time g`gh-ost` only supports `binlog_row_image=FULL`. This should be both clarified in the docs and enforced by the app upon startup. 

Which also means I will not look into the particular bug you have thoughtfully described - thank you!

shlomi-noach:
Looking into the code this situation should already be handled well. Can you please paste the complete command line you executed?

Also, can you please paste the output log (ass `--debug`), from the beginning, and up to the first data copy step?

Roguelazer:
@shlomi-noach the complete command line is included in the reproduction steps above. Full output is at https://gist.github.com/Roguelazer/c279dfe2f74bdaa4b389890341a6c95f.

shlomi-noach:
Thank you again for your elaborate report. Confirmed.

#575 enforces `binlog_row_image` to be `FULL`.

PR #575, opened 2018-04-24
Enforcing binlog_row_image=FULL

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/573

This PR enforces `binlog_row_image=FULL` on inspected server. Previously, the value was enforces only when the chose key wasn't the `PRIMARY KEY`, but in fact must be enforced at all times, or else `gh-ost` doesn't have enough information on `UPDATE` operations on missing columns.


Roguelazer:
hurray! thanks!

PR #586, opened 2018-05-04
go/base: Do not ignore f.Close error

SchumacherFM:
### Description

Do not ignore os.File.Close error.


shlomi-noach:
sent to testing

Issue #571, opened 2018-04-17
build.sh fails if you're running go version 1.10


FortuneOSarcasm:
When compiling gh-ost the version check fails if you're running go version 1.10 or higher.

shlomi-noach:
Addressed by https://github.com/github/gh-ost/pull/589

PR #589, opened 2018-05-07
build.sh: supporting go1.9 and above

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/571

`gh-ost` requires `1.9` or above, and supports `1.10` - `1.19` (at this time latest go version is `1.10`)

cc @FortuneOSarcasm


shlomi-noach:
> This section of the README could be updated

done, thank you

PR #595, opened 2018-05-22
Support for GENERATED (aka virtual) columns

shlomi-noach:
With this PR `gh-ost` supports [generated columns](https://dev.mysql.com/doc/refman/5.7/en/create-table-generated-columns.html). It is based on @dveeden 's https://github.com/github/gh-ost/pull/140, but takes a slightly different approach: generated columns on original/ghost table are not considered to be _shared_ between the tables. This simplifies query generation & choice of unique key.


shlomi-noach:
Closes https://github.com/github/gh-ost/issues/138

shlomi-noach:
Noteworthy that questions raised in #138 are irrelevant:

```
Error 3106: 'Changing the STORED status' is not supported for generated columns.
```

- It is impossible to change a column from non-generated to generated
- It is impossible to change a column from generated to non-generated

shlomi-noach:
It is OK to rename a generated column.

Issue #610, opened 2018-07-23
gh-ost occasionally hangs at waitForEventsUpToLock


wfxiang08:
I use gh-ost to migrate tables on my macbook pro, and occasionally find  it hands when at waitForEventsUpToLock

repeatedly print the logs below

> Waiting for events up to lock
> Timeout while waiting for events up to lock

The environment is：
- go version go1.10.3 darwin/amd64
- MacOS Sierra 10.12.6
- MySQL 5.7

After debuging, I found be deadlocks in `migrator.go#iterateChunks#copyRowsFunc#CalculateNextIterationRangeEndValues`

The code in problem is as below:
```go
go this.executeWriteFuncs()
go this.iterateChunks()
if err := retrier(this.cutOver); err != nil {
}
```

cutOver#cutOverTwoStep#LockOriginalTable might lock the Original table before the last call of CalculateNextIterationRangeEndValues, then deadlock might happen.

shlomi-noach:
Closed by #611

PR #611, opened 2018-07-23
fix copyRowsFunc hangs bug

wfxiang08:
Related issue: https://github.com/github/gh-ost/issues/610
### Description

When copyRowsFunc finishes copy rows,  the status `hasFurtherRange` is kept across different iterations of copyRowsFunc, and when `hasFurtherRange` is false, never try to call `CalculateNextIterationRangeEndValues`. 

When `hasFurtherRange` is false, the original table also might be locked and in this case `CalculateNextIterationRangeEndValues` will be hang forever.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1087,24 +1087,30 @@ func (this *Migrator) iterateChunks() error {
 		log.Debugf("No rows found in table. Rowcopy will be implicitly empty")
 		return terminateRowIteration(nil)
 	}
+
+	var hasNoFurtherRangeFlag int64
 	// Iterate per chunk:
 	for {
-		if atomic.LoadInt64(&this.rowCopyCompleteFlag) == 1 {
+		if atomic.LoadInt64(&this.rowCopyCompleteFlag) == 1 || atomic.LoadInt64(&hasNoFurtherRangeFlag) == 1 {
 			// Done
 			// There's another such check down the line
 			return nil
 		}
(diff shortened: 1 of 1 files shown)


go/logic/migrator.go
```diff
@@ -1135,6 +1140,9 @@ func (this *Migrator) iterateChunks() error {
 		}
 		// Enqueue copy operation; to be executed by executeWriteFuncs()
 		this.copyRowsQueue <- copyRowsFunc
+		if !hasFurtherRange {
+			break
+		}
```
shlomi-noach: Since this code is outside `copyRowsFunc`, then it only skips enqueuing of new `copyRowsFunc` functions, but still there could be multiple `copyRowsFunc` instacnes inside `this.copyRowsQueue` that haven't been evaluated yet.

This is why I believe this PR _mitigates_ the potential for deadlock but does not actually _solve it_.

I think we should check `hasFurtherRange` from _within_ `copyRowsFunc` to solve it completely.

wfxiang08: @shlomi-noach yes, you are right. I will make some modification.

wfxiang08:
@shlomi-noach I have make some modification, will you pls. have a review.

shlomi-noach:
Apologies for the time it took, I was sidetracked. On the flip side, this PR has been actively tested for some 50 days, so it's been well hammered. Merging!

Issue #483, opened 2017-08-31
gh-ost  miss data  When  I alter table 


yuan-make:
> This is the place to report a bug, ask a question, or suggest an enhancement.  

**Dead locjk Happend,when i using  gh-ost  alter tale**

show engin innodb  status\G
```
2017-08-31 11:28:44 7f28b99ff700
*** (1) TRANSACTION:
TRANSACTION 32091935401, ACTIVE 0 sec fetching rows
mysql tables in use 2, locked 2
LOCK WAIT 47 lock struct(s), heap size 6544, 172 row lock(s), undo log entries 128
MySQL thread id 821336333, OS thread handle 0x7f298f7be700, query id 12799839393 10.205.134.13 make Sending data
insert /* gh-ost `db1`.`test1` */ ignore into `db1`.`_test1_gho` (`sa_id`, `sa_cg_id`, `sa_sb_id`, `sa_ss_id`, `sa_si_id`, `sa_brand_id`, `sa_act_id`, `sa_code`, `sa_uid`, `sa_username`, `sa_cat_id`, `sa_cat_topid`, `sa_cat_threeid`, `sa_tao_type`, `sa_tao_iid`, `sa_link`, `sa_title`, `sa_title_long`, `sa_pic`, `sa_pic_long`, `sa_price`, `sa_cprice`, `sa_inventory`, `sa_add_time`, `sa_start_time`, `sa_end_time`, `sa_show_stime`, `sa_show_etime`, `sa_sort`, `sa_status`, `sa_is_onsale`, `sa_is_soldout`, `sa_free_post`, `sa_free_shipping`, `sa_country`, `sa_province`, `sa_city`, `sa_deliver_time`, `sa_limit_buy`, `sa_rcmd`, `sa_back`, `sa_back_img`, `sa_lastmodify`, `sa_isbusiness`, `sa_isdemo`, `sa_isinformation`, `sa_is_add`, `sa_is_advance`, `sa_is_push`, `sa_is_hot`, `sa_to_business`, `sa_to_operate`, `sa_timelimit`, `sa_bc_id`, `sa_cga_id`, `sa_cd_id`, `sa_con
*** (1) WAITING FOR THIS LOCK TO BE GRANTED:
RECORD LOCKS space id 1992 page no 3844625 n bits 72 index `PRIMARY` of table `db1`.`test1` trx id 32091935401 lock mode S waiting
*** (2) TRANSACTION:
[...]

PR #616, opened 2018-07-25
fix retry failed bug which might result in early quit of migration

wfxiang08:
Related issue: https://github.com/github/gh-ost/issues/483

### Description
`applyCopyRowsFunc` should be retriable. But when there is error, terminateRowIteration is called, which results in a panic. So terminateRowIteration should not be called inside applyCopyRowsFunc

```go
// Copy task:
applyCopyRowsFunc := func() error {
	if atomic.LoadInt64(&this.rowCopyCompleteFlag) == 1 {
		return nil
	}
	_, rowsAffected, _, err := this.applier.ApplyIterationInsertQuery()
	if err != nil {
		return err // terminateRowIteration shouldn't be called if you want retry
	}
	atomic.AddInt64(&this.migrationContext.TotalRowsCopied, rowsAffected)
	atomic.AddInt64(&this.migrationContext.Iteration, 1)
	return nil
}
if err := this.retryOperation(applyCopyRowsFunc); err != nil {
	return terminateRowIteration(err)
}
```

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1122,7 +1122,7 @@ func (this *Migrator) iterateChunks() error {
 				}
 				_, rowsAffected, _, err := this.applier.ApplyIterationInsertQuery()
 				if err != nil {
-					return terminateRowIteration(err)
+					return err
 				}
 				atomic.AddInt64(&this.migrationContext.TotalRowsCopied, rowsAffected)
 				atomic.AddInt64(&this.migrationContext.Iteration, 1)


wfxiang08:
@shlomi-noach

shlomi-noach:
@wfxiang08 oh, I'm sorry. I got a notification of you closing this issue. I must have missed it initially. I apologize for not responding to this issue. 

Did you close it because you gave up, or because the fix is already done?

Issue #660, opened 2018-11-01
gh-ost is not properly identifying column types


esnunes:
I'm facing an issue when trying to alter a table that contains a `json` column.

Original table
```sql
CREATE TABLE `xxx_shipping` (
  `xxx_id` bigint(20) unsigned NOT NULL,
  `zones` json NOT NULL,
  `updated_at` timestamp NULL DEFAULT NULL,
  PRIMARY KEY (`xxx_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

Final table
```sql
CREATE TABLE `_xxx_shipping_gho` (
  `xxx_id` bigint(20) unsigned NOT NULL,
  `my_new_column` json DEFAULT NULL,
  `zones` json NOT NULL,
  `updated_at` timestamp NULL DEFAULT NULL,
  PRIMARY KEY (`xxx_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
```

At some point during the migration I get the following error message:

```
2018-10-31 14:59:57 ERROR Error 3144: Cannot create a JSON value from a str  
  ing with CHARACTER SET 'binary'.; query=                                     
  			replace /* gh-ost `my_database`.`_xxx_shipping_gho` */ in  
  to                                                                           
  				`my_database`.`_xxx_shipping_gho`                         
  					(`xxx_id`, `zones`, `updated_at`)                                   
  				values                                                                   
  					(?, ?, ?)                                                               
  		; args=[15680576 [123 34 97 100 34 58 91 93 44 34 97 101 34 58 91 93 44 3  
  4 97 102 34 58 91 93 44 34 97 103 34 58 91 93 44 34 97 105 34 58 91 93 44 3  
  4 97 108 34 58 91 93 44 34 97 109 34 58 91 93 44 34 97 111 34 58 91 93 44 3  
  4 97 113 34 58 91 93 44 34 97 114 34 58 91 93 44 34 97 115 34 58 91 93 44 3  
  4 97 116 34 58 91 93 44 34 97 117 34 58 91 93 44 34 97 119 34 58 91 93 44 3
...
```

The `gh-ost` `--alert` parameter is:
```
--alter="add my_new_column json null after xxx_id"
```

Part of the data is properly migrated, usually the migration fails after the first batch. I created a PR already that solves the problem. I will link this issue in the PR.

PR #661, opened 2018-11-01
Fix inspector column types

esnunes:
## Issue

Related issue: https://github.com/github/gh-ost/issues/660

### Description

This PR fix an issue in `Inspector.applyColumnTypes`. The function fails when a `columnList` does not contain all table columns. For some reason the code fails silently, so any other column after the first failure will be identified as type `0` (default `int` value).

As a side-effect when copying rows from original table to `gh-ost` temporary one, it may fail due to bad type conversion as describe in the related issue.

### Notes

I've considered adding tests but they would require extract the column type logic to a different function or add an external library to mock `sql.DB`.


go/logic/inspect.go
```diff
@@ -173,8 +173,7 @@ func (this *Inspector) inspectOriginalAndGhostTables() (err error) {
 	// This additional step looks at which columns are unsigned. We could have merged this within
 	// the `getTableColumns()` function, but it's a later patch and introduces some complexity; I feel
 	// comfortable in doing this as a separate step.
-	this.applyColumnTypes(this.migrationContext.DatabaseName, this.migrationContext.OriginalTableName, this.migrationContext.OriginalTableColumns, this.migrationContext.SharedColumns)
-	this.applyColumnTypes(this.migrationContext.DatabaseName, this.migrationContext.OriginalTableName, &this.migrationContext.UniqueKey.Columns)
+	this.applyColumnTypes(this.migrationContext.DatabaseName, this.migrationContext.OriginalTableName, this.migrationContext.OriginalTableColumns, this.migrationContext.SharedColumns, &this.migrationContext.UniqueKey.Columns)
```
esnunes: This is just a small improvement, instead of calling `applyColumnTypes` twice for the same `databaseName` and `tableName`, I've moved the `&this.migrationContext.UniqueKey.Columns` to this call.

go/logic/inspect.go
```diff
@@ -552,44 +551,35 @@ func (this *Inspector) applyColumnTypes(databaseName, tableName string, columnsL
...
-				columnsList.SetUnsigned(columnName)
+		for _, columnsList := range columnsLists {
+			column := columnsList.GetColumn(columnName)
+			if column == nil {
+				continue
 			}
-		}
-		if strings.Contains(columnType, "mediumint") {
-			for _, columnsList := range columnsLists {
-				columnsList.GetColumn(columnName).Type = sql.MediumIntColumnType
```
esnunes: This code fails when `columnsList.GetColumn(columnName)` returns `nil`. This is going to happen in `this.migrationContext.MappedSharedColumns` and `this.migrationContext.SharedColumns` due to the fact they they only have the intersection of original and ghost tables. If you add a new column by the end of the column list this problem won't be noticed as the code is going to fail when setting the type of the new column (which is not used during copy of data), but if you alter table and add a column in the second position, all other columns will have the default Type (0) because the code will not iterate over all of them.

go/logic/inspect.go
```diff
@@ -552,44 +551,35 @@ func (this *Inspector) applyColumnTypes(databaseName, tableName string, columnsL
 	err := sqlutils.QueryRowsMap(this.db, query, func(m sqlutils.RowMap) error {
 		columnName := m.GetString("COLUMN_NAME")
 		columnType := m.GetString("COLUMN_TYPE")
-		if strings.Contains(columnType, "unsigned") {
-			for _, columnsList := range columnsLists {
-				columnsList.SetUnsigned(columnName)
+		for _, columnsList := range columnsLists {
+			column := columnsList.GetColumn(columnName)
+			if column == nil {
```
esnunes: The for loop was changed in order to prevent accessing a column that is not available in the `columnsList` object.

esnunes: This code ignores columns that are not available in the columnsList and continues iterating over the next columns.

shlomi-noach:
Thank you for replying. I'll take this change to testing. It'll like take a few days or over a week for me to respond to this PR, given my current workload, sorry for the expected delay.

esnunes:
No worries and thanks for the heads up regarding the time to release.

esnunes:
Hey @shlomi-noach 👋, any news on this one? Thanks!

shlomi-noach:
sent to production testing

shlomi-noach:
Meanwhile sorry for the delay.

shlomi-noach:
This looks good in testing and I'll merge this PR early next week.

PR #680, opened 2018-11-20
testing: convert to character set utf8mb4

shlomi-noach:
Adding a test for a `convert to character set utf8mb4` statement, where the table is populated with utf8 data. The test confirms the existing text does not get corrupted.

### DIFF
--- a/localtests/convert-utf8mb4/create.sql
+++ b/localtests/convert-utf8mb4/create.sql
@@ -1,0 +1,25 @@
+drop table if exists gh_ost_test;
+create table gh_ost_test (
+  id int auto_increment,
+  t varchar(128) charset utf8 collate utf8_general_ci,
+  ta varchar(128) charset ascii not null,
+  primary key(id)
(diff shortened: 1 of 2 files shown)


ggunson:
Are we testing a specific utf8->utf8mb4 conversion, or `convert to character set` in general? AFAIK utf8 is a subset of utf8mb4, but there might still be gh-ost users with databases in latin1 needing to go utf8*, and it would be interesting to know if it worked there.

Also, could one convert utf8mb4->utf8, would it behave the same way as a straight alter table re: any characters in the former and not in the latter? 

Have I made this too complicated now? 😄

shlomi-noach:
right now the test tests conversion of both `utf8` and `ascii` to `utf8mb4`. I'll add a `latin1` column.

> Also, could one convert utf8mb4->utf8, would it behave the same way as a straight alter table re: any characters in the former and not in the latter?

For another day... /o\

Issue #681, opened 2018-11-26
Error 1264: Out of range value


mohsen3:
We are trying to use `gh-ost` to add a new column to an existing table with a few hundreds of millions of rows and we are getting an `Error 1264: Out of range value for column '...' at row 1` error message. Unfortunately, I cannot reproduce this error on my dev environment. In the prod environment, the migration goes fine until a new row is inserted into the table (so it seems to me that it has something to do with live updates, although I am just guessing since I am not too familiar with the MySql replication protocol that gh-ost uses).

The column that is causing this issue is of type `Decimal(65, 30)` and the values in there are pretty large (`10**35-1` which is slightly smaller than the max possible value). Also in production, we have a master/slave setup that I don't have on my dev machine. Does any of these ring a bell? Is this a known issue in `gh-ost`?

mohsen3:
This is more of a question right now, but it seems that I don't have the permission to label it so.

mohsen3:
@ggunson Thank you for your quick response. This is a blocker for us right now. So, if there is something we can help with, please let me know. We are trying to reproduce it locally right now that may help you pin pointing the issue as well.

Also, does it sound like a reasonable theory to you? If yes, we can stop insertions into that table for a while and does the migration. It means readonly access to the table for a few hours (which is not ideal, but acceptable to us if it works). Do you have any documentation on that issue that helps us understand or debug the issue better?

ggunson:
> Also, does it sound like a reasonable theory to you? 

It does, because we've seen the same sort of error message in regards to the binary log parsing. For example, https://github.com/github/gh-ost/issues/157

Also, we don't use `DECIMAL` columns much at GitHub so if there's a bug we would likely not have seen it in our internal testing infrastructure. 

If you're not altering that specific decimal column, I would guess that pausing writes to that table would allow gh-ost to run (or, if it failed it would be on something else). 

Could you please provide (so we can test this):

1. mysql version of master and replicas (if both are being used to run gh-ost)
1. gh-ost version
1. gh-ost command
1. `show create table` for the table being migrated
1. the column value it's failing on (if you know it)

mohsen3:
We are actually making those columns nullable (before migration, they are not nullable).
Here is the answers to your questions:

> mysql version of master and replicas (if both are being used to run gh-ost)

```
5.6.39
```



> gh-ost version

```
1.0.44
```

> gh-ost command

```
[...]

PR #684, opened 2018-11-28
Fixed DECIMAL datatype

shlomi-noach:
Testing https://github.com/github/gh-ost/issues/681

Initial commit confirms the bug reported in https://github.com/github/gh-ost/issues/681


shlomi-noach:
@Tomas2046 please explain your comments to prove you're not a bot.

shlomi-noach:
Updated underlying `go-mysql` library to get the fixes in https://github.com/siddontang/go-mysql/pull/332.

However, had to re-apply `TIMESTAMP` and `DATETIME` patches; `go-mysql` uses a new mechanism for parsing both old and new `TIMESTAMP` and `DATETIME` which seem to break `gh-ost`'s tests whenever a non-UTC timezone is involved. I haven't been successful in finding out exactly why this happens, and so I've re-applied old temporal logic on top of the new changes.

shlomi-noach:
tested in prod, looking good.

shlomi-noach:
I'll release this/next week

PR #695, opened 2019-01-14
Adding --force-named-panic option

shlomi-noach:
When `--force-named-panic` is given, a `panic` command must also indicate the name of migrated table, or else it is ignored.
This can be useful in mass-distributing a `panic` request to potentially multiple `gh-ost` migrations, when in fact only seeking to terminate a specific migration.


shlomi-noach:
This PR also supports non-forced named `throttle`. i.e. you may for example `echo throttle table_name | socat - /tmp/gh-ost.sock`, and the migration will only throttle on a table name match.

Issue #698, opened 2019-01-22
inspect.getCandidateUniqueKeys avoidably slow on large schemas


twotwotwo:
I'm looking at gh-ost for use on a multitenant app where we have tens thousands of tables on one server. Because number of tables and columns, a couple of inspect.go INFORMATION_SCHEMA queries take more than a minute. 

It looks like for getCandidateUniqueKeys' big query, the slowness is easy to fix: the ON clause on the join checks that TABLE_SCHEMA and TABLE_NAME match, when that isn't necessary because in the inner and outer queries both check for the right TABLE_NAME and TABLE_SCHEMA in their WHERE clauses. Without those additional conditions, MySQL 5.6 is able to run the query in seconds.

Without the change we'd spend hours just on gh-ost startup time on every migration. With that change and -skip-foreign-key-checks, it's much more workable for us. I have a simple change ready that I'll submit as a PR; writing here because the PR instructions suggest starting discussions on issues.

PR #699, opened 2019-01-22
inspect: remove redundant join conditions

twotwotwo:
Related issue: https://github.com/github/gh-ost/issues/698

### Description

The TABLE_SCHEMA and TABLE_NAME in the INFORMATION_SCHEMA query in getCandidateUniqueKeys are already guaranteed to have the same value in COLUMNS and UNIQUES because of the WHEREs in each query. Dropping it from the ON clause makes it complete much faster.

On (at least) MySQL 5.6 installs with thousands of tables, this query completes in a few seconds without the additional join conditions, but takes more than a minute with it.


shlomi-noach:
closes #698

Issue #521, opened 2017-11-13
Add support for SSL


igroene:
I came across some use cases where strict compliance rules enforce that all connections to the db's need to use SSL. It would be nice if gh-ost supported that option.
It looks like go-mysql-driver already has support (https://github.com/go-sql-driver/mysql/pull/89)

Also for reference: https://dev.mysql.com/doc/refman/5.7/en/using-encrypted-connections.html

zmoazeni:
We have been looking into this as well. One option Percona mentioned to me was to define the SSL options in the my.cnf and set the user you use to require SSL.

I haven't had a chance to test this yet but it has been on my radar. You may also need to setup the existing replication user the same way.

For what it's worth, I've also been thinking about this in the context of orchestrator (failing over replication using SSL when that happens)

igroene:
> We have been looking into this as well. One option Percona mentioned to me was to define the SSL options in the my.cnf and set the user you use to require SSL.

That won't work for gh-ost as it doesn't use the standard mysql client to connect

zmoazeni:
Ah that makes sense. That might only work for orchestrator issuing `CHANGE MASTER TO...` without supplying the SSL options. (Since it's `mysqld` connecting to `mysqld`). Bummer.

shlomi-noach:
Well, orchestrator uses the exact same client to connect to the servers via SSL. It just supports SSL code-wise, as per https://github.com/github/orchestrator/blob/master/docs/ssl-and-tls.md , and we'd need to apply roughly the same to `gh-ost`.

wesleyk:
Any update on this request @shlomi-noach? If no one has started on this, is there perhaps a sample PR from orchestrator that you could link here that we might use to add SSL support in?

Thanks!

shlomi-noach:
@wesleyk I have no immediate plan to work on this. Relevant `orchestrator` PRs:

- https://github.com/github/orchestrator/pull/155
- https://github.com/github/orchestrator/pull/242
- https://github.com/github/orchestrator/pull/307

brandonbodnar-wk:
We experienced a recent need for SSL support while using the gh-ost tool on AWS RDS instances. I've submitted #705 as an initial pass of what I think is the minimum support we need for our use case.  While it doesn't cover every ssl setting that users might want to select, it can provide at least a starting point to adding in some level of SSL/TLS support into gh-ost.

RafeKettler:
[...]

PR #705, opened 2019-01-31
Initial SSL Connection Support

brandonbodnar-wk:
Related issue: https://github.com/github/gh-ost/issues/521

### Description

This PR adds initial support for enforcing TLS/SSL connections for the databases involved in the schema changes. In this initial pass, a gh-ost user can enforce ssl be used for connections using the `--ssl` flag on the command line. If the user wishes to enforce checking validity of the database's certificate, the user can also specify a CA certificate in pem format with the `--ssl-ca` option.

Adding these two options is sufficient for our use case right now in connecting to [AWS RDS instances](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/CHAP_MySQL.html#MySQL.Concepts.SSLSupport). While there are plenty more options that can be specified by the native mysql client regarding ssl, this can provide a good starting point for other work by those with more complex needs.

### Testing

We tested this updated code against an Aurora2 cluster running a MySQL 5.7 compatible engine. 
-  For testing we added a user to the mysql cluster with `REQUIRE SSL` set. 
- We then attempted to run an alter table command using `gh-ost` against the cluster, but without the new ssl options enabled. As expected we received an access denied when attempting to connect with `gh-ost`. 
- We then attempted the same alter table command using all the same options, except also adding on the new `--ssl` flag, along with specifying the `--ssl-ca` certificate for AWS RDS. The job executed flawlessly.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -94,16 +94,16 @@ type MigrationContext struct {
 	AliyunRDS                bool
 	GoogleCloudPlatform      bool
 
-	config                ContextConfig
-	configMutex           *sync.Mutex
-	ConfigFile            string
-	CliUser               string
-	CliPassword           string
-	UseTLS                bool
-	TLSInsecureSkipVerify bool
-	TLSCACertificate      string
-	CliMasterUser         string
-	CliMasterPassword     string
+	config            ContextConfig
+	configMutex       *sync.Mutex
+	ConfigFile        string
+	CliUser           string
+	CliPassword       string
+	UseTLS            bool
+	TLSAllowInsecure  bool
+	TLSCACertificate  string
+	CliMasterUser     string
+	CliMasterPassword string
 
 	HeartbeatIntervalMilliseconds       int64
 	defaultNumRetries                   int64
@@ -700,7 +700,7 @@ func (this *MigrationContext) ApplyCredentials() {
 
 func (this *MigrationContext) SetupTLS() error {
 	if this.UseTLS {
-		return this.InspectorConnectionConfig.UseTLS(this.TLSCACertificate)
+		return this.InspectorConnectionConfig.UseTLS(this.TLSCACertificate, this.TLSAllowInsecure)
 	}
 	return nil
 }
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -99,6 +99,8 @@ type MigrationContext struct {
 	ConfigFile        string
 	CliUser           string
 	CliPassword       string
+	UseTLS            bool
+	TlsCACertificate  string
 	CliMasterUser     string
 	CliMasterPassword string
 
@@ -693,6 +695,13 @@ func (this *MigrationContext) ApplyCredentials() {
 		// Override
 		this.InspectorConnectionConfig.Password = this.CliPassword
 	}
+}
+
+func (this *MigrationContext) SetupTLS() error {
+	if this.UseTLS {
+		return this.InspectorConnectionConfig.UseTLS(this.TlsCACertificate)
+	}
+	return nil
 }
 
 // ReadConfigFile attempts to read the config file, if it exists
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -94,15 +94,16 @@ type MigrationContext struct {
 	AliyunRDS                bool
 	GoogleCloudPlatform      bool
 
-	config            ContextConfig
-	configMutex       *sync.Mutex
-	ConfigFile        string
-	CliUser           string
-	CliPassword       string
-	UseTLS            bool
-	TLSCACertificate  string
-	CliMasterUser     string
-	CliMasterPassword string
+	config                ContextConfig
+	configMutex           *sync.Mutex
+	ConfigFile            string
+	CliUser               string
+	CliPassword           string
+	UseTLS                bool
+	TLSInsecureSkipVerify bool
+	TLSCACertificate      string
+	CliMasterUser         string
+	CliMasterPassword     string
 
 	HeartbeatIntervalMilliseconds       int64
 	defaultNumRetries                   int64
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -100,7 +100,7 @@ type MigrationContext struct {
 	CliUser           string
 	CliPassword       string
 	UseTLS            bool
-	TlsCACertificate  string
+	TLSCACertificate  string
 	CliMasterUser     string
 	CliMasterPassword string
 
@@ -699,7 +699,7 @@ func (this *MigrationContext) ApplyCredentials() {
 
 func (this *MigrationContext) SetupTLS() error {
 	if this.UseTLS {
-		return this.InspectorConnectionConfig.UseTLS(this.TlsCACertificate)
+		return this.InspectorConnectionConfig.UseTLS(this.TLSCACertificate)
 	}
 	return nil
 }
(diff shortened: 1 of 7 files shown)


go/base/context.go
```diff
@@ -99,6 +99,8 @@ type MigrationContext struct {
 	ConfigFile        string
 	CliUser           string
 	CliPassword       string
+	UseTLS            bool
+	TlsCACertificate  string
```
tomdeering-wf: Typical Go style would have this named `TLSCACertificate`
https://github.com/golang/go/wiki/CodeReviewComments#initialisms

go/logic/applier.go
```diff
@@ -73,7 +73,7 @@ func (this *Applier) InitDBConnections() (err error) {
 	if this.db, _, err = mysql.GetDB(this.migrationContext.Uuid, applierUri); err != nil {
 		return err
 	}
-	singletonApplierUri := fmt.Sprintf("%s?timeout=0", applierUri)
+	singletonApplierUri := fmt.Sprintf("%s&timeout=0", applierUri)
```
brandonbodnar-wk: This was an existing bug we found when adding the tls option onto the uri. Previously, this was causing the uri to take the form of `gromit:penguin@tcp(myhost:3306)/test?interpolateParams=true&autocommit=true&charset=utf8mb4,utf8,latin1?timeout=0`. The DSN parser for go-mysql-driver/mysql was then not parsing the timeout setting, but instead including that as part of the value for `charset`.

go/mysql/connection.go
```diff
@@ -42,13 +50,42 @@ func (this *ConnectionConfig) Duplicate() *ConnectionConfig {
...
+
+	this.tlsConfig = &tls.Config{
+		RootCAs:            rootCertPool,
+		InsecureSkipVerify: skipVerify,
+	}
+
+	if err := mysql.RegisterTLSConfig(this.Key.StringCode(), this.tlsConfig); err != nil {
+		return err
+	}
+	return nil
```
tomdeering-wf: Any reason not to return the result of the function call directly?

go/cmd/gh-ost/main.go
```diff
@@ -14,6 +14,7 @@ import (
 
 	"github.com/github/gh-ost/go/base"
 	"github.com/github/gh-ost/go/logic"
+	_ "github.com/go-sql-driver/mysql"
```
brandonbodnar-wk: Added this to make it clear that gh-ost was registering the `go-sql-driver/mysql` driver as its `mysql` driver. We had some confusion previously and thought that the driver might have been coming from the `siddontang/go-mysql` library referenced in the `go/binlog/gomysql_reader.go`. But, the driver was in fact the `go-sql-driver/mysql`, pulled in though an import statement located within `github.com/outbrain/golib/sqlutils`. Hoping that adding the import directly here instead of relying on the indirect import will make it more clear for others.

go/mysql/connection.go
```diff
@@ -57,5 +94,9 @@ func (this *ConnectionConfig) GetDBUri(databaseName string) string {
 		hostname = fmt.Sprintf("[%s]", hostname)
 	}
 	interpolateParams := true
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?interpolateParams=%t&autocommit=true&charset=utf8mb4,utf8,latin1", this.User, this.Password, hostname, this.Key.Port, databaseName, interpolateParams)
+	tlsOption := "false"
```
tomdeering-wf: It might be worth leaving a comment that "false" is defaulted, so your explicit use of it is for code path simplicity. Up to you.

go/cmd/gh-ost/main.go
ericklaus-wf: Should we add a comment explaining why we're importing this?

go/logic/applier.go
johnlockwood-wf: What if `applierUri` has no query string yet?

brandonbodnar-wk: @johnlockwood-wf `applierUri` has query strings from the return value of [ConnectionConfig.GetDBUri](https://github.com/github/gh-ost/pull/705/files#diff-0ec6a3deacd5e2577724a82d8347226fR101) both before and after this PR.

go/cmd/gh-ost/main.go
brandonbodnar-wk: I don't believe so. Not use to seeing comments within import statements, and this brings the usage much closer to the standard usage described in the [go-sql-driver/mysql](https://github.com/go-sql-driver/mysql#usage) documentation. Potentially we could move this into the one of the classes where `database/sql` is being imported to exactly match the documentation, but that seems less discoverable to me than adding here in main.

go/logic/applier.go
brandonbodnar-wk: I rather not refactor every place where GetDBUri is currently being used to address this bug. Because we are adding the tls param into GetDBUri, we are directly impacted by this line using the wrong symbol and thus I am including the fix here to use the correct symbol.

Additionally, these are not really URI's but DSNs recognized by [go-my-sql/driver](https://github.com/go-sql-driver/mysql#dsn-data-source-name). While [net/url.Values](https://golang.org/pkg/net/url/#Values) may have some helpful functions, I don't think the work involved in shoehorning that in is worthwhile here where each code path that hits this spot already has params set for `applierUri`.

brandonbodnar-wk: go-sql-driver's [Config.FormatDSN](https://godoc.org/github.com/go-sql-driver/mysql#Config.FormatDSN) would likely be a great improvement on the manually created DSN, but I think that is something to reserve for a separate issue or PR.

go/cmd/gh-ost/main.go
```diff
@@ -238,6 +245,7 @@ func main() {
 	migrationContext.SetThrottleHTTP(*throttleHTTP)
 	migrationContext.SetDefaultNumRetries(*defaultRetries)
 	migrationContext.ApplyCredentials()
+	migrationContext.SetupTLS()
```
natewernimont-wk: Should this check the error result and log something if an error is encountered?

go/mysql/connection.go
```diff
@@ -16,6 +22,7 @@ type ConnectionConfig struct {
 	User       string
 	Password   string
 	ImpliedKey *InstanceKey
+	tlsConfig  *tls.Config
```
natewernimont-wk: Any reason not to make this variable public since you have a simple getter down below anyway?

go/cmd/gh-ost/main.go
```diff
@@ -54,6 +55,9 @@ func main() {
 	flag.StringVar(&migrationContext.ConfigFile, "conf", "", "Config file")
 	askPass := flag.Bool("ask-pass", false, "prompt for MySQL password")
 
+	flag.BoolVar(&migrationContext.UseTLS, "ssl", false, "Enable SSL encrypted connections to MySQL")
+	flag.StringVar(&migrationContext.TLSCACertificate, "ssl-ca", "", "CA certificate in PEM format for TLS connections. Requires --ssl")
```
natewernimont-wk: Should there be a note that these settings will apply to both the replica and master with no option to differentiate them?

go/mysql/connection.go
brandonbodnar-wk: Wanted to force callers consuming this struct to set up the TLS for the ConnectionConfig through the `UseTLS` function below if they wanted to set up the tls.Config. This prevents passing around a ConnectionConfig whose tls.Config has not be registered with the mysql driver.

brandonbodnar-wk:
@shlomi-noach If there is anything I can add to this PR that would make review on the maintainers side easier, please let me know. Adding SSL support is a priority for us and I would like to make it as easy as possible for you.

go/logic/applier.go
shlomi-noach: Agreed. Let's postpone use of `FormatDNS` to a next PR.

brandonbodnar-wk:
Thanks @shlomi-noach. I've added documentation for the flags into command-line-flags.md in this latest commit.

shlomi-noach:
Thank you

Issue #730, opened 2019-04-18
script/cibuild - cannot execute binary file errors


bancer:
I've got errors while running `script/cibuild` command after I cloned gh-ost repository to my Raspberry Pi machine:
```
pi@raspberrypi:~/git/gh-ost $ script/cibuild
go: cannot find GOROOT directory: /home/pi/git/gh-ost/.vendor/go19/go
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100 99.4M  100 99.4M    0     0  4359k      0  0:00:23  0:00:23 --:--:-- 5305k
de874549d9a8d8d8062be05808509c09a88a248e77ec14eb77453530829ac02b  go1.9.2.linux-amd64.tar.gz
script/ensure-go-installed: line 46: /home/pi/git/gh-ost/.vendor/go19/go/bin/go: cannot execute binary file: Exec format error
 installed in /home/pi/git/gh-ost/.vendor/go19: Go Binary: /home/pi/git/gh-ost/.vendor/go19/go/bin/go
++ rm -rf .gopath
++ mkdir -p .gopath/src/github.com/github
++ ln -s /home/pi/git/gh-ost .gopath/src/github.com/github/gh-ost
++ export GOPATH=/home/pi/git/gh-ost/.gopath:/home/pi/git/gh-ost/.vendor
++ GOPATH=/home/pi/git/gh-ost/.gopath:/home/pi/git/gh-ost/.vendor
+ echo 'Verifying code is formatted via '\''gofmt -s -w  go/'\'''
Verifying code is formatted via 'gofmt -s -w  go/'
+ gofmt -s -w go/
script/cibuild: line 8: /home/pi/git/gh-ost/.vendor/go19/go/bin/gofmt: cannot execute binary file: Exec format error

```

shlomi-noach:
`gh-ost` not supported on rasperry pi.

kazndillski:
What is this

bancer:
Is there a good reason for that? I installed go on raspberry pi without any problems.

shlomi-noach:
@bancer there is no good reason other than not having attempted this; I'm not sure how to check why it's not workign as I don't own a raspberry pi myself, nor do I see how we will integrate Raspberry Pi tests in our CI/CD. And so the conscience choice is to not support it ("support" means we test it and verify it works).
[...]

Issue #425, opened 2017-05-29
gh-ost-test-mysql-master for localtests/test.sh missing?


shuhaowu:
I'm trying to setup ghost locally and I noticed that `localtests/test.sh` doesn't seems to work due to the missing command `gh-ost-test-mysql-master` and `gh-ost-test-mysql-replica`. The specific error while running this is:

```
localtests/test.sh: line 23: gh-ost-test-mysql-master: command not found
Cannot verify gh-ost-test-mysql-master
```

From what I can tell in the code, it seems like this may be an alias to the master mysql/replica instance (perhaps via mysql -u .. -h ..). However, I don't see it being defined anywhere in the code. Is this a bug?

shlomi-noach:
@shuhaowu thank you, and you are correct on all account. This isn't strictly a bug, but rather some technical debt in implementing this on TravisCI or otherwise on a generic CI.

The deal is that to run the `test.sh` suite you need a master-replica setup. So you need two servers.

The two commands are either aliases or scripts; here's how both look on my personal dev box:

```
~$ cat bin/gh-ost-test-mysql-master
#!/bin/bash

/Users/shlomi-noach/sandboxes/rsandbox_mysql-5_6_28/m "$@"

~$ cat bin/gh-ost-test-mysql-replica
#!/bin/bash

/Users/shlomi-noach/sandboxes/rsandbox_mysql-5_6_28/s4 "$@"
```

(I'm using `mysqlsandbox` for my local replication setup).

And so the TravisCI suite does not actually run these tests (for now); I'm doing some manual work still...

shuhaowu:
I actually have a working ish implementation of the localtests on Travis. I'm going to clean it up and send a PR on Monday.

shlomi-noach:
Oh wonderful! Thank you!

PR #731, opened 2019-04-18
Use arm package to install go on raspberry pi

bancer:
Related issue: https://github.com/github/gh-ost/issues/730

### Description

Updates script/ensure-go-installed script for Raspberry Pi.  I was not sure about bumping up PREFERRED_GO_VERSION for other OS.


bancer:
@shlomi-noach Could this be merged?

shlomi-noach:
Thank you, I've been out of office. I'll catch up with this in a couple days.

shlomi-noach:
Two things that concern me:

- This is not tested
- If it were tested, it would currently fail, since `go1.11` uses different `go fmt` output than `go1.9`, something that is enforced by tests.

I'm still unsure how I feel about this. I appreciate that it would be nice to deliver non-linux build capabilities; but as soon as this is merged I would need to own those builds, which I will not be able to support.

bancer:
Understood. I tried to set up tests but faced the same problem as reported in [this](https://github.com/github/gh-ost/issues/425) issue. Are there any instructions or recommendations how to configure gh-ost-test-mysql-master and gh-ost-test-mysql-replica servers? MySQL-Sandbox seems to be [deprecated](https://github.com/datacharmer/mysql-sandbox#warning) so I think about using [DBdeployer](https://github.com/datacharmer/dbdeployer).

The lowest `*.linux-armv6l.tar.gz` package version available on https://storage.googleapis.com/golang/ is 1.10.1. Would that be better?

shlomi-noach:
Please see http://code.openark.org/blog/mysql/using-dbdeployer-in-ci-tests as for how we run (non-public facing) CI tests with `dbdeployer`. The repos and the code are public, but the tests themselves run internally at GitHub.

timvaillancourt:
@bancer I think this is safe to integrate. Could you update this PR branch?

Issue #737, opened 2019-05-17
drop old table twice at the same time


luoxiaocai91:
it drop old table twice at the same time,lead to slave delay.because the slave open MTS.
the slave processlist:
```
2019-05-17 00:03:01 | 5619159 | system user |   | payorder05 | Connect | 70 | Waiting for preceding transaction to commit | DROP TABLE IF EXISTS `_xx_del` /* generated by server */ | 70474 | 0 |  
-- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
2019-05-17 00:03:01 | 5619169 | system user |   |   | Connect | 78807 | Waiting for an event from Coordinator |   | 78806907 | 0 |  
2019-05-17 00:03:01 | 5619168 | system user |   |   | Connect | 44253 | Waiting for an event from Coordinator |   | 44253003 | 0 |  
2019-05-17 00:03:01 | 5619166 | system user |   |   | Connect | 17994 | Waiting for an event from Coordinator |   | 17993946 | 0 |  
2019-05-17 00:03:01 | 5619157 | system user |   | payorder05 | Connect | 70 | Waiting for table metadata lock | DROP TABLE IF EXISTS `_xx_del` /* generated by server */ | 70475 | 0
```

the command： 
```
gh-ost_v1.0.40 
--max-load=Threads_running=32             
--critical-load=Threads_running=48             
--chunk-size=1000             
--throttle-control-replicas=\"xxx:xx,xxx:xxx\"             
--max-lag-millis=1000             
--initially-drop-old-table             
--initially-drop-ghost-table             
--initially-drop-socket-file 
--ok-to-drop-table 
--conf=\"xxx\"                 
--host=\"xx\" 
--port=\"xx\" 
--database=\"xx\" 
--table=\"xx\"                 
--verbose 
--alter=\" xxx\"                 
--switch-to-rbr=false                 
--cut-over=default                 
--assume-rbr=true                 
--exact-rowcount=false                 
--default-retries=3                 
--critical-load-interval-millis=3000                 
--panic-flag-file=/tmp/ghost.panic.flag.xx.xx.66506955                 
--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag.xx.xx.66506955                 
--throttle-flag-file=/tmp/ghost.throttle.flag.xx.xx.66506955                 
--execute 
--allow-on-master
```
mysql version：5.7.21-21-log Percona Server (GPL), Release 21, Revision 2a37e4e
the log：

```
[...]

PR #745, opened 2019-05-25
avoid to send drop cutover sentry table to mysql twice v2

MOON-CLJ:
this pr is another way to deal with issue: #737 

the first two commits are the same as https://github.com/github/gh-ost/pull/743 ，instead use get_lock，this pr use tableUnlocked channel to sync。


MOON-CLJ:
@shlomi-noach yes, you're right。 

can we just delete the logic "defer this.DropAtomicCutOverSentryTableIfExists()"? because a successful cutover will drop the sentry table eventually。or do you have any other idea? 

ps, even this bug is even more a MTS's bug，i am thinking what can we do in gh-ost to eliminate the pain。

shlomi-noach:
OK, a less-pretty and yet simple solution is to protect the dropping of the table by a mutex, and always skip the 2nd operation. It is imperative that the table is dropped no matter what, which is why the `defer DropAtomicCutOverSentryTableIfExists()` is there.

MOON-CLJ:
@shlomi-noach "to protect the dropping of the table by a mute", yes, can i gave another pr for this?

shlomi-noach:
@MOON-CLJ sure, thank you!

Issue #740, opened 2019-05-17
Leveraging Docker for the build/test process?


zmoazeni:
This is a meatier issue, but I'm interested in simplifying some of the build/cibuild scripts by using [the golang-cross docker image](https://hub.docker.com/r/dockercore/golang-cross) instead of downloading/verifying the actual Go version for the host environment.

Some other areas of simplification I can see:

* Drop the `.gopath`. This would be part of the mapped docker volume instead.
* Possibly bundling the binary of dbdeployer inside the image which might affect `gh-ost-ci-env`. But a side effect of this could be possibly parallelizing the MySQL version localtests.

It'll take a bit of effort and experimentation, but I wanted to reach out to see if this was interesting before spending I began spending the effort. We should be [a-okay on the Travis-CI's side](https://docs.travis-ci.com/user/docker/).

My goal here would be to simplify bootstrapping and testing the codebase. Both with the script maintenance and the dev-groking.

shlomi-noach:
@zmoazeni I'm on board with this. You may have noticed I'm a bit slow to respond, but otherwise this should be good.

zmoazeni:
> You may have noticed I'm a bit slow to respond, but otherwise this should be good.

No worries. I just figured you're busy on other things at the moment. I'll put something together and see if you like it.

zmoazeni:
@shlomi-noach I have something working locally for build/fmt/testing. I wanted to ask though, is `./script/cibuild-gh-ost-build-deploy-tarball` used? I'm wondering if I need to convert that or if I can just drop it.

shlomi-noach:
@zmoazeni yeah, `./script/cibuild-gh-ost-build-deploy-tarball` is used internally at GitHub and is the build script used by our CI/CD to deploy `gh-ost`.

PR #738, opened 2019-05-17
Upgrade Go to v1.12.4

zmoazeni:
Related issue: https://github.com/github/gh-ost/issues/727

### Description

This PR upgrades gh-ost to v1.12.4

### Some notes ###

I ran `./script/cibuild-gh-ost-replica-tests` and ran into some flakey failures in the v5.5.x MySQL line. Primarily around `datetime-1970` and `mixed-charset`. I'm not seeing those tests in TravisCI. Do you run those on a private box?

I extended the script to allow us to pass a filter for the specific test and they intermittently seemed to be fine/fail individually (more often than not, they passed). So that has me scratching my head a little.

```diff
diff --git a/script/cibuild-gh-ost-replica-tests b/script/cibuild-gh-ost-replica-tests
index ab8c689..2d73f1c 100755
--- a/script/cibuild-gh-ost-replica-tests
+++ b/script/cibuild-gh-ost-replica-tests
@@ -22,6 +22,9 @@ test_mysql_version() {
   local mysql_version
   mysql_version="$1"
 
+  local test_filter
+  test_filter="$2"
+
   echo "##### Testing $mysql_version"
 
   echo "### Setting up sandbox for $mysql_version"
@@ -53,7 +56,7 @@ test_mysql_version() {
   gh-ost-test-mysql-master -uroot -e "grant all on *.* to 'gh-ost'@'%' identified by 'gh-ost'"
 
   echo "### Running gh-ost tests for $mysql_version"
-  ./localtests/test.sh -b bin/gh-ost
+  ./localtests/test.sh -b bin/gh-ost $test_filter
 
   find sandboxes -name "stop_all" | bash
 }
@@ -62,5 +65,5 @@ echo "Building..."
 . script/build
 # Test all versions:
 find gh-ost-ci-env/mysql-tarballs/ -name "*.tar.gz" | while read f ; do basename $f ".tar.gz" ; done | sort -r | while read mysql_version ; do
-  test_mysql_version "$mysql_version"
+  test_mysql_version "$mysql_version" "$1"
 done
``` 

_Let me know if you're in favor of that diff, and I'll issue a separate PR for it_

PR #750, opened 2019-05-31
WIP: Uses docker to build Go and packages

zmoazeni:
This is related to the conversation at https://github.com/github/gh-ost/issues/740

**Note: this is a draft PR to request feedback**

I still need to convert `script/cibuild-gh-ost-replica-tests` and `localtests/test.sh`.

Overall this works really well. I have made a few changes to the output directories, which I imagine you'll have opinions on. `./build.sh` still builds the binaries and deb/rpm/tarball packages.

I added a new option for `./script/cibuild` to optionally skip building everything. Though I do wonder if `build.sh` should be pushed into `./script/build` and `./script/cibuild` either be deleted (and folded into the build script). I'm not sure, something feels off organizationally with those behaviors.


.gitignore
```diff
@@ -2,3 +2,5 @@
 /bin/
 /libexec/
 /.vendor/
+/.cache
+/build
```
zmoazeni: I changed the output of the build artifacts. Now everything ends up in `./build`:

```bash
% find ./build
./build
./build/dist
./build/dist/gh-ost-binary-osx-20190531141809.tar.gz
./build/dist/gh-ost_1.0.48_amd64.deb
./build/dist/gh-ost-1.0.48-1.x86_64.rpm
./build/dist/gh-ost-binary-linux-20190531141809.tar.gz
./build/bin
./build/bin/gh-ost-linux-amd64
./build/bin/gh-ost-darwin-amd64
```

Totally get it if you don't like this. I thought I'd offer it up first though. That way we're not splitting the `./bin` and `/tmp/gh-ost-release` directories.

build.sh
```diff
@@ -1,73 +1,77 @@
...
+    -e GOCACHE=/go/src/${PACKAGE_PATH}/.cache \
+    -e RELEASE_VERSION=${RELEASE_VERSION} \
+    -e BIN_PATH=${buildpath}/bin"
 
-  if ! go version | egrep -q 'go(1[.]9|1[.]1[0-9])' ; then
-    echo "go version is too low. Must use 1.9 or above"
-    exit 1
+  if [ "$goos" == "darwin" ]; then
+    # specify cross compiler
+    envs="$envs -e CC=x86_64h-apple-darwin14-cc"
```
zmoazeni: This is the magic of `dockercore/golang-cross`. By specifying the `CC` that's already bundled, we can cross compile for darwin.

I'm 90% sure this is fine for the release. I copied this over to a mac machine I have and verified that I can run at least `--version`. (I might eat my hat once the localtests, but I'm pretty confident this will be okay)

build.sh
```diff
@@ -1,73 +1,77 @@
...
+    -v $(pwd):/go/src/${PACKAGE_PATH} \
+    gh-ost/fpm"
+
+  $fpm_cmd bash -c 'mkdir -p /gh-ost/usr/bin \
+      && cp ${BIN_PATH}/gh-ost-linux-amd64 /gh-ost/usr/bin/gh-ost \
+      && cd /gh-ost \
+      && fpm -v "${RELEASE_VERSION}" --epoch 1 -f -s dir -n gh-ost -m "shlomi-noach <shlomi-noach+gh-ost-deb@github.com>" --description "GitHub'"'"'s Online Schema Migrations for MySQL " --url "https://${PACKAGE_PATH}" --vendor "GitHub" --license "Apache 2.0" -C /gh-ost --prefix=/ -t rpm . \
+      && cp *.rpm ${OUTPUT_PATH}'
+
+  $fpm_cmd bash -c 'mkdir -p /gh-ost/usr/bin \
```
zmoazeni: You can test out the output of any of these in docker too:

```bash
% docker run --rm -it -v $(pwd):/ghost ubuntu dpkg -c /ghost/build/dist/gh-ost_1.0.48_amd64.deb
drwxrwxrwx 0/0               0 2019-05-31 18:18 ./
drwxr-xr-x 0/0               0 2019-05-31 18:18 ./usr/
drwxr-xr-x 0/0               0 2019-05-31 18:18 ./usr/share/
drwxr-xr-x 0/0               0 2019-05-31 18:18 ./usr/share/doc/
drwxr-xr-x 0/0               0 2019-05-31 18:18 ./usr/share/doc/gh-ost/
-rw-r--r-- 0/0             155 2019-05-31 18:18 ./usr/share/doc/gh-ost/changelog.gz
drwxr-xr-x 0/0               0 2019-05-31 18:18 ./usr/bin/
-rwxr-xr-x 0/0        10029736 2019-05-31 18:18 ./usr/bin/gh-ost

% docker run --rm -it -v $(pwd):/ghost fedora rpm -qlp /ghost/build/dist/gh-ost-1.0.48-1.x86_64.rpm
/usr/bin/gh-ost
```

build.sh
```diff
@@ -1,73 +1,77 @@
...
-  fi
-
-
-  buildpath=/tmp/gh-ost-release
-  target=gh-ost
-  timestamp=$(date "+%Y%m%d%H%M%S")
-  ldflags="-X main.AppVersion=${RELEASE_VERSION}"
+function distro_packages() {
+  echo "Creating Distro full packages"
+  docker build -t gh-ost/fpm -f docker/Dockerfile.fpm docker
```
zmoazeni: It's true that this will build the `fpm` image every time, but the docker layer cache will make this super fast on repeated runs.

script/cibuild
```diff
@@ -2,16 +2,16 @@
...
-gofmt -s -w  go/
-git diff --exit-code --quiet
+$DGO gofmt -s -w go
+git diff --exit-code --quiet go
 
-echo "Building"
-script/build
-
-cd .gopath/src/github.com/github/gh-ost
+if [ "$1" != "fast" ]; then
```
zmoazeni: If you just want to run gofmt and test, you can run this as `./script/cibuild fast`.

However, see my top-level comment. Something feels off that this is separate from `build.sh`. I'm thinking that this code should just be combined into one script at this point.

script/cibuild
```diff
@@ -2,16 +2,16 @@
...
 set -e
 
-. script/bootstrap
+. script/common
 
 echo "Verifying code is formatted via 'gofmt -s -w  go/'"
-gofmt -s -w  go/
-git diff --exit-code --quiet
+$DGO gofmt -s -w go
+git diff --exit-code --quiet go
```
zmoazeni: This bit me a few times. Sometimes I have non-go changes locally and this currently checks all stuff. Not just any go code changes.

script/common
```diff
@@ -0,0 +1,19 @@
...
+
+RELEASE_VERSION=
+if [ -z "${RELEASE_VERSION}" ] ; then
+  RELEASE_VERSION=$(git describe --abbrev=0 --tags | tr -d 'v')
+fi
+if [ -z "${RELEASE_VERSION}" ] ; then
+  RELEASE_VERSION=$(cat RELEASE_VERSION)
+fi
+
+DGO="docker run -u $(id -u):$(id -g) --rm \
```
zmoazeni: `-u $(id -u):$(id -g)` I do this so that the user/group permissions of the files generated stay the host user's. Otherwise it becomes `root:root` and a permissions pain.

script/go
```diff
@@ -2,10 +2,6 @@
...
 
-. script/bootstrap
+. script/common
 
-mkdir -p bin
-bindir="$PWD"/bin
-
-cd .gopath/src/github.com/github/gh-ost
-go "$@"
+$DGO go "$@"
```
zmoazeni: This still felt somewhat useful. Shouldn't be necessary if you have go installed locally, but :man_shrugging:

test.sh
```diff
@@ -1,13 +0,0 @@
-#!/bin/bash
```
zmoazeni: I'm not sure where this script runs. This is equivalent to `./script/go test ./go/...` (or `./script/go test -count 1 ./go/...` if you want to disable test caching)

build.sh
```diff
@@ -1,73 +1,77 @@
...
+timestamp=$(date "+%Y%m%d%H%M%S")
 
+function build() {
+  goos=$1
+  osshort=$goos
 
+  envs="-e GOOS=$goos \
+    -e GOARCH=amd64 \
+    -e CGO_ENABLED=1 \
+    -e GOCACHE=/go/src/${PACKAGE_PATH}/.cache \
```
zmoazeni: This is necessary since we don't run these commands as root inside the container (so we can't write to `/`).

Maintaining these files in the host directory does have a benefit of repeated builds being faster if you keep around the `./.cache` folder.

build.sh
```diff
@@ -1,73 +1,77 @@
...
+. ./script/common
+buildpath=build
+timestamp=$(date "+%Y%m%d%H%M%S")
 
+function build() {
+  goos=$1
+  osshort=$goos
 
+  envs="-e GOOS=$goos \
+    -e GOARCH=amd64 \
```
zmoazeni: Technically, nothing is stopping this from extending to also offering `386` or windows at this point. But I didn't go that far.

zmoazeni:
@shlomi-noach here's a first look at the updates to the build process. This is still WIP. My next steps are to convert `script/cibuild-gh-ost-replica-tests` and `localtests/test.sh` but it's at a spot to get some initial feedback.

build.sh
shlomi-noach: There's an open issue+PR for supporting Raspberry Pi. Truly.

build.sh
shlomi-noach: At some point I will want to get rid of `fpm`.

build.sh
zmoazeni: Hah. I haven't tried compiling for ARM but it doesn't look that difficult https://www.thepolyglotdeveloper.com/2017/04/cross-compiling-golang-applications-raspberry-pi/

I imagine ARM is going to get more popular with the Intel NUC crowd and the rumors MacBook Pro will ditch Intel and move to it.

zmoazeni:
@shlomi-noach Ok this is getting very close to a full PR. However in this process I started running into the intermittent tests that I mentioned in #738.

So I noodled on it a bit. At first I started with a similar tactic you already had of starting a single server for the version and running all tests against it.

Then after seeing the intermittent failures, I decided to start/kill a docker container for each test. It's slower but safer because we don't leave any remnants of a previous test on a fresh container boot. But even after that I'm still seeing errors crop up.

I thought perhaps this was a v1.12.x only issue, but even changing the GO_VERSION file back to 1.9.2. But I still saw errors.

I wonder if perhaps there's a default variable with the dbdeployer mysql servers that isn't lenient enough that we're tripping over. Or maybe something with the default values we're passing gh-ost. I'm not sure.

I haven't dug deeper yet, but here are two snippets from intermittent failures: https://gist.github.com/zmoazeni/4ea85d89fa07280f6fc5c41c0e2020c0

Note: they are not the same tests that regularly fail. Seems to happen all over place.

.gitignore
shlomi-noach: `build/` makes perfect sense! Let's add `/build/` to `.gitignore`

shlomi-noach:
@zmoazeni thank you for your work thus far! 

I'm not sure either about these intermittent replicaiton-not-running failures. I've seen these from time to time and I'm just not sure why they keep coming up.

OK, some news: at a strange timing coincidence, we (GitHub) are to now move all our internal builds to run on Docker. All our `*-build-deploy-tarball` builds mentioned earlier will need to be dockerized. 

This makes me both grateful for the work you've done, as well as uncomfortable that if you continue doing it, you'll be contributing your volunteer work on something we now need to do internally on our pay job. This is all new, so I don't have much insight into the work yet, but please consider this as heads up -- and for your consideration ❤️

.gitignore
zmoazeni: Will change! 

For my own education, how does git treat `/build` and `/build/` differently?

zmoazeni:
Ah! Thanks for letting me know. I really appreciate that. What I'll do then is back off this initiative until you get more information and let you push it forward. I'll leave this PR up so feel free to pick-n-choose what makes sense for that. And if you have any questions about why I did something a certain way, definitely feel free to ask.

I was also wondering if we should do this on orchestrator too, but GH's internal build changes will require that too.

I'll probably poke at these intermittent failures with this code, maybe I can find a reliable way to reproduce it on a single test. I feel like it's pretty close. My first intuition is that this isn't gh-ost code itself running into an issue.

With the "ERROR replication not running; Slave_IO_Running=No, Slave_SQL_Running=No", I might try turning on the general log and inspecting the MySQL container when the test fails to see what it ran up to that point to break replication.

PR #755, opened 2019-06-11
use sync.Once to avoid to send drop cutover sentry table to mysql twice

MOON-CLJ:
ref: https://github.com/github/gh-ost/pull/745#issuecomment-500678715

this pr is another way to deal with issue: #737

use mutex to skip the second drop table。

@shlomi-noach pls review, thx in advance。


MOON-CLJ:
@shlomi-noach 

i am feeling difficult to use https://golang.org/pkg/sync/#Once to ensure the following three places?

1, CreateAtomicCutOverSentryTable https://github.com/github/gh-ost/blob/49262c523a81b231d032aa3da6a83fec672eab64/go/logic/applier.go#L827

2, drop cut over table https://github.com/github/gh-ost/blob/49262c523a81b231d032aa3da6a83fec672eab64/go/logic/applier.go#L845

3, defer clean sentry table https://github.com/github/gh-ost/blob/49262c523a81b231d032aa3da6a83fec672eab64/go/logic/migrator.go#L613

even more, atomicCutOver will retry some times。although will sleep 1s between retry，two call to goroutine AtomicCutOverMagicLock is not serialize。

shlomi-noach:
@MOON-CLJ the way I see it, there's these two places:

https://github.com/github/gh-ost/blob/b38814f90ebfd429bd1bdf0ebd258472e5bb3965/go/logic/applier.go#L862-L865

https://github.com/github/gh-ost/blob/b38814f90ebfd429bd1bdf0ebd258472e5bb3965/go/logic/migrator.go#L613

which are racing. 
These do not retry per cut-over.

You are absolutely right that the entire cut-over operation can retry. We can create a `Once` variable per cut-over.

MOON-CLJ:
@shlomi-noach 

"These do not retry per cut-over." yeah, you're right。

sorry, my mistake。i am confused, because in pr https://github.com/github/gh-ost/pull/745 i mv defer clean sentry table https://github.com/github/gh-ost/blob/49262c523a81b231d032aa3da6a83fec672eab64/go/logic/migrator.go#L613 from atomicCutOver to goroutine AtomicCutOverMagicLock。 so there is a race condition between 

the last one cut-over

https://github.com/github/gh-ost/blob/1635183664f9c2a274f092a79aad548d5b49299d/go/logic/applier.go#L823

and the next one cut-over

https://github.com/github/gh-ost/blob/1635183664f9c2a274f092a79aad548d5b49299d/go/logic/applier.go#L819

> We can create a Once variable per cut-over.

yes, let me try。

MOON-CLJ:
@shlomi-noach updated

go/logic/applier.go
```diff
@@ -781,7 +782,7 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
 }
 
 // AtomicCutOverMagicLock
-func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocked chan<- error, okToUnlockTable <-chan bool, tableUnlocked chan<- error) error {
+func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocked chan<- error, okToUnlockTable <-chan bool, tableUnlocked chan<- error, dropCutOverSentryTableOnce sync.Once) error {
```
shlomi-noach: We must pass a pointer to `sync.Once` or else it is a copy of `sync.Once` which makes it a sync.Twice 😆

MOON-CLJ: okay, let me fix it.

MOON-CLJ: @shlomi-noach updated

MOON-CLJ:
@shlomi-noach we have test this pr in hundreds of clusters in our production env for more than 3 months。consider go forward  to merge?

MOON-CLJ:
@shlomi-noach ping

shlomi-noach:
Sorry for the delay. It takes a context switch to dive into this PR again; I'd need to clear my desk for this. I assume this will be delayed for a bit longer.

shlomi-noach (approved):
This look good to me.

timvaillancourt:
This was merged in https://github.com/github/gh-ost/pull/888. Thanks @MOON-CLJ!

shlomi-noach:
Thank you @timvaillancourt :crown:

MOON-CLJ:
Thank you @timvaillancourt

PR #759, opened 2019-06-22
Opt out of binlogs

### LINKED ISSUE #146: Disable binlogging when running with --test-on-replica?
What would you think about running `SET SESSION sql_log_bin=0` before applying changes when running in `--test-on-replica` mode? Right now, after testing on a replica, we need to wipe it and restore it from scratch (to prevent any invalid GTIDs from creeping in the event of a later promotion of that host to master). This is kind of irritating, and if we just turn off binlogging for that session, we wouldn't necessarily have to reset the box afterwards.

### LINKED ISSUE #149: support `--sql-log-bin` (bool)

### LINKED ISSUE #254: FR/RFC: optional flag for GTID cleanup after test/migrate on replica
As Shlomi mentioned in his recent blog post, GTID is problematic with `--test-on-replica` or `--migrate-on-replica` because the replica will no longer be promotable due to having extra gtid_executed transactions. As alluded to in the post, there are workarounds for this. On the GTID side of things, I can think of a couple different ways:

(a) Insert empty transactions with correct gtid_next on the master (and replicating down to all replicas), so that gtid_executed of all nodes now effectively matches the modified replica's
or
(b) Adjust the modified replica's gtid_executed via convoluted RESET MASTER song-and-dance

Both solutions require SUPER. Hopefully there are RDS equivalents but this needs more research.

If there's interest, I might be able to spend some time on a PR in a few months (just not in Oct, hoping to release another project before the Percona Live submission deadline :) Any guidance on which solution is more palatable? 

Option (a) is a bit cumbersome in the `--migrate-on-replica` case, since it may require a very large number of empty transactions to be inserted, and it involves interacting with the master. OTOH, this could be implemented as a general-purpose gtid_executed-syncing external tool which gh-ost just shells out to at the appropriate time.

Option (b) limits its impact to just the modified replica. The drawbacks are that all binlogs on the replica get deleted (probably ok since they're polluted with all those extra writes anyway); and the modified replica must not have replicas of its own. The automated procedure is something like this:

BEFORE MIGRATION:
* On the target replica, parse the chunk of gtid_executed corresponding to the replica's own server_uuid (if any) and store in a variable. This will let us figure out what to "restore" to in order to effectively undo the gtids assigned via gh-ost writes.

AFTER MIGRATION:
* Use PURGE BINARY LOGS TO '\<current log file\>' to delete all old binlogs. This is just done here first so that the later RESET MASTER step is fast, in cases where there are lots of large binlogs and filesystems where deletion is slow
* STOP SLAVE -- so that we can get a consistent position. The steps after are fast so replication won't need to be stopped for more than a split second.
* Query @@global.gtid_executed and store in a variable, and then modify it so that the chunk corresponding to the replica's own server_uuid matches its prior state. i.e., figure out what its gtid_executed "should" be if the gh-ost transactions had not occurred.
* RESET MASTER -- this will nuke gtid_executed and gtid_purged, along with the one remaining current binlog
* SET GLOBAL GTID_PURGED = '\<gtid_executed value calculated from prior step\>' -- this also inherently sets gtid_executed
* FLUSH LOCAL BINARY LOGS -- roll to a new binlog immediately, so that the gtid_executed value gets persisted in 5.6, which weirdly uses the binlogs as only persisted source of truth
* PURGE BINARY LOGS TO '\<current binlog file\>' -- erase the first binlog file in case binlog_gtid_simple_recovery is in use, since it would cause issues upon mysqld restart
* START SLAVE (unless leaving this as the user's responsibility, I think that's already the assumption in `--test-on-replica`)

### LINKED ISSUE #646: --test-on-replica should not write to binlogs
In our environment, we do not use GTID's and we run most clusters in an Active/Passive Master/Master configuration.  The current behavior of gh-ost's test on replica function is that it assumes the host is a leaf node with no replicas and still writes the changes to the binlog.

It would be more beneficial if there was an additional flag to not write to the binlog.  

The specific situation I am in is that I am compressing some large tables.  Based on small tests and estimates, we should have enough space to complete the compression of one table without running out of disk space.  I had intended on running the migration on the passive master, which is taking no traffic and could run out of disk space without any negative issues.  With the migration being written to the binlogs, though, this would also cause the active master to run out of space as well.

zmoazeni:
This PR allows users to opt out of writing to binary logs for --test-on-replica, --migrate-on-replica, and --allow-on-master

* closes #646
* closes #254
* closes #149
* closes #146

This is still WIP. I have a few things that I want to clarify or clean up:

* [ ] Should we allow this on `--allow-on-master`?
* [ ] Run shellcheck on the new bash scripts
* [ ] Fail early on v5.5
* [ ] Fail early if the server has replicas
* [ ] Should we offer a way to write empty GTID statements to a file?
* [X] Should we generate `gtid_next` for every statement? _Answer: No_


go/cmd/gh-ost/main.go
```diff
@@ -90,6 +90,7 @@ func main() {
 	cutOver := flag.String("cut-over", "atomic", "choose cut-over type (default|atomic, two-step)")
 	flag.BoolVar(&migrationContext.ForceNamedCutOverCommand, "force-named-cut-over", false, "When true, the 'unpostpone|cut-over' interactive command must name the migrated table")
 	flag.BoolVar(&migrationContext.ForceNamedPanicCommand, "force-named-panic", false, "When true, the 'panic' interactive command must name the migrated table")
+	flag.BoolVar(&migrationContext.UseBinLog, "sql-log-bin", true, "When true, gh-ost will write to the binlog for all operations that perform --test-on-replica, --migration-on-replica, and --allow-on-master")
```
zmoazeni: I think we should discuss whether we should allow `--allow-on-master` to work considering the scope of the GTID surgery. Feels super dangerous.

go/logic/applier.go
```diff
@@ -179,7 +180,7 @@ func (this *Applier) CreateGhostTable() error {
 		sql.EscapeName(this.migrationContext.DatabaseName),
 		sql.EscapeName(this.migrationContext.GetGhostTableName()),
 	)
-	if _, err := sqlutils.ExecNoPrepare(this.db, query); err != nil {
+	if _, err := sqlutils.ExecNoPrepare(this.db, this.migrationContext.DisableBinLogForQuery(query)); err != nil {
```
zmoazeni: When the user opts out, I wrap almost all queries in a `set sql_log_bin=off; statement; set sql_log_bin=on`

go/logic/applier.go
```diff
@@ -986,13 +990,23 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
 	var totalDelta int64
 
 	err := func() error {
+		if !this.migrationContext.UseBinLog {
+			if _, err := sqlutils.ExecNoPrepare(this.db, `SET sql_log_bin = OFF`); err != nil {
```
zmoazeni: This statement happens within a transaction so I need to selectively turn it off/on again outside the transaction.

go/logic/gtidset.go
```diff
@@ -0,0 +1,56 @@
+package logic
```
zmoazeni: With this being a new file, I didn't know if this needed the copyright statement at the top or not... dunno if that's necessary.

go/mysql/connection.go
```diff
@@ -116,5 +116,5 @@ func (this *ConnectionConfig) GetDBUri(databaseName string) string {
 	if this.tlsConfig != nil {
 		tlsOption = TLS_CONFIG_KEY
 	}
-	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?interpolateParams=%t&autocommit=true&charset=utf8mb4,utf8,latin1&tls=%s", this.User, this.Password, hostname, this.Key.Port, databaseName, interpolateParams, tlsOption)
+	return fmt.Sprintf("%s:%s@tcp(%s:%d)/%s?interpolateParams=%t&autocommit=true&charset=utf8mb4,utf8,latin1&multiStatements=true&tls=%s", this.User, this.Password, hostname, this.Key.Port, databaseName, interpolateParams, tlsOption)
```
zmoazeni: `multiStatements=true` is necessary to allow multiple statements within the query with our driver. Otherwise the query errors.

localtests/test.sh
```diff
@@ -134,6 +134,16 @@ test_single() {
   echo_dot
   sleep 1
   #
+
+  local callback_return
+  if [ -f $tests_path/$test_name/before.sh ] ; then
```
zmoazeni: I added a new way for tests to have before.sh/after.sh to expand on our tests. Each of the new tests depend on this.

go/logic/gtidset.go
shlomi-noach: If it's a new file, you get to pick the copyright statement, as long as it is compliant with this project's license. Best advice: MIT under your name.

shlomi-noach:
@zmoazeni thank you for working on this. I'm ambivalent and am more inclined to run a `RESET MASTER...` operation at end of migration so as to cleanup errant GTIDs, as opposed to avoiding writes in binlog in the first place.

There is at least one crucial logic whose path goes through the binary logs: the [waitForEventsUpToLock](https://github.com/github/gh-ost/blob/b38814f90ebfd429bd1bdf0ebd258472e5bb3965/go/logic/migrator.go#L538) logic. I don't see this PR handling the `waitForEventsUpToLock` logic, and to handle it outside the binary log means changing the logic that identifies the cut-over timing.

When `gh-ost` identifies that all row copy is complete, it wants to cut-over. However, there can also be a few more transactions in the binary log, so it can't cut-over just yet. The logic is:
- lock the original table for writes. 
- There can still be binlog entries for that table, of course, that must be handled. But no _new_ binlog entries are possible.
- `gh-ost` [injects](https://github.com/github/gh-ost/blob/b38814f90ebfd429bd1bdf0ebd258472e5bb3965/go/logic/migrator.go#L546) a "hey everything is processed" challenge 
- the migrator meanwhile keeps on applying binlog events
- but when it [sees](https://github.com/github/gh-ost/blob/b38814f90ebfd429bd1bdf0ebd258472e5bb3965/go/logic/migrator.go#L225-L238) the `AllEventsUpToLockProcessed` challenge in the binary log, it knows beyond doubt that all table events are processed (remember `AllEventsUpToLockProcessed` is injected after the table is locked).
- (on timeout, though, a new challenge is required)
- if all is well, proceed to cut-over.

On a `test-on-replica`, I guess this can be worked around, but... it will change the logic for the cut-over. And that would make the testing less reliable. `test-on-replica` really works through the entire logic...

go/logic/gtidset.go
```diff
@@ -0,0 +1,56 @@
...
+
+import (
+	"strings"
+)
+
+type GTIDSet struct {
```
zmoazeni: Ooh TIL about `gtid_subtract` https://github.com/github/orchestrator/blob/19c0209e6790912909ff6849600251c0e2a439ab/go/inst/instance_topology_dao.go#L1164 (per @shlomi-noach's link)

Perhaps I don't need this logic after all.

go/logic/applier.go
```diff
@@ -1043,3 +1062,73 @@ func (this *Applier) Teardown() {
...
+	})
+
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (this *Applier) PurgeGTIDs() error {
+	if _, err := sqlutils.ExecNoPrepare(this.singletonDB, `PURGE BINARY LOGS BEFORE NOW()`); err != nil {
```
zmoazeni: Note I don't feel strongly about this `PURGE BINARY LOGS BEFORE NOW()`

This may change over to `PURGE BINARY LOGS TO ...`, I just didn't understand @evanelias's warning on https://github.com/github/gh-ost/issues/254#issuecomment-502524514

However per @shlomi-noach's comment on https://github.com/github/gh-ost/issues/254#issuecomment-504996069 I'm not sure if we even need to run any purge. Going to wait until the discussion ends first before changing anything.

zmoazeni:
> I'm ambivalent and am more inclined to run a RESET MASTER... operation at end of migration so as to cleanup errant GTIDs, as opposed to avoiding writes in binlog in the first place.

Ah right on. I can drop the `sql_log_bin=OFF` business. At this point I think it's just space saver.

> There is at least one crucial logic whose path goes through the binary logs: the waitForEventsUpToLock logic. I don't see this PR handling the waitForEventsUpToLock logic, and to handle it outside the binary log means changing the logic that identifies the cut-over timing.

This all might be moot if I drop the `sql_log_bin=OFF` wrapping, but I believe I'm explicitly allowing those to write to the binary log via https://github.com/github/gh-ost/pull/759/files#diff-55e7b291008bc31a7fac46abdad84feaR293 - the tests would hang/wait at the cutover phase without that line.

Unless I'm missing something important here. Are you saying that by excluding binary log writing of the other statements perhaps interferes with the locking logic and assumptions?

go/logic/gtidset.go
zmoazeni: Yeah this was going to stay MIT. However after looking at your orchestrator code, I wonder if I should just use `gtid_subtract` instead.

shlomi-noach:
> Unless I'm missing something important here. Are you saying that by excluding binary log writing of the other statements perhaps interferes with the locking logic and assumptions?

What I was saying:
- You can get rid of _many_ binlog entries
_ But you can't get rid of all of them. There's a few which are required for the correct operation of `gh-ost`.
- So you end up with errant GTID anyway.
- And so you still need to e.g. `RESET MASTER`
- In which case why go through the effort of all the `sql_log_bin=OFF` stuff.

The above is my preferred view/opinion. It could be possible to get rid of _all_ binlog entries, but at the cost of changing the cut-over logic; something I'm not keen on.

zmoazeni:
> In which case why go through the effort of all the sql_log_bin=OFF stuff.

Ahh I see your point. I was doing that to help #149:

> The specific situation I am in is that I am compressing some large tables. Based on small tests and estimates, we should have enough space to complete the compression of one table without running out of disk space. I had intended on running the migration on the passive master, which is taking no traffic and could run out of disk space without any negative issues. With the migration being written to the binlogs, though, this would also cause the active master to run out of space as well.

I read that as: they have enough space on the server to duplicate the table but when we also account for the extra space in the binary log, things go haywire for that person.

This PR would allow all the changelog events to still write to the binary log (which I believe appropriately handles `waitForEventsUpToLock`), but I need to do a bit more reading on it to make sure I understand it.

> The above is my preferred view/opinion. It could be possible to get rid of all binlog entries, but at the cost of changing the cut-over logic; something I'm not keen on.

I totally agree with you here. I'm not wanting to change over the cut-over logic. I'll spend some more time reading up on `waitForEventsUpToLock` to see if I screwed up something here. At the very least I could pare this PR down to not do any `sql_log_bin=OFF` and could re-introduce that in a separate PR, treating that as a separate concern (optimizing disk space).

go/logic/applier.go
evanelias: My concern is there may be some cases where `PURGE BINARY LOGS BEFORE NOW()`, when run after the GTID surgery, might not actually purge the first and only binlog. `NOW()` does not have sub-second precision. So if that binlog was created in the current clock second, does it actually get purged? In other words, is `BEFORE` like `<` or is it like `<=`?

If that is indeed a problem, it may be avoidable with something goofy like `NOW() + INTERVAL 1 DAY`. But using `SHOW MASTER STATUS` to get the current binlog feels more precise, in my opinion.

go/base/context.go
```diff
@@ -736,3 +738,12 @@ func (this *MigrationContext) ReadConfigFile() error {
...
 	return nil
 }
+
+// DisableBinLogForQuery if UseBinLog is false will wrap the query by wrapping
+// it in SET sql_log_bin = OFF/ON
+func (this *MigrationContext) DisableBinLogForQuery(query string) string {
+	if this.UseBinLog {
+		return query
+	}
+	return fmt.Sprintf("SET sql_log_bin = OFF; %s; SET sql_log_bin = ON;", query)
```
evanelias: I realize this code may be going away, but just a general word of caution with this multiStatements=true trick: wrapping the query will break the ability of DML queries to return results about # of rows affected or last-insert-id. I'm also not sure what happens if the query errors -- you may need to be very careful to avoid having some random connection returned to your connection pool with sql_log_bin=0 still in place...

The idiomatic way to set session variables in database/sql is via [DB.Conn()](https://golang.org/pkg/database/sql/#DB.Conn), which allows you to operate on a single session across multiple queries.

go/logic/applier.go
evanelias: This won't do what you expect... `this.db` is a `sql.DB` here right? If so, it represents a connection pool, not a single connection. There's no guarantee that the connection used for `SET sql_log_bin = OFF` will be the same connection used for `tx, err := this.db.Begin()`... you'd need to use [this.db.Conn()](https://golang.org/pkg/database/sql/#DB.Conn) here too I believe.

zmoazeni: Thanks @evanelias!

go/logic/applier.go
zmoazeni: > NOW() does not have sub-second precision

Ahh thanks! Will change.

go/base/context.go
zmoazeni: > The idiomatic way to set session variables in database/sql is via DB.Conn(), which allows you to operate on a single session across multiple queries.

Yeah the issue was that we didn't want all statements to have `sql_log_bin = OFF`. This was a way for me to opt-in on the statement level.

However, I wonder if it would have been better handled like the `singletonDB` to direct statements with/without binary logging to a particular connection (or connection pool).

Either way no biggy. @shlomi-noach suggested to rip out this part of the PR for now and I'm good with that.

zmoazeni:
Thanks for the feedback gang. I've been taking some vacation time lately. Will jump back in soon-ish to make changes based on the feedback! I believe the next iteration will be a legit non-draft PR.

zmoazeni:
Closing unmerged. I have since moved my focus elsewhere.

Issue #763, opened 2019-07-15
Accidental deletion of _ghc or _gho should cause gh-ost fatally exit.


tomkrouper:
We had a situation today were we accidentally renamed the _ghc and _gho tables, however, gh-ost continued to run with just an error saying that the tables didn't exist. The removal, rename, or otherwise loss of the ghost tables, I would think, should cause the process to die a most horrible death. I looked around, and since I don't know `go`, I wasn't able to provide a PR. Please take a look. Thanks.

```
2019-07-15 13:42:50 ERROR Error 1146: Table 'a_database._a_table_ghc' doesn't exist                                                                                     2019-07-15 13:42:50 ERROR Error 1146: Table 'a_database._a_table_gho' doesn't exist; query=
... snip ...
```

ggunson:
Other points:

- `sup` reported the migration as still running a minute or so after the rename occurred

```
Copy: 26033500/51222779 50.8%; Applied: 60947; Backlog: 391/1000; Time: 2h23m50s(total), 2h23m47s(copy); streamer: mysql-bin.081589:124308099; State: migrating; ETA: 2h19m7s
```

- the error message only appeared in the error log when the migration was killed via `panic`

shlomi-noach:
We run with `--default-retries=120`, meaning for example a `REPLACE INTO` query that attempts to apply a binlog event would retry 120 times at 1sec interval, and would eventually cause `gh-ost` to panic, roughly after `120sec` of failed attempts. That's for whatever reason for failure.

The cut-over itself will also be susceptible to the missing table problem, and would also retry 120 times (120 * 1sec interval) until panics. So even if there's no more rows to copy, the cut-over would not be able to take place.

From our internal chat it seems like the accidental rename took place at `13:41:XX` and that upon identifying the mistake we issued a manual `.migration panic` at `13:43:54` -- but then also at that exact second `gh-ost` seemed to also log a `2019-07-15 13:42:54 FATAL Error 1146: Table XXX doesn't exist;...`. Please correct me if I'm wrong here.

I think the "natural" panic and the human invoked panic were coincidentally at the exact same second. A human-generated `panic` would result with a different error message; the "natural" panic seemed to have preceded the human invoked one by a fraction of a second. I believe `gh-ost` was operating correctly; I'm drafting a couple test scenarios (PR to follow up) to confirm.

shlomi-noach:
[...]

PR #772, opened 2019-08-01
Panic on missing _ghc and _gho tables

shlomi-noach:
Fixes https://github.com/github/gh-ost/issues/763

In case of accidental removal of `_gho` table while migration was running, `gh-ost` was correctly bailing out on failing to apply rows to the `_gho` table, but did not panic during cut-over phase.

With this PR:

- `gh-ost` correctly panics at cut-over phase
- Also, `gh-ost` now correctly panics when the `_ghc` table is accidentally removed throughout the migration.

cc @tomkrouper @ggunson

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -736,3 +736,20 @@ func (this *MigrationContext) ReadConfigFile() error {
 
 	return nil
 }
+
+func (this *MigrationContext) PanicAbortIfTableError(err error) {
+	if err == nil {
+		return
+	}
+	if strings.Contains(err.Error(), mysql.Error1146TableDoesntExist) || strings.Contains(err.Error(), mysql.Error1017CantFindFile) {
+		this.PanicAbortOnError(err)
+	}
+	// otherwise irrelevant error and we do not panic
+}
+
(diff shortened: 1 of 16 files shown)


shlomi-noach:
This seems too strict. We have a known race condition after cut-over where the `gho` table can be renamed; this PR then fails the migration even though everything is fine.

timvaillancourt (approved):

shlomi-noach:
As per https://github.com/github/gh-ost/pull/772#issuecomment-518127278 I recommend holding off this PR.

timvaillancourt:
> This seems too strict. We have a known race condition after cut-over where the `gho` table can be renamed; this PR then fails the migration even though everything is fine.

@shlomi-noach would this PR be safer if we had a mutex at the cut-over, eg: https://github.com/github/gh-ost/pull/755?

shlomi-noach:
> would this PR be safer if we had a mutex at the cut-over

@timvaillancourt I guess maybe? Depends on the mutex; there's already much locking/scheduling in `gh-ost` so we'd need to be careful so as not to introduce a deadlock or any other mistake.

PR #774, opened 2019-08-05
fix returned error values in defer functions

zmoazeni:
This is copied from https://github.com/github/orchestrator/pull/946 with
a slight tweak to the an indent.

Originates from the PR https://github.com/github/orchestrator/issues/943
that applies to this codebase too.

When we define the variable and attempt to set it in a defer function,
the value we're returning isn't what we expect.

Instead we can name the return values as variables and set that instead.

Thanks @yangeagle for the original report.

### DIFF
--- a/vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
+++ b/vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
@@ -21,11 +21,12 @@ import (
 	"encoding/json"
 	"errors"
 	"fmt"
-	_ "github.com/go-sql-driver/mysql"
-	"github.com/outbrain/golib/log"
 	"strconv"
 	"strings"
 	"sync"
+
+	_ "github.com/go-sql-driver/mysql"
+	"github.com/outbrain/golib/log"
 )
 
 // RowMap represents one row in a result set. Its objective is to allow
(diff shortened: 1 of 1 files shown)


timvaillancourt (approved):

vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
```diff
@@ -195,34 +196,33 @@ func ScanRowsToMaps(rows *sql.Rows, on_row func(RowMap) error) error {
...
-	rows, err := db.Query(query, args...)
+	var rows *sql.Rows
+	rows, err = db.Query(query, args...)
 	defer rows.Close()
 	if err != nil && err != sql.ErrNoRows {
 		return log.Errore(err)
 	}
 	err = ScanRowsToMaps(rows, on_row)
-	return err
+	return
```
timvaillancourt, line 213: @zmoazeni I believe this func should `return err` or `return nil`. Can you confirm?

timvaillancourt (changes requested):

vendor/github.com/outbrain/golib/sqlutils/sqlutils.go
zmoazeni: @timvaillancourt That seems to be the case.

zmoazeni:
I'm going to go ahead and close this. But feel free to revive this if it's important.

PR #776, opened 2019-08-11
Initial GitHub CI/CD

shlomi-noach:
GitHub CI/CD is [a thing](https://github.com/features/actions)! We will attempt:

- CI: build + unit test
- replica tests: running `gh-ost` tests on multiple MySQL versions

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -20,9 +20,9 @@ function build {
 
 
 
-  if ! go version | egrep -q 'go(1[.]9|1[.]1[0-9])' ; then
-    echo "go version is too low. Must use 1.9 or above"
-    exit 1
+  if ! go version | egrep -q 'go(1\.1[234])' ; then
+    echo "go version must be 1.12 or above"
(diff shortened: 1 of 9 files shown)


shlomi-noach:
That was quick. All seems to be running well.

Along the way, we've upgraded to `go1.12` which from now on is required.

build.sh
```diff
@@ -20,9 +20,9 @@ function build {
 
 
 
-  if ! go version | egrep -q 'go(1[.]9|1[.]1[0-9])' ; then
-    echo "go version is too low. Must use 1.9 or above"
-    exit 1
+  if ! go version | egrep -q 'go(1\.1[234])' ; then
+    echo "go version must be 1.12 or above"
+    ok=1
```
zmoazeni: Is this used anywhere? I didn't see it in the scripts.

shlomi-noach: nice catch! fixed.

Issue #224, opened 2016-09-02
Feedback requested: gh-ost on Galera / XtraDB cluster?


shlomi-noach:
Dear community,
At this time we are not using Galera/XtraDB Cluster. We haven't tested `gh-ost` on Galera. We're curious to hear:

- If anyone has
- If there are specific blockers or problems
- If `gh-ost` on Galera makes sense, given you can use Rolling Schema Upgrades.

Thank you!

Roguelazer:
That's actually on my list to test next week. It seems like the easiest way to do it would be to hang an asynchronous replica off of the cluster (which is a supported operation) and run gh-ost there to minimize load on the synchronous cluster.

`gh-ost` definitely makes sense; RSU is a huge pain in the butt since you are de-syncing the node (and need to stay in gcache to prevent undoing the migration and forcing a SST).

I'll let you know how the testing in our dev cluster works.

SuperQ:
We use Galera/XtraDB and would definitely like to use `gh-ost`.  We also have async slaves replicating from Galera "master cluster"s to provide additional read capacity and analytics access.  Therefore it does make sense to continue to use an online schema tool.

shlomi-noach:
Thank both! Do you foresee a particular issue with the final `lock/rename` steps?

SuperQ:
This applies to all writes that `gh-ost` does, not just the final step.

One thing that happens with Galera/XtraDB is the potential to deadlock "frequently".  In theory this can happen between threads on a normal mysql master, but it's very rare.

This is caused by the rollup of multiple transactions into a single locking operation between the multiple galera masters.  The default for this is 16 transactions, but is tunable[0].  I personally don't recommend changing this, but some people do.

`gh-ost` needs to be able to receive these deadlock rejected queries and retry them.  This is safe, but slightly tedious.  When dealing with distributed multi-master systems, this kind of problem will become more common. :)

[0]: https://www.percona.com/doc/percona-xtradb-cluster/5.6/wsrep-provider-index.html#gcs.fc_limit

shlomi-noach:
@SuperQ `gh-ost` already has retries, which are also configurable via `-default-retries`

SuperQ:
:+1: :100: :8ball:

Frank-Leemkuil:
@shlomi-noach We test gh-ost on a Galera staging cluster and get a lot of deadlocks it looks like it never stops:

[...]

PR #780, opened 2019-09-10
Allow use gh-ost on Galera / XtraDB cluster

Gonlo2:
Related issue: https://github.com/github/gh-ost/issues/224

### Description

Currently _gh-ost_ can't work on Galera-based MySQL like PXC because it needs to perform `LOCK TABLES` for both `atomic` and `two-step` cut-over phase.
A know workaround is to use `two-step` cut-over strategy but only with two not so common requirements:
1) All the write queries must be sent to one single node in the cluster
2) It's possible to have missing table for some time during the RENAME steps.
In order to use _gh-ost_ without the above requirements, on all Galera-based systems I created a new cut-over strategy named `trigger`
Using this option, during the _cut-over_ phase, _gh-ost_ will behave exactly like the well known tool _pt-online-schema-change_: adding triggers on the original table to keep the ghost one synchronised.

### How it works
In order to have an atomic table swap without the use of a `LOCK TABLES`, three triggers are created during the cut-over phase, like what _pt-online-schema-change_ does, and a special event is injected in the _binlog_ to handle the write delegation phase from _gh-ost_ to MySQL with triggers.
_gh-ost_ needs to handle that write delegation phase carefully because adding the triggers and stopping applying writes from _gh-ost_ can't be done in the exact same moment or atomically.
After this delegation phase is necessary to sanitise all the write events occurred from the start to the end of that phase.
For manage safely this scenario the `trigger` cut-over strategy does this steps:

- First _gh-ost_ inject a stop writes event in the _binlog_ and _gh-ost_ disable the writes once it receive it.
- Triggers are created to handle the modifications on the current table to the ghost one.
- A created triggers event is injected in the _binlog_ to notify _gh-ost_ that triggers are in place.
- The affected rows between the write delegation swap can be in an inconsistent state, so this events are sanitised by removing and adding again the values from the original table to the ghost one, that will be in sync.
- After this, and in the case that some error take place, the triggers are removed.
- Finally the tables are swapped.

### Triggers

Even though _gh-ost_ philosophy is avoid the use of triggers in Galera-based MySQL using this new cut-over _gh-ost_ is able to archive a zero-error table swapping while supporting Galera-based MySQL with writes on multiple nodes, like in a common setup.
Given the cut-over phase should take very short time, it shouldn't be a problem to add triggers only in this phase.


Gonlo2:
@shlomi-noach I know you don't have a lot of free time lately but, could you at least look into the look at the approach and the possibility of adding a trigger-based cut-over? We are using it in production for a couple of months and it is showing some good results.

If you have any questions, don't hesitate to let me know.

Gonlo2:
@jfudally I don't know if you can tell me how feasible it would be to integrate this new cutover for galera / XtraDB

In case it helps, I created a post some time ago explaining this particular case https://gonlo2.github.io/blog/en/blog/2020-03-04-why-dont-you-see-ghosts-in-galeras/

jfudally:
Unfortunately, I don't know enough about Galera to make a decision on this PR.

/cc @timvaillancourt @gtowey maybe?

lastthyme (approved):

PR #796, opened 2019-10-31
Mysql8 rename table

raidivyansh074:
Adds traversing organizations you belong to for `git clone repo

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -518,19 +518,23 @@ func (this *Migrator) cutOver() (err error) {
 			}
 		}
 	}
-	if this.migrationContext.CutOverType == base.CutOverAtomic {
+	var cutOverFunc func() error
+	if !mysql.IsSmallerMajorVersion(this.migrationContext.ApplierMySQLVersion, "8.0.13") {
+		// This is MySQL 8.0 or above. We can utilize a new ALTER TABLE featiure that supports
+		// RENAME while the table is locked.
+		cutOverFunc = this.cutOverMySQL8013
+	} else if this.migrationContext.CutOverType == base.CutOverAtomic {
 		// Atomic solution: we use low timeout and multiple attempts. But for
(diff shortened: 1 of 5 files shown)


go/logic/migrator.go
```diff
@@ -518,19 +518,23 @@ func (this *Migrator) cutOver() (err error) {
 			}
 		}
 	}
-	if this.migrationContext.CutOverType == base.CutOverAtomic {
+	var cutOverFunc func() error
+	if !mysql.IsSmallerMinorVersion(this.migrationContext.ApplierMySQLVersion, "8.0.13") {
```
timvaillancourt, line 522: @raidivyansh074 let's allow the user to decide if they want this style of cut-over, at least for now

Could you please add a new cut-over type the command line [here](https://github.com/github/gh-ost/blob/master/go/cmd/gh-ost/main.go#L98) and [here](https://github.com/github/gh-ost/blob/master/go/cmd/gh-ost/main.go#L263-L270)

The check for `mysql.IsSmallerMinorVersion` could move to be a check at the start of the new cut-over type only

go/mysql/utils.go
```diff
@@ -203,3 +204,47 @@ func GetTableColumns(db *gosql.DB, databaseName, tableName string) (*sql.ColumnL
 	}
 	return sql.NewColumnList(columnNames), sql.NewColumnList(virtualColumnNames), nil
 }
+
+func versionTokens(version string, digits int) []int {
```
timvaillancourt, line 208: @raidivyansh074 I suggest we use https://github.com/hashicorp/go-version for this instead so there is less code to maintain into the future

timvaillancourt (changes requested):
@raidivyansh074 thanks for the PR 🙇. There are some inline changes requested

Also, could you please open an issue and update the summary with this + a description of what is being solved here?

Issue #832, opened 2020-03-25
Gh-ost continues if HTTP throttler in error/offline


jfudally:
### Problem
During a migration if Freno or some other HTTP throttling services returns an error, or is unresponsive, gh-ost will continue to write unthrottled despite target load.

##### Using an invalid throttler endpoint
```
bin/gh-ost --database test --table messages --alter "ADD reciever varchar(25) AFTER sender" --allow-on-master --user ... --password=... --host localhost --throttle-http http://some.freno.tld --initially-drop-old-table --initially-drop-ghost-table --execute
[2020/03/25 14:16:04] [info] binlogsyncer.go:133 create BinlogSyncer with config {99999 mysql localhost 3306 root    false false <nil> false UTC true 0 0s 0s 0 false}
[2020/03/25 14:16:04] [info] binlogsyncer.go:354 begin to sync binlog from position (mysql-bin.000003, 164224)
[2020/03/25 14:16:04] [info] binlogsyncer.go:203 register slave for master server localhost:3306
[2020/03/25 14:16:04] [info] binlogsyncer.go:723 rotate to (mysql-bin.000003, 164224)
# Migrating `test`.`messages`; Ghost table is `test`.`_messages_gho`
# Migrating e398ffae37a9:3306; inspecting e398ffae37a9:3306; executing on MacBook-Pro-10
# Migration started at Wed Mar 25 14:16:04 -0500 2020
# chunk-size: 1000; max-lag-millis: 1500ms; dml-batch-size: 10; max-load: ; critical-load: ; nice-ratio: 0.000000
# throttle-additional-flag-file: /tmp/gh-ost.throttle
# Serving on unix socket: /tmp/gh-ost.test.messages.sock
Copy: 0/3 0.0%; Applied: 0; Backlog: 0/1000; Time: 1s(total), 0s(copy); streamer: mysql-bin.000003:170998; Lag: 0.02s, State: migrating; ETA: N/A
Copy: 0/3 0.0%; Applied: 0; Backlog: 0/1000; Time: 2s(total), 1s(copy); streamer: mysql-bin.000003:175364; Lag: 0.01s, State: migrating; ETA: N/A
Copy: 3/3 100.0%; Applied: 0; Backlog: 0/1000; Time: 2s(total), 1s(copy); streamer: mysql-bin.000003:176125; Lag: 0.01s, State: migrating; ETA: due
[...]

PR #833, opened 2020-03-25
Throttle on HTTP throttle error

jfudally:
Related issue: #832 

### Description

This PR changes the default throttling behavior when using an HTTP throttler (such as https://github.com/github/freno) that is unresponsive, or returns unexpected/missing data.  This change throttles gh-ost during these failure scenarios.  Because this changes the existing behavior, a `--ignore-http-errors` flag was added to allow the migration to continue if the HTTP throttler goes offline.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@
 /bin/
 /libexec/
 /.vendor/
+.idea/
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -119,7 +119,7 @@ type MigrationContext struct {
 	ThrottleAdditionalFlagFile          string
 	throttleQuery                       string
 	throttleHTTP                        string
-	ignoreHTTPErrors					bool
+	ignoreHTTPErrors                    bool
 	ThrottleCommandedByUser             int64
 	HibernateUntil                      int64
 	maxLoad                             LoadMap
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -119,6 +119,7 @@ type MigrationContext struct {
 	ThrottleAdditionalFlagFile          string
 	throttleQuery                       string
 	throttleHTTP                        string
+	ignoreHTTPErrors					bool
 	ThrottleCommandedByUser             int64
 	HibernateUntil                      int64
 	maxLoad                             LoadMap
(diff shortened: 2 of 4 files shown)


timvaillancourt:
@jfudally, I think this is pretty close 👍. I have 2 x suggestions:

`collectFunc` is called in a loop a 2nd-time in the `collectThrottleHTTPStatus` function: https://github.com/github/gh-ost/blob/df60fa4204bc7e84a02b5a9b82a19938353149e9/go/logic/throttler.go#L309-L311
So we need to do the same error catching in the looped version too (or in `collectFunc` itself)

Secondly, I think we should add an error message for the `-1` HTTP code in these blocks here: https://github.com/github/gh-ost/blob/df60fa4204bc7e84a02b5a9b82a19938353149e9/go/logic/throttler.go#L21-L37. Without that the user will get a less-useful error message from the fallback here: https://github.com/github/gh-ost/blob/df60fa4204bc7e84a02b5a9b82a19938353149e9/go/logic/throttler.go#L67

jfudally:
@timvaillancourt Thanks for the review!  I took care of these points you brought up.  I did have some error logging, but it was pretty spammy once I caught the error out of the second call to `collectFunc()`, so I added an appropriate error message to the `httpStatusMessages` structs, which makes for a much cleaner message:
```
Copy: 0/0 100.0%; Applied: 0; Backlog: 0/1000; Time: 2s(total), 0s(copy); streamer: mysql-bin.000004:134210; Lag: 0.01s, State: throttled, Connection error (http=-1); ETA: due
Copy: 0/0 100.0%; Applied: 0; Backlog: 0/1000; Time: 3s(total), 0s(copy); streamer: mysql-bin.000004:138599; Lag: 0.01s, State: throttled, Connection error (http=-1); ETA: due
Copy: 0/0 100.0%; Applied: 0; Backlog: 0/1000; Time: 4s(total), 0s(copy); streamer: mysql-bin.000004:142993; Lag: 0.02s, State: throttled, Connection error (http=-1); ETA: due

```

timvaillancourt:
@jfudally thanks 🎉, the latest changes LGTM!

PR #834, opened 2020-03-27
Add CodeQL Analysis

jhutchings1:
This PR adds a new Actions workflow which performs CodeQL analysis. This enables the code scanning staff ship, which will show you important security bugs on PRs and in the Security tab. We're shipping as a beta at Satellite, so we'd love to get your help to test the functionality. Please let me know if you have any questions/concerns!


jhutchings1:
This seems to be blocked in a bug in the Action. No need to review just yet.

PR #837, opened 2020-04-10
Hidden password from /proc/{pid}/cmdline or ps command output

martianzhang:
### Description

If the user sets the password with a command-line argument, when gh-ost running, another user can get the password from /proc/{pid}/cmdline file or from `ps` command. It seems to be not secure. As `mysql` client does, it will be more secure to replace the password with several x letters.


kian-boom:
bump

Issue #600, opened 2018-05-28
Fix copy % count inaccuracy by changing the heuristics 


ggunson:
gh-ost's statistics, when run on a large, busy table with a lot of inserts, can become inaccurate over time. Internally we've sometimes seen cutovers become available when the migration is reported as 90% complete or less. 

```
Copy: 1045877600/1264045717 82.7%; Applied: 884999578; Backlog: 12/1000; Time: 460h0m0s(total), 459h59m53s(copy); streamer: mysql-bin.012574:687201905; State: migrating; ETA: 95h57m17s
```

The `1045877600/1264045717 82.7%` above is equivalent to `count of rows in _gho table/count of rows in original table`. Compared to the actual counts, the size of the new _gho table was 11% greater, so the % completed was 93%. (The count of rows in the original table was quite accurate, if not exact).

Currently gh-ost determines the table counts by

1. Getting the row count at the beginning (whether the exact row count or an estimate),
1. Parsing the binlogs: +1 for inserts and -1 for deletes.
1. Getting the rows_affected from the `insert into _gho select ... from original_table`.

The counts are likely inaccurate because due to the concurrent threads of row copying and binlog parsing/applying, we don't know if a binlog INSERT results in a net 1 row increase (since it becomes a REPLACE, and might cancel out via delete+insert) and we don't know if a DELETE deletes a row (because it hadn't been copied over yet by the copy thread). Likely since gh-ost is erring on the side of too-low counts, @shlomi-noach guesses it's the parsed delete counts that are the bigger cause of the discrepancy.

If we have gh-ost get all the _gho row counts from the rows_affected of the writes to the _gho table, this should result in more accurate statistics. E.g. https://github.com/github/gh-ost/blob/6524a15ed343e6780c0d4a2e75042e6cb4121aa3/go/logic/applier.go#L1011

timvaillancourt:
This has been implemented in https://github.com/github/gh-ost/pull/844. Thanks @ajm188!

PR #844, opened 2020-05-10
Check RowsAffected when applying DML events to get more accurate statistics

ajm188:
Addresses #600.

When applying a DML event, check the RowsAffected on the `Result` struct. Since all DML event queries are point queries, this will only ever be 0 or 1. The applier then takes this value and multiplies by the `rowsDelta` of the event, resulting in a properly-signed, accurate row delta to use in the statistics.

If an error occurs here, log it, but do not surface this as an actual error .. simply assume the DML affected a row and move on. It will be inaccurate, but this is already the case.


timvaillancourt (approved):

shlomi-noach (approved):

shlomi-noach:
Noting that we've seen issues with `exact-rowcount` taking longer to run than the entire migration, and are looking to recommend disabling `exact-rowcount`.

go/logic/applier.go
```diff
@@ -1043,11 +1044,20 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+				result, err := tx.Exec(buildResult.query, buildResult.args...)
+				if err != nil {
 					err = fmt.Errorf("%s; query=%s; args=%+v", err.Error(), buildResult.query, buildResult.args)
 					return rollback(err)
 				}
-				totalDelta += buildResult.rowsDelta
+
+				rowsAffected, err := result.RowsAffected()
+				if err != nil {
+					log.Warningf("error getting rows affected from DML event query: %s. this might not be supported by go-sql-driver. i'm going to assume that the DML affected a row, but this may result in inaccurate statistics", err)
```
timvaillancourt, line 1055: @ajm188 if rows affected is not supported by the driver this will print a LOT of log messages during a migration

Would it be acceptable to silently default to `1` row affected here, ie: remove the `log.Warningf()`? cc @shlomi-noach for thoughts

Aside from resolving this question, this PR is ready to merge after a lot of successful tests 👍

shlomi-noach: > if rows affected is not supported by the driver

I don't see that there should be a driver support. The driver we embed supports it. I think the error message is misleading. There could be an error because we're talking to a database and there can always be some connection issue. But the feature _is_ supported. So I can't see that this log will be spamming.

timvaillancourt: Right, `database/sql` mentions that not all drivers support `.RowsAffected()` but because we hard-import the `go-sql-driver` that DOES support it this is a non-issue. Good point 👍 

In that case I suggest this change to the error:
```suggestion
					log.Warningf("error getting rows affected from DML event query: %s. i'm going to assume that the DML affected a single row, but this may result in inaccurate statistics", err)
```

cc @shlomi-noach and @ajm188

ajm188: works for me!

timvaillancourt:
Merging. Thanks @ajm188!

PR #846, opened 2020-05-18
Cancel any row count queries before attempting to cut over

### LINKED ISSUE #830: The --exact-rowcount option is buggy on large tables
### Context

During a schema change, gho-st copies every row from the original table to a ghost table. This results in a lock on the original table and a long running query similar to the following

```
mysql> show open tables where in_use>0;
+------------+----------------------+--------+-------------+
| Database   | Table                | In_use | Name_locked |
+------------+----------------------+--------+-------------+
| my_service | my_table             |      1 |           0 |
+------------+----------------------+--------+-------------+
1 row in set (0.00 sec)

mysql> show full processlist;
...
| 1594199 | my_service_ddl | 10.137.166.91:35626  | my_service         | Query       |    0 | Sending data                                                  | select  /* gh-ost `my_service`.`my_table` iteration:108 */
						`id`
					from
						`my_service`.`my_table`
					where ((`id` > _binary'1131891')) and ((`id` < _binary'310834793') or ((`id` = _binary'310834793')))
					order by
						`id` asc
					limit 1
					offset 9999 |
```

It appears that when [`--exact-rowcount`](https://github.com/github/gh-ost/blob/master/doc/command-line-flags.md#exact-rowcount) is enabled, ghost grabs another lock on the original table and starts an addition query to calculate a more accurate progress ETA. By default, row counting happens in parallel to row copying. 

```
mysql> show open tables where in_use>0;
+------------+----------------------+--------+-------------+
| Database   | Table                | In_use | Name_locked |
+------------+----------------------+--------+-------------+
| my_service | my_table             |      2 |           0 |
+------------+----------------------+--------+-------------+
1 row in set (0.00 sec)

mysql> show full processlist;
...
| 1594046 | my_service_ddl | 10.137.166.91:57666  | my_service         | Query       |   71 | Sending data                  | select /* gh-ost */ count(*) as rows from `my_service`.`my_table`                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                |
...
```

### Problem
After all the rows are copied to the ghost table, gh-ost attempts to cut-over the original table. 

Usually, the row counting query finishes before all the rows are copied to the ghost table. At the time of cut-over, because both queries have finished, there is no lock on the origin table. Hence the schema change is successful. 

However, we repeatedly found that on a large table (> 400GB), row counting query takes longer than row copying. At the time of cut-over, the row counting query is still running and holding a lock on the original table. 

This causes the schema change to fail with this error: "Error 1205: Lock wait timeout exceeded; try restarting transaction", which was described in https://github.com/github/gh-ost/issues/773

### Proposals
If you run into this issue, try disabling `--exact-rowcount`.

Improvements that we could make
* Document this bug; add a warning.
* Teach gh-ost to terminate the row counting query before cutting over the original table.

ajm188:
Closes #830. Switches from using `QueryRow` to `QueryRowContext`, and
stores a context.CancelFunc in the migration context, which is called to
halt any running row count query before beginning the cut over.

Note: I'm 99% sure `database/sql` does what I think it should do here. I wanted to open this now in case anyone else happens to know off-hand, but in the meantime I'll keep reading through the go stdlib to convince myself this is right.

Update on the above: I smartened up and read the docs instead of the code, which say [`go-sql-driver` does support cancelling queries via contexts](https://github.com/go-sql-driver/mysql#contextcontext-support)


go/logic/migrator.go
```diff
@@ -401,6 +406,10 @@ func (this *Migrator) Migrate() (err error) {
 	}
 	this.printStatus(ForcePrintStatusRule)
 
+	if this.migrationContext.CountTableRowsCancelFunc != nil {
```
shlomi-noach, line 409: Needs to be thread protected. Potential concurrent change: https://github.com/github/gh-ost/pull/846/files#diff-5aa50dde7ca1baf7a819eecf740ea220R543

go/logic/inspect.go
```diff
@@ -521,17 +522,28 @@ func (this *Inspector) estimateTableRowsViaExplain() error {
...
 		return err
 	}
+	select {
+	case <-ctx.Done():
+		log.Infof("exact row count cancelled (%s), likely because I'm about to cut over", ctx.Err())
+		return nil
+	default:
+		// row count query finished. nil out the cancel func, so the main migration thread
+		// doesn't bother calling it after row copy is done.
+		this.migrationContext.CountTableRowsCancelFunc = nil
```
shlomi-noach, line 543: Needs to be thread protected. Potential concurrent read: https://github.com/github/gh-ost/pull/846/files#diff-e2990679ccfa83d6fc15414958333bcbR409

go/logic/inspect.go
```diff
@@ -521,17 +522,28 @@ func (this *Inspector) estimateTableRowsViaExplain() error {
...
+func (this *Inspector) CountTableRows(ctx context.Context) error {
 	atomic.StoreInt64(&this.migrationContext.CountingRowsFlag, 1)
 	defer atomic.StoreInt64(&this.migrationContext.CountingRowsFlag, 0)
 
 	log.Infof("As instructed, I'm issuing a SELECT COUNT(*) on the table. This may take a while")
 
 	query := fmt.Sprintf(`select /* gh-ost */ count(*) as rows from %s.%s`, sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.OriginalTableName))
 	var rowsEstimate int64
-	if err := this.db.QueryRow(query).Scan(&rowsEstimate); err != nil {
+	if err := this.db.QueryRowContext(ctx, query).Scan(&rowsEstimate); err != nil {
```
shlomi-noach, line 533: To the best of my understanding, the `go-sql-driver` does not actually kill the query when context is cancelled, it is on the user to issue a `KILL` query. In which case we should first obtain a `Connection` so that we can get its connection_id, which we can then use to kill it.

ajm188: You're right! I regret this statement 😅 :

> Update on the above: I smartened up and read the docs instead of the code, which say go-sql-driver does support cancelling queries via contexts

I think the row count has to be pushed into a goroutine then, so `CountTableRows` can simultaneously wait on it and on a signal to kill the count query.

timvaillancourt (approved):

timvaillancourt:
@ajm188 this PR makes a lot of sense to me. Thanks a lot!

Do you mind rebasing the PR? 🙏

timvaillancourt:
@ajm188 I was able to fix the CI failures myself via the GitHub UI 🤯

@dm-2 / @rashiq can we merge this and test in a train of other recent PRs?

ajm188:
> @ajm188 I was able to fix the CI failures myself via the GitHub UI 🤯

awesome, thanks for taking care of it!

dm-2 (approved):

7koma95 (approved):

Issue #857, opened 2020-07-06
Run integration tests (gh-ost-replica-tests) in docker


ajm188:
Currently the integration tests are only runnable in CI, or if you do a bunch of mucking with your machine to get it runnable.

Providing a way for running the `localtests` with only docker and go as dependencies makes it easier for devs to more confidently make changes to gh-ost.

PR #858, opened 2020-07-06
Run `localtests` via docker

ajm188:
Related issue: #857 

### Description

This PR adds support for running `localtests` via docker.

It adds:
* Dockerfile.itest -- set up with dbdeployer and ready to run tests
* script/itest -- wrapper that builds gh-ost from source, builds the itest container and runs it
* itest.sh -- thing that actually runs the tests, expects to only be run in the itest container

### Testing

Example output: https://gist.github.com/ajm188/c83312ef82f06d51ca1f182c22b45d28

### DIFF
--- a/Dockerfile.itest
+++ b/Dockerfile.itest
@@ -1,0 +1,33 @@
+# See https://gist.github.com/ajm188/9488bc9c2b5b10d645f5e168e94cfb77
+# for why this is xenial
+FROM ubuntu:xenial
+
+ARG DBDEPLOYER_VERSION=1.52.0
+ARG MYSQL_VERSION_APT=5.7
+ARG MYSQL_VERSION=5.7.26
+
+RUN echo ${MYSQL_VERSION} > /mysql_version
+
+# We install mysql-server as a hack to make sure dbdeployer has all libs it
+# needs to run mysqls.
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        ca-certificates \
+        mysql-server-${MYSQL_VERSION_APT} \
+        wget \
+    && apt-get autoremove
+
+RUN wget https://github.com/datacharmer/dbdeployer/releases/download/v${DBDEPLOYER_VERSION}/dbdeployer-${DBDEPLOYER_VERSION}.linux.tar.gz \
+    && tar -xzvf dbdeployer-${DBDEPLOYER_VERSION}.linux.tar.gz \
+    && chmod +x dbdeployer-${DBDEPLOYER_VERSION}.linux \
+    && mv dbdeployer-${DBDEPLOYER_VERSION}.linux /usr/local/bin/dbdeployer
+
+RUN dbdeployer init --skip-all-downloads --skip-shell-completion \
+    && dbdeployer downloads get-by-version ${MYSQL_VERSION} --minimal \
+    && dbdeployer unpack mysql-${MYSQL_VERSION}.tar.xz --unpack-version ${MYSQL_VERSION}
+
+ADD itest.sh /itest.sh
+ADD bin/gh-ost /usr/local/bin/gh-ost
+ADD localtests /localtests
+
+CMD ["/itest.sh"]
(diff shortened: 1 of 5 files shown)


Dockerfile.itest
```diff
@@ -0,0 +1,33 @@
...
+ARG MYSQL_VERSION=5.7.26
+
+RUN echo ${MYSQL_VERSION} > /mysql_version
+
+# We install mysql-server as a hack to make sure dbdeployer has all libs it
+# needs to run mysqls.
+RUN apt-get update \
+    && DEBIAN_FRONTEND=noninteractive apt-get install -y --no-install-recommends \
+        ca-certificates \
+        mysql-server-${MYSQL_VERSION_APT} \
```
shlomi-noach, line 16: Why do we need `mysql-server`? We are using dbdeployer-based server, right?

Dockerfile.itest
```diff
@@ -0,0 +1,33 @@
...
+        wget \
+    && apt-get autoremove
+
+RUN wget https://github.com/datacharmer/dbdeployer/releases/download/v${DBDEPLOYER_VERSION}/dbdeployer-${DBDEPLOYER_VERSION}.linux.tar.gz \
```
shlomi-noach, line 20: Would you be willing to use https://github.com/github/gh-ost-ci-env for:
- getting `dbdeployer`
- getting the various MySQL server versions?

This way both this docker image and the  `migration tests` CI job will use the exact same dbdeployer+mysql versions. Please see how https://github.com/github/gh-ost/blob/master/.github/workflows/replica-tests.yml pulls the binaries.

itest.sh
```diff
@@ -0,0 +1,219 @@
...
+
+if [ -f /mysql_version ];
+then
+  MYSQL_VERSION="$(cat /mysql_version)"
+else
+  MYSQL_VERSION="5.7.26"
+fi
+
+echo "Creating replication sandbox"
+dbdeployer deploy replication $MYSQL_VERSION \
```
shlomi-noach, line 43: This sets up one master, two replicas (the default for `dbdeployer` replication setup). Let's save time and memory and set up a single replica, instead. Again, see https://github.com/github/gh-ost/blob/master/script/cibuild-gh-ost-replica-tests:

https://github.com/github/gh-ost/blob/4dab06e92b20954b46301e3b176d3a873615e3af/script/cibuild-gh-ost-replica-tests#L43

itest.sh
```diff
@@ -0,0 +1,219 @@
+#!/bin/bash
```
shlomi-noach, line 1: This looks like a lot of duplication from https://github.com/github/gh-ost/blob/master/script/cibuild-gh-ost-replica-tests

Can we instead just call https://github.com/github/gh-ost/blob/master/script/cibuild-gh-ost-replica-tests? Or refactor into a more generic script? There's otherwise too much duplication; the entire test logic is rewritten here; I'd like to see both Docker and the GitHub CI use the same code.

script/dock
```diff
@@ -15,11 +15,11 @@ case "$command" in
   "pkg")
     packages_path="${2:-/tmp/gh-ost-release}"
     docker_target="gh-ost-packaging"
-    docker build . -f Dockerfile.packaging -t "${docker_target}" && docker run --rm -it -v "${packages_path}:/tmp/pkg" "${docker_target}:latest" bash -c 'find /tmp/gh-ost-release/ -maxdepth 1 -type f | xargs cp -t /tmp/pkg'
+    docker build . -f Dockerfile.packaging -t "${docker_target}" && docker run --rm -it --mount type=bind,source="${packages_path}",target=/tmp/pkg "${docker_target}:latest" bash -c 'find /tmp/gh-ost-release/ -maxdepth 1 -type f | xargs cp -t /tmp/pkg'
```
shlomi-noach, line 18: could you please explain this change?

script/itest
```diff
@@ -0,0 +1,8 @@
+#!/bin/sh
+#
+# Wrapper for running integration tests in docker.
+
+./script/dock pkg $(pwd)/bin
```
shlomi-noach, line 5: This seems to assume that the `gh-ost` binary in `$(pwd)/bin` is up to date and built from the current branch/changes. This may not be the case.

ajm188:
I haven't forgotten about this! I should have time later this week.

meiji163:
Closing because dbdeployer is deprecated (https://github.com/github/gh-ost/pull/1529)

PR #872, opened 2020-08-09
Support ALTER ONLINE TABLE syntax

shlomi-noach:
This PR supports an experimental syntax: `ALTER ONLINE TABLE ...`, which has the same syntax as `ALTER TABLE` except for the addition of `ONLINE`.

This syntax isn't supported by MySQL: it is a Vitess experiment (PR link to follow) with special magic syntax. To that effect, `gh-ost` will silently accept `--alter="alter online table ..."` syntax as valid. It otherwise doesn't do anything special if the word `ONLINE` appears.


shlomi-noach:
Sorry, meant to open this on https://github.com/openark/gh-ost

PR #879, opened 2020-08-20
Fix buster tarball build

timvaillancourt:
### Description

The `gh-ost-build-deploy-tarball` CI job is now running under Debian `buster` and is no longer creating `stretch` and `jessie` tarballs like before

This PR continues a hack that was previously added for `jessie`/`stretch` until a better solution emerges

### DIFF
--- a/script/build-deploy-tarball
+++ b/script/build-deploy-tarball
@@ -28,8 +28,10 @@ gzip $tarball
 mkdir -p "$BUILD_ARTIFACT_DIR"/gh-ost
 cp ${tarball}.gz "$BUILD_ARTIFACT_DIR"/gh-ost/
 
-### HACK HACK HACK ###
-# blame @carlosmn and @mattr-
-# Allow builds on stretch to also be used for jessie
+### HACK HACK HACK HACK ###
+# blame @carlosmn, @mattr and @timvaillancourt-
+# Allow builds on buster to also be used for streth + jessie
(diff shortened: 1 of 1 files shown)


script/build-deploy-tarball
```diff
@@ -28,8 +28,10 @@ gzip $tarball
...
-### HACK HACK HACK ###
-# blame @carlosmn and @mattr-
-# Allow builds on stretch to also be used for jessie
-jessie_tarball_name=$(echo $(basename "${tarball}") | sed s/-stretch-/-jessie-/)
+### HACK HACK HACK HACK ###
+# blame @carlosmn, @mattr and @timvaillancourt-
+# Allow builds on buster to also be used for stretch + jessie
+stretch_tarball_name=$(echo $(basename "${tarball}") | sed s/-buster-/-stretch-/)
+jessie_tarball_name=$(echo $(basename "${stretch_tarball_name}") | sed s/-stretch-/-jessie-/)
+cp ${tarball}.gz "$BUILD_ARTIFACT_DIR/gh-ost/${stretch_tarball_name}.gz"
```
eguevara, line 36: Does it need the extra stretch artifact?  I would think it only needs to find the jessie build?

timvaillancourt: Yes, we use `stretch` and `jessie` still @eguevara. It's actually the `buster` build that will go unused

shlomi-noach:
LOL

Issue #381, opened 2017-03-03
The AUTO_INCREMENT has been changed


JiaqingXu:
> This is the place to report a bug, ask a question, or suggest an enhancement.

When we  delete some data, and the max value of AUTO_INCREMENT columns is not match the AUTO_INCREMENT of the table. Then we use ghost to do some alter .The  AUTO_INCREMENT of the table have be changed.

> This is also the place to make a discussion before creating a PR.

> If this is a bug report, please provide a test case (e.g., your table definition and gh-ost command) and the error output.
> Please use markdown to format code or SQL: https://guides.github.com/features/mastering-markdown/

```
CREATE TABLE ghost_test(
    -> `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
    -> val int(11) NOT NULL DEFAULT '0',
    -> PRIMARY KEY (`id`)
    -> )ENGINE=InnoDB;
```
we insert 3 rows:
```
>select * from ghost_test;
+----+-----+
| id | val |
+----+-----+
|  1 |   1 |
|  2 |   2 |
|  3 |   3 |
+----+-----+
3 rows in set (0.00 sec)
```
delete 1 row 
```
>delete from ghost_test where id =3;
Query OK, 1 row affected (0.00 sec)

>select * from ghost_test;
+----+-----+
| id | val |
+----+-----+
|  1 |   1 |
|  2 |   2 |
+----+-----+
>show create table ghost_test\G
*************************** 1. row ***************************
       Table: ghost_test
Create Table: CREATE TABLE `ghost_test` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `val` int(11) NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=4 DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_unicode_ci
1 row in set (0.00 sec)
```
Then we used gh-ost to do "Alter table ghost_test engine=innodb"
```
show create table ghost_test\G
*************************** 1. row ***************************
       Table: ghost_test
Create Table: CREATE TABLE `ghost_test` (
  `id` bigint(20) unsigned NOT NULL AUTO_INCREMENT,
  `val` int(11) NOT NULL DEFAULT '0',
  PRIMARY KEY (`id`)
[...]

PR #911, opened 2021-01-05
Copy auto increment

shlomi-noach:
Solves https://github.com/github/gh-ost/issues/381

### Description

This PR ensures to copy `AUTO_INCREMENT` value to `gh-ost` table. The problem in https://github.com/github/gh-ost/issues/381 is that if rows are deleted from end of table, then the new table is created (or ends up) with a lower auto_increment value. I think most migrations do not encounter this issue because new rows are commonly added to the table while it's being migrated, naturally bumping up the _ghost_ table's `AUTO_INCREMENT` value to match the original table's. But for static tables this can happen. 

- parses user's `-alter` statement to see if the user has specified any explicit `AUTO_INCREMENT`
- reads original table's `AUTO_INCREMENT` value (if exists)
- if `AUTO_INCREMENT` exists, and user has not overriden, apply original table's `AUTO_INCREMENT` value onto the _ghost_ table.
- tests added to verify behavior.

Resubmission of https://github.com/openark/gh-ost/pull/12


timvaillancourt:
Merged in https://github.com/github/gh-ost/pull/967 👍

shlomi-noach:
Woohoo!

Issue #485, opened 2017-09-01
Add unique key cause data-integrity problem


cenkore:
#167 #477 
This problem is due to apply dml binlog and row-copy thread concurrent issue. And gh-ost self-mechanism would replace `insert` to `insert ignore into`(row-copy) or `replace into`(binlog apply).

When table have a primary key , and wanna add a unique key (or opposite, add primary key),there have two conditions would cauze data-integrity problem
> 1. original table data could not match, and would be duplicate in new unique key
> 2. app write "duplicate" rows into original table when gh-ost running

In fact this two condition is similar , and gh-ost do not need to care it. As try to fix it

`/*
create table member (id int , name varchar(10) , nickname varchar(10),primary key(id));
insert into member values(1,'thsun','cenkore');
*/`

> -- ddl: add unique key uk(name,nickname);

If i execute `insert into member values(2,'thsun','cenkore')` in a right opportunity when gh-ost is running

1. First binlog apply and then row-copy

> if gh-ost apply binlog firstly and execute `replace into _del_member_gho values(2,'thsun','cenkore')` -> success
> then row-copy (1,'thsun','cenkore'),(2,'thsun','cenkore') and execute
> `insert ignore into _del_member_gho values(1,'thsun','cenkore')`  -> ignore , data-integrity problem
> `insert ignore into _del_member_gho values(2,'thsun','cenkore')`  -> ignore

2. First row-copy and then binlog apply

> if row-copy firstly and execute `insert ignore into _del_member_gho values(1,'thsun','cenkore')` -> success
> then have two sub-conditon , 1.1 row-copy and binlog apply 1.2 binlog apply and row-copy
> 	1.1 row-copy continuously and execute `insert ignore into _del_member_gho values(2,'thsun','cenkore')` -> ignore
> 	at last binlog apply and execute `replace into _del_member_gho values(2,'thsun','cenkore')`  -> data-integrity problem
> 
> 	1.2 binlog apply firstly and execute `replace into _del_member_gho values(2,'thsun','cenkore')` -> data-integrity problem
> 	at last row-copy another record and execute `insert ignore into _del_member_gho values(2,'thsun','cenkore')` -> ignore

In this two condition , user may be hope failure and bail out , but not data loss.

Condition 1 `First binlog apply and then row-copy`:
before row-copy , gh-ost can validate whether there already have primary key data in the _gho table,
if true, execute `insert ignore into`, otherwise switch to  execute `insert into` , As follows replay condition 1
[...]

Issue #477, opened 2017-08-28
fail in case of adding unique index by default


pznamensky:
Hi,
I'm about https://github.com/github/gh-ost/issues/167.
We're faced with the same behavior and it was unexpected for us.
Previously we used `pt-online-schema-change` and it failed and showed a query to check if there were duplicated entries when it detected adding unique index.
Didn't you think about adding flag like `-allow-unique-key` and fail in case of adding an unique key by default?
Or adding something like `-fail-on-adding-unique-key` which will enable checking for adding an unique key.
`pt-online-schema-change`'s `check-unique-key-change` flag documented here: https://www.percona.com/doc/percona-toolkit/LATEST/pt-online-schema-change.html#id7
It would be great to have the same behavior in `gh-ost`.

shlomi-noach:
Hi,
I'm not sure how I feel about adding the flag. I don't wholly object to adding it, but it feels to me like if there were a `-allow-add-unique-key` flag, people would be annoyed at migrations continuously failing on adding unique keys, and would script the `-allow-add-unique-key` flag, hard coded, anyhow. Warnings are bound to be ignored. At least that's my impression.

I like the use of `-fail-on-adding-unique-key`, which would make `gh-ost` execution backwards-compatible, but then again, people may not realize this flag exists anyhow.

I'm interested in more opinions on this.

Issue #167, opened 2016-08-18
when add unique index may lose data


rj03hou:
table and data
```
create table test (id int not null primary key, name char(255));
insert into test values(1,'a'),(2,'b'),(3,'a');

alter table test add unique index uniq_name(name);
```
when execute the alter ddl, the value of (1,'a') should loss, and pt-online-schema-change have this problem tool.
I think the good solution is check if have duplicate value then fail, or have a config not allow add unique index.

shlomi-noach:
This is beyond the scope of `gh-ost` to handle, and is within the scope of the DBA avoiding shooting herself in the foot.

Checking for duplicates could take hours, and by the time the check is complete, new duplicates may be introduced.

rj03hou:
when we use the pt-osc, and some RD complain that use our tool have lose data. so if they add unique index we suggest they must be careful.

shlomi-noach:
Good suggestion.

isegal:
Is there a way to make gh-ost fail with an exit code if there is an attempt to insert a row with a duplicate key during the copy operation? In other words, we are ok with taking hours on large tables so long as there is no silent failure with a data loss. The preferred way is to notify the user and exit, allowing the invoking tools/scripts to take appropriate action such as cleaning up the temporary gh-ost tables, etc.

shlomi-noach:
> Is there a way to make gh-ost fail with an exit code if there is an attempt to insert a row with a duplicate key during the copy operation?

I believe this to be impractical, adding a `SELECT` for every single `UPDATE` or `INSERT` `gh-ost` executes, potentially on a non-indexed column.

isegal:
Unless you are doing upsert, the INSERT operation itself will throw a 'duplicate key' error if an attempt is made to insert a unique value that already exists. Not sure about UPDATE, but I can do a quick test if you like.

isegal:
Just did a quick glance, apologize ahead of time if I'm looking in the wrong code.
https://github.com/github/gh-ost/blob/master/go/sql/builder.go#L218
it looks like it's doing an 'insert ignore', which would (by design?) ignore any issues such as a duplicate key. What we'd like to propose is some kind of command line switch that will do a non-ignore insert, and cause gh-ost to exit with an error if there is an issue writing to the destination table. I.e. a strict mode so to speak.

shlomi-noach:
[...]

PR #912, opened 2021-01-08
fix data-integrity problem when add unique key

cenkore:
Hi,
Add a PR for #485  #477  #167 . 

Insert syntax in gh-ost will be rewritten with `insert ignore into` (row-copy) and `replace into`(binlog apply goroutine) . if the data in the original table or data after apply would cause conflicts, means does not meet the conditions for adding a unique key, data will be lost in current version. This PR can detect data conflicts and exit, including conflicts in the original table data, and the data after binlog apply. PTAL.

The error looks like the following and is very easy to reproduce：
- original conflict
 >> 2021-01-08 16:58:15 ERROR Error 1062: Duplicate entry 'cenkore' for key 'name'
- apply confict
>> 2021-01-08 16:59:37 ERROR Error 1062: Duplicate entry 'cenkore' for key 'name'; query=
			insert /* gh-ost `adetestdb`.`_testuk_gho` */ into
				`adetestdb`.`_testuk_gho`
					(`id`, `name`, `location`, `inserttime`)
				values
					(?, ?, ?, ?)
		; args=[3 cenkore cn 2021-01-08 16:59:30]

thanks.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -222,6 +222,8 @@ type MigrationContext struct {
 	recentBinlogCoordinates mysql.BinlogCoordinates
 
 	Log Logger
+
+	IsAddUniqueKey	bool
 }
 
 type Logger interface {
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -478,6 +478,7 @@ func (this *Applier) ApplyIterationInsertQuery() (chunkSize int64, rowsAffected 
 		this.migrationContext.MigrationIterationRangeMaxValues.AbstractValues(),
 		this.migrationContext.GetIteration() == 0,
 		this.migrationContext.IsTransactionalTable(),
+		this.migrationContext.IsAddUniqueKey,
 	)
 	if err != nil {
 		return chunkSize, rowsAffected, duration, err
@@ -959,7 +960,7 @@ func (this *Applier) updateModifiesUniqueKeyColumns(dmlEvent *binlog.BinlogDMLEv
 
 // buildDMLEventQuery creates a query to operate on the ghost table, based on an intercepted binlog
 // event entry on the original table.
-func (this *Applier) buildDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) (results [](*dmlBuildResult)) {
+func (this *Applier) buildDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent, secondary bool) (results [](*dmlBuildResult)) {
 	switch dmlEvent.DML {
 	case binlog.DeleteDML:
 		{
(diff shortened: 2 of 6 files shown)


shlomi-noach:
Adding a `UNIQUE KEY` while table is being written to, is by definition a race condition, unless you can guarantee the application never writes duplicates.

Can you please clarify what is the exact use case (possibly wit ha test) that this PR solves?

Thank you!

cenkore:
Thanks for your quick reply. @shlomi-noach 

In this PR, if duplicates are on the newly added unique key , it will fail instead of ignore.

### Background
Gh-ost contains two threads to process the data of the published table, one thread copies the data of the original table, using Insert ignore into, and the other thread is used to process incremental data, by parsing and applying the binlog, using replace into; Same data is likely to be processed twice because of two threads are processed concurrently. In order to avoid conflicts leading to failure, insert ignore into and replace into are designed. However, when adding a unique key will cause the actual duplicate data on the newly added unique key would be ignored, because duplicate data will be ignored or replaced. This will cause data loss. In extreme cases, if you add a new column and add a unique key to this column at the same time, then there will be only one record left after the table is published. It may cause serious problems.

### Proposal 
Based on the above background, we can do special processing for the scene. For data copy, only copy the data that primary key value is not in _xx_gho, which can be rewritten as insert into _xx_gho select * from xx a left join _xx_gho b on a.pk = b.pk where b.pk is null; for binlog apply, You can rewrite replace into as delete + insert.

### Rationale

The data copy, only the primary key value is not in the gho table is copied, but we don’t need use `ignore` , so we only ignore the conflict of the primary key value, but the conflict on the newly added unique key would return error; and the incremental part, assuming that the data has not been copied, delete is a dry-run. If there is a conflict with the newly added unique key, it will fail, but replace cannot do this. If the data has been copied, delete the primary key value and execute insert again. Similarly, if the data conflicts with the newly added unique key, it will also fail.

Thank you!

cenkore:
Add 2 test cases

original table definition: `create table uk(id int, name varchar(10),inser_time datetime, primary key(id))`
alter: `add unique key(name)`

### Test case 1:

prepare: `insert into uk values(1,'cenkore',now())`

execute: `./gh-ost --host=10.0.0.1 --database=adetestdb --table=uk  --verbose --alter="add unique key(name)" --concurrent-rowcount --conf="/usr/local/gh-ost.cnf"  --port=66666 --nice-ratio=0.1 --execute  --initially-drop-ghost-table  --initially-drop-old-table --postpone-cut-over-flag-file="/tmp/uk.postpone"`

We can insert a duplicate row before cutover table by popstpone flag, insert sql: `insert into uk values(2,'cenkore',now())`

result:
  - `current version`: handle success, BUT loss one row which id is 1, new table value as follows:
    > +----+---------+---------------------+  
     | id | name    | inser_time          |   
     +----+---------+---------------------+  
     |  2 | cenkore | 2021-01-21 11:06:45 |  
     +----+---------+---------------------+  

  - `this pr` : handle failure with error:
    > 2021-01-21 11:59:31 FATAL Error 1062: Duplicate entry 'cenkore' for key 'name'; query=  
                         insert /* gh-ost `adetestdb`.`_uk_gho` */ into  
                                 `adetestdb`.`_uk_gho`  
                                         (`id`, `name`, `inser_time`)  
                                 values  
                                        (?, ?, ?)  
                 ; args=[2 cenkore 2021-01-21 11:58:31]  


### Test case 2:

prepare: `insert into uk values (1,'cenkore',now()),(2,'cenkore',now())`

execute: `./gh-ost --host=10.0.0.1 --database=adetestdb --table=uk  --verbose --alter="add unique key(name)" --concurrent-rowcount --conf="/usr/local/gh-ost.cnf"  --port=66666 --nice-ratio=0.1 --execute  --initially-drop-ghost-table  --initially-drop-old-table`

result:
  - `current version`: handle success, BUT loss one row which id is 2, new table value as follows:
    > +----+---------+---------------------+  
     | id | name    | inser_time          |  
     +----+---------+---------------------+  
     |  1 | cenkore | 2021-01-21 11:51:01 |  
     +----+---------+---------------------+  

  - `this pr` : handle failure with error:
    > 2021-01-21 11:56:38 FATAL Error 1062: Duplicate entry 'cenkore' for key 'name'

shlomi-noach:
@cenkore thank you for the PR and for the explanation.

I've been thinking, possibly we don't need to change the `insert ignore`. We can keep that as it is, without the `left join`. It may be enough to replace the `replace into` with (`delete where pk=?` + `insert into`). I think the logic is still correct.
I'm not sure yet. Can you see a problem with that?

shlomi-noach:
Oh, strike that. The trivial case when the table is pre-populated with conflicting values + zero traffic in the binary logs is enough to prove my thinking was wrong.

shlomi-noach:
@cenkore would you mind submitting an identical PR on https://github.com/openark/gh-ost/, where I can take a closer look please?

cenkore:
> @cenkore would you mind submitting an identical PR on https://github.com/openark/gh-ost/, where I can take a closer look please?

Ok, I will submit later.  Thank you.

cenkore:
@shlomi-noach  I submitted to `openark/gh-ost`, but I can’t guarantee the accuracy of my operation, because it's difficult for me which merge is between two forked sibling projects, and your project is ahead of `github/gh-ost`. :D

shlomi-noach:
@cenkore thank you so much for the effort! I will double check.

dtaveras:
@shlomi-noach @cenkore Is there any way to ignore when update set produces a unique key constraint violation? Sorry this is slightly unrelated but was hoping you guys knew the answer. This awesome work both the fork and gh-ost.

jiangnanora:
When will this pr be merged

AndrewHuangMiao:
When this will be fixed?

Issue #909, opened 2020-12-31
Data loss on tables with binary primary key


sfairat15:
# Data loss on tables with binary primary key

Hello!

We found bug on our production environment when migrate tables with binary columns as primary key.
It raises occationally, only at some updated rows, when gh-ost apply changes from binlog.
Usually, only few rows affected, looks like there are rows with special symbols at primary key.

## How to reproduce

- Create table

```sql
CREATE TABLE `testorders` (
  `UUId` binary(16) NOT NULL,
  `Version` int(11) NOT NULL DEFAULT '1',
  `ClientUUId` binary(16) DEFAULT NULL,
  `CreatedAtUtc` datetime NOT NULL,
  `ModifiedAtUtc` datetime DEFAULT NULL,
  PRIMARY KEY (`UUId`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4;
```

- Run gh-ost

```console
./gh-ost \
      --assume-rbr \
      --assume-master-host="localhost" \
      --master-user="ghost_user" \
      --master-password="PASSWORD" \
      --ssl \
      --ssl-allow-insecure \
      --max-load=Threads_running=32 \
      --max-lag-millis=1500 \
      --critical-load=Threads_running=500 \
      --chunk-size=5000 \
      --host="localhost" \
      --user="ghost_user" \
      --password="PASSWORD" \
      --alter="ADD INDEX IDX_ClientUUId(ClientUUId)" \
      --database="testghost" \
      --table="testorders" \
      --skip-strict-mode \
      --debug \
      --stack \
      --postpone-cut-over-flag-file="/tmp/ghost.cutover.testorders.flag" \
      --panic-flag-file="/tmp/ghost.panic.testorders.flag" \
      --allow-on-master \
      --execute
```

- Insert rows into the table

```sql
# bad UUId
insert into testorders (UUId, Version, ClientUUId, CreatedAtUtc)
values(0x150B7B5B516F87125E8597FFB44BF200, 1, 0xE53C111AFEF8A2D3B9A3484108D08F2B, '2020-12-31 07:18:00');

# good UUId
insert into testorders (UUId, Version, ClientUUId, CreatedAtUtc)
values(0x002E363F305044DE79E9B3D0105BC23E, 1, 0xE53C111AFEF8A2D3B9A3484108D08F2E, '2020-12-31 07:19:00');
```

- Update rows

```sql
# bad UUId - 0 rows affected at gh-ost log
update testorders
set Version = 2,
[...]

PR #915, opened 2021-01-13
fix: issue 909

cenkore:
Add a PR for #909 

The reason is that the last byte of 0x00 was truncated when mysql wrote Binlog, so the data obtained when gh-ost parsed binlog through go-mysql was already truncated

you can read binlog with hexdump as follows, or debug at [here](https://github.com/siddontang/go-mysql/blob/master/replication/row_event.go#L1169) , you will get the actual length is 15

> Position  Timestamp   Type   Master ID        Size      Master Pos    Flags     
     c86 60 14 fc 5f   20   35 23 00 00   7f 00 00 00   05 0d 00 00   00 00    
     c99 99 5e 00 00 00 00 01 00  02 00 05 ff f0 10 00 2e |................|    
     ca9 36 3f 30 50 44 de 79 e9  b3 d0 10 5b c2 3e 01 00 |6.0PD.y.........|    
     cb9 00 00 10 e5 3c 11 1a fe  f8 a2 d3 b9 a3 48 41 08 |.............HA.|    
     cc9 d0 8f 2e 99 a8 3e 74 c0  e0 0f **15 0b 7b 5b 51 6f** |......t.......Qo|    
     cd9 **87 12 5e 85 97 ff b4 4b  f2** 02 00 00 00 10 e5 3c |.......K........|    
     ce9 11 1a fe f8 a2 d3 b9 a3  48 41 08 d0 8f 2b 99 a8 |........HA......|    
     cf9 3e 74 80 99 a8 3e 75 00  43 17 a8 3b             |.t....u.C...|    
      Delete_rows: table id 24217 flags: STMT_END_F

### DIFF
--- a/go/sql/types.go
+++ b/go/sql/types.go
@@ -44,14 +44,14 @@ type Column struct {
 	timezoneConversion *TimezoneConversion
 }
 
-func (this *Column) convertArg(arg interface{}) interface{} {
+func (this *Column) convertArg(arg interface{}, isUniqueKeyColumn bool) interface{} {
 	if s, ok := arg.(string); ok {
 		// string, charset conversion
 		if encoding, ok := charsetEncodingMap[this.Charset]; ok {
 			arg, _ = encoding.NewDecoder().String(s)
 		}
 
-		if this.Type == BinaryColumnType {
+		if this.Type == BinaryColumnType && isUniqueKeyColumn {
 			arg2Bytes := []byte(arg.(string))
 			size := len(arg2Bytes)
 			if uint(size) < this.BinaryOctetLength {
--- a/go/sql/types.go
+++ b/go/sql/types.go
@@ -10,6 +10,7 @@ import (
 	"reflect"
 	"strconv"
 	"strings"
+	"bytes"
 )
 
 type ColumnType int
@@ -22,6 +23,7 @@ const (
 	MediumIntColumnType
 	JSONColumnType
 	FloatColumnType
+	BinaryColumnType
 )
 
 const maxMediumintUnsigned int32 = 16777215
@@ -35,6 +37,10 @@ type Column struct {
 	IsUnsigned         bool
 	Charset            string
 	Type               ColumnType
+
+	// add Octet length for binary type, fix bytes with suffix "00" get clipped in mysql binlog.
+	// https://github.com/github/gh-ost/issues/909
+	BinaryOctetLength  uint
 	timezoneConversion *TimezoneConversion
 }
 
@@ -44,6 +50,18 @@ func (this *Column) convertArg(arg interface{}) interface{} {
 		if encoding, ok := charsetEncodingMap[this.Charset]; ok {
 			arg, _ = encoding.NewDecoder().String(s)
 		}
+
+		if this.Type == BinaryColumnType {
+			size := len([]byte(arg.(string)))
+			if uint(size) < this.BinaryOctetLength {
+				buf := bytes.NewBuffer([]byte(arg.(string)))
+				for i := uint(0); i < (this.BinaryOctetLength - uint(size)); i++ {
+					buf.Write([]byte{0})
+				}
+				arg = buf.String()
+			}
+		}
+
 		return arg
 	}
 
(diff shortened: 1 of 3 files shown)


rashiq:
Hi @cenkore! Thanks for the contribution!

It seems like the build is failing because it's it's not [properly formatted](https://github.com/github/gh-ost/runs/1695481331?check_suite_focus=true#step:4:16) with [gofmt](https://golang.org/cmd/gofmt/).

Could you please run the following command on the branch?

```
gofmt -s -w  go/
```

cenkore:
@rashiq , thanks for your suggestion, it looks good!

sfairat15:
Hi, @cenkore! We checked your PR on test db and it works like a charm! Thank you!

rashiq:
@cenkore Thanks! We've deployed it to our testing cluster, will monitor it the coming few days and if everything is fine we'll merge it 👍

cenkore:
> @cenkore Thanks! We've deployed it to our testing cluster, will monitor it the coming few days and if everything is fine we'll merge it 👍

Thank you!

vitalyu:
@rashiq is any updates? thx

go/sql/types.go
```diff
@@ -44,6 +50,18 @@ func (this *Column) convertArg(arg interface{}) interface{} {
...
+			size := len([]byte(arg.(string)))
+			if uint(size) < this.BinaryOctetLength {
+				buf := bytes.NewBuffer([]byte(arg.(string)))
+				for i := uint(0); i < (this.BinaryOctetLength - uint(size)); i++ {
+					buf.Write([]byte{0})
+				}
+				arg = buf.String()
+			}
+		}
+
```
shlomi-noach, line 64: Could you please explain what the logic above does? Thank you!

vitalyu: This is all about this issue https://github.com/github/gh-ost/issues/909
Also, U can read this https://dev.mysql.com/doc/refman/5.7/en/binary-varbinary.html
```
You should consider the preceding padding and stripping 
characteristics carefully if you plan to use the BINARY 
data type for storing binary data and you require that 
the value retrieved be exactly the same as the value stored. 
The following example illustrates how 0x00-padding 
of BINARY values affects column value comparisons
```
Please, read @cenkore's first message:
```
The reason is that the last byte of 0x00 was truncated 
when mysql wrote Binlog, so the data obtained when 
gh-ost parsed binlog through go-mysql was already truncated
```
So, this part of code appends missing zeroes.

We use this fork on production about a month, and all is OK.

go/sql/types.go
```diff
@@ -44,6 +50,18 @@ func (this *Column) convertArg(arg interface{}) interface{} {
 		if encoding, ok := charsetEncodingMap[this.Charset]; ok {
 			arg, _ = encoding.NewDecoder().String(s)
 		}
+
+		if this.Type == BinaryColumnType {
+			size := len([]byte(arg.(string)))
```
shlomi-noach, line 55: Minor nitpick: create a variable for `[]byte(arg.(string))` so that this doesn't need to evaluate twice.

go/sql/types.go
shlomi-noach: Thank you. Is there a limit on how many zeros you need to pad with, or can this number be very high?

go/sql/types.go
sfairat15: @cenkore could you please update code ^

cenkore: Sorry for the late reply，i'll update now.

go/sql/types.go
cenkore: @shlomi-noach Filling a high number zero bytes may affect processing efficiency. after thinking about it, we can fix it more lighter as below, other scenarios will not cause data consistency problems. https://github.com/github/gh-ost/pull/915/commits/40bc5aedf2ff1d85c10c4d31a594d0b176d149d7
- only handle the binary column in the unique key
- only handle update/delete dml event

go/sql/types.go
cenkore: https://github.com/github/gh-ost/pull/915/commits/960d8012762f7140096f8727e0ac9e6e68f72636

shlomi-noach (commented):
LGTM pending testing

rashiq:
Sorry it took so long! merged! 🎉

PR #935, opened 2021-03-03
Add go mod

timvaillancourt:
### Description

This is a merge PR of the user-submitted PR: https://github.com/github/gh-ost/pull/932

Unfortunately this step is required as our `gh-ost` testing won't run on external forks

cc @natewernimont-wk / @shlomi-noach


tomdeering-wf (approved):
Thanks for being willing to accept this change! Now we'll be able to build gh-ost with Go 1.16 without having to revert `go build` back to the `GO111MODULE=auto` (the default in Go 1.15).

natewernimont-wk (approved):

shlomi-noach (approved):
Shoot for the stars

timvaillancourt:
👋 @natewernimont-wk this ran into automated testing failures with the following output:
```
2021-03-10 04:24:24 FATAL Unable to read osc chunk size: strconv.ParseInt: parsing "": invalid syntax
```
A rollback to `master` was able to resolve this. Whatever the issue it is strange it didn't get caught in CI 🤔

I haven't had time to dig into "why", but the command that triggered the problem was:
```
gh-ost --conf=/etc/mysql/gh-ost.cnf --test-on-replica --database=REDACTED --table=REDACTED \
--verbose --alter=engine=innodb --execute --initially-drop-ghost-table --initially-drop-old-table \
--max-load=Threads_running=30 --critical-load=Threads_running=1000 --critical-load-interval-millis=3000 \
--switch-to-rbr --chunk-size=2500 --cut-over=default --timestamp-old-table --exact-rowcount \
--concurrent-rowcount --serve-socket-file=/tmp/gh-ost.test.sock --panic-flag-file=/tmp/ghost-test-panic.flag \
--hooks-path=REDACTED --hooks-hint=@root --ssl --ssl-allow-insecure --verbose
```

timvaillancourt:
@shlomi-noach this is failing on an error I could use some context around:
```
 go/logic/inspect.go:619:20: m.GetUint64 undefined (type sqlutils.RowMap has no field or method GetUint64)
```
It appears that function doesn't exist in `github.com/outbrain/golib/sqlutils` and I can't find a point in the history where it did exist. That lead me to wondering if a patch was committed to the `vendor/` dir of this repo but the git history doesn't support that either

I'm curious if you had any ideas how this may have happened. I suppose the options are to add it to `github.com/outbrain/golib` or make a copy of that logic as a utility-func in `go/` in this repo

shlomi-noach:
@timvaillancourt right. `GetUint64` was added downstream, directly into `vendor/` path, in https://github.com/openark/gh-ost/commit/31069ae4

What I'm suggesting:

- I have this better maintained repo: https://github.com/openark/golib
- I'll add `GetUint64` functionality in that repo
- you will add `replace github.com/outbrain/golib => github.com/openark/golib` in `go.mod`

I'm just trying to think whether there's been more changes local to `golib` under `vendor`. I'll either `git blame` later on, or just confirm the code builds properly with `openark/golib`.

shlomi-noach:
Related: https://github.com/outbrain/golib/pull/10

shlomi-noach:
https://github.com/openark/golib/pull/11 is merged, and `gh-ost` builds well with it.

@timvaillancourt either that `replace` directoive, or just search&replace `github.com/outbrain/golib` with `github.com/openark/golib` throughout the code at your discretion.

timvaillancourt:
> [openark/golib#11](https://github.com/openark/golib/pull/11) is merged, and `gh-ost` builds well with it.

@shlomi-noach thanks for the PR and background ❤️! I'll look out for this if this PR runs into any other problems 👍

shlomi-noach:
@timvaillancourt FYI I'm looking to use this `go mod` PR to upgrade `go-mysql` module, in https://github.com/openark/gh-ost/pull/22

timvaillancourt:
> @timvaillancourt FYI I'm looking to use this `go mod` PR to upgrade `go-mysql` module, in [openark#22](https://github.com/openark/gh-ost/pull/22)

Sounds good @shlomi-noach, I was debating on doing the same here. I think I'll make the `go-mysql` a 2nd PR

shlomi-noach:
> I think I'll make the go-mysql a 2nd PR

yeah let's do it in different steps. Otherwise the move `go mod` look good!

timvaillancourt:
Similar to my comment earlier, this is failing on the following error when ran on real hosts:
```
2021-06-08 05:27:28 FATAL Unable to read osc chunk size: strconv.ParseInt: parsing "": invalid syntax
```

I plan to dig into why this compiles/passes unit tests and then a fix 👍

shlomi-noach:
yes! Just reported yesterday by user: https://github.com/openark/gh-ost/pull/19#issuecomment-856025790

I think `go mod` updated the `cfg` package which in turn is not backwards-compatible. So this is the code that reads a `-conf=/some/file` config.

devkadirselcuk (approved):

timvaillancourt:
> yes! Just reported yesterday by user: [openark#19 (comment)](https://github.com/openark/gh-ost/pull/19#issuecomment-856025790)
> 
> I think `go mod` updated the `cfg` package which in turn is not backwards-compatible. So this is the code that reads a `-conf=/some/file` config.

@shlomi-noach right, there were some changes to the config logic here that seems suspicious: https://github.com/github/gh-ost/pull/935/files#diff-5cbecce2c65b6c6af5bf73697d1beeee2277d4a2ef4e8bdaaed6e1f5296ca5a8R809-R827

Hoping to dig into that soon unless you beat me to it! 😆

shlomi-noach:
> unless you beat me to it! laughing

I'll give you the honors

timvaillancourt:
This is passing real-world testing as https://github.com/github/gh-ost/pull/935/commits/71d27d0930e48bbc3567a50a96d60638c3f57518

If anyone has time to test this on more environments that would be much appreciated! 🙇

cc @natewernimont-wk / @tomdeering-wf / @shlomi-noach

timvaillancourt:
This has passed several-100 tests on real tables, so I'll merge this 👍

d-bytebase:
Thanks for providing go mod for gh-ost! When are we going to have 1.1.5 release for the change?

dm-2:
Hi @d-bytebase, we're currently investigating a data corruption bug that we believe may have been introduced by an updated dependency in this PR - once we've identified the cause and fixed it, the plan is to put out a new release based on the `master` branch (`1.1.3` and `1.1.4` are based on `1.1.2`).

chenrui333:
Looks like this change was never been released? Can we cut a release for this PR? Thanks!

Issue #953, opened 2021-04-06
Unwanted directories added to RPM


yakirgb:
When creating an RPM,  unwanted directories and files of `/usr/lib/.build-id/` are included in the RPM, .
```
[root@u1 ~]# rpm -ql gh-ost-1.1.2-1
/usr/bin/gh-ost
/usr/lib/.build-id
/usr/lib/.build-id/c4
/usr/lib/.build-id/c4/dd5acac09d7d035b4a07fdce618a4eab59366c
```
Based on: https://fedoraproject.org/wiki/Releases/FeatureBuildId and https://fedoraproject.org/wiki/Changes/ParallelInstallableDebuginfo
Add `--rpm-rpmbuild-define "_build_id_links none"` to `fpm` should fix the issue

timvaillancourt:
This was resolved in https://github.com/github/gh-ost/pull/954

PR #954, opened 2021-04-06
Remove build_id files from rpm

yakirgb:
Related issue: https://github.com/github/gh-ost/issues/953

### Description

Right now the rpm including unwanted files of `/usr/lib/.build-id/`  for example:
```
[root@u1 ~]# rpm -ql gh-ost-1.1.2-1
/usr/bin/gh-ost
/usr/lib/.build-id
/usr/lib/.build-id/c4
/usr/lib/.build-id/c4/dd5acac09d7d035b4a07fdce618a4eab59366c
```
Based on https://fedoraproject.org/wiki/Changes/ParallelInstallableDebuginfo and https://fedoraproject.org/wiki/Releases/FeatureBuildId , add `--rpm-rpmbuild-define "_build_id_links none"` will fix the issue.

After the fix:
```
[root@u1 ~]# rpm -ql  gh-ost-1.1.2-1
/usr/bin/gh-ost
```

Thank you, Yakir.


shlomi-noach (approved):
LGTM

PR #962, opened 2021-04-29
fix #961

### LINKED ISSUE #961: mysql8.0.2+ -exact-rowcount bug
https://dev.mysql.com/doc/refman/8.0/en/keywords.html
ROWS (R); became reserved in 8.0.2

if we use `-exact-rowcount` option, gh-ost will execute query
```
select /* gh-ost */ count(*) as rows from `db`.`t1`;
```
which will cause
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'rows from `db`.`t1`' at line 1
```

```
select /* gh-ost */ count(*) as `rows` from `db`.`t1`;
```

Fanduzi:
fix https://github.com/github/gh-ost/issues/961

### Description

https://dev.mysql.com/doc/refman/8.0/en/keywords.html
ROWS (R); became reserved in 8.0.2

if we use `-exact-rowcount` option, gh-ost will execute query
```
select /* gh-ost */ count(*) as rows from `db`.`t1`;
```
which will cause
```
You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near 'rows from `db`.`t1`' at line 1
```
Please use `\``
```
select /* gh-ost */ count(*) as `rows` from `db`.`t1`;
```


shlomi-noach (changes requested):
Suggestion: change to an altogether different alias, such as `count_rows`. I find that it's generally not a good idea to qualify reserved words; analysis tools, scripts etc. do not play well with qualified reserved words.

Fanduzi:
> Suggestion: change to an altogether different alias, such as `count_rows`. I find that it's generally not a good idea to qualify reserved words; analysis tools, scripts etc. do not play well with qualified reserved words.

Sorry, actually I don't know golang programming... So I just added a quote. I have now changed the column alias to `count_rows`

shlomi-noach (approved):
LGTM

timvaillancourt:
Merging without replica testing as this is low risk

PR #968, opened 2021-05-07
Use matrix build for replica test CI

timvaillancourt:
### Description

This PR moves the replica testing CI to matrix builds to make them faster

This reduces the CI runtime from roughly 25 -> 7 minutes 🎉 


shlomi-noach:
Woohoo!

PR #969, opened 2021-05-08
fix data-integrity: timestamp convert to datetime scenario

cenkore:
related:  #161 #247 #269 ..

This PR is to fix the data integrity problem in timestamp conversion datetime scenarios.

**Env**

Variable_name  |  Value
:------------  |  ------------
system_time_zone  |  CST
time_zone 	|  +08:00

**Reason**
The timestamp time_zone parsed from binlog is CST +08:00, go-sql-driver will convert it to UTC +00:00 by default, and finally when applying to the database, even though add session query `set @@session.time_zone='+00:00'`, but because the column type has been changed to datetime, the time zone did not do conversion again, so the time zone was actually only converted by half, causing the data written to _xx_gho is 8 hours smaller.

**Solution**
In this case, change the time zone from +00:00 to the local system time zone, which is the opposite of datetime convert to timestamp.

Thank you !


dm-2:
Fixed merge conflict for testing

dm-2 (approved):

timvaillancourt (approved):

timvaillancourt:
@cenkore / @dm-2 / @rashiq: as this converts data I'd like to be very sure it works now and in the future

Are there any tests that we can add to the `localtests/` dir that would:
1. Confirm `master` currently has a problem (test fails before changes in this PR)
2. Confirm the logic in this PR resolves the problem?
🤔

timvaillancourt (changes requested):
@cenkore see my previous comment 🙏

cenkore:
> @cenkore see my previous comment 🙏
@timvaillancourt  Sorry for taking a long time to reply you.   PTAL.

meiji163:
Is this still an issue? `localtests/tz-ts-datetime` test passes on the master branch. Also see https://github.com/github/gh-ost/pull/1322

meiji163:
@cenkore Is this still an issue? `localtests/tz-ts-datetime` test passes on the master branch. Also see https://github.com/github/gh-ost/pull/1322

cenkore:
> @cenkore Is this still an issue? `localtests/tz-ts-datetime` test passes on the master branch. Also see #1322

It looks good. By the way, is this [restriction](https://github.com/github/gh-ost/blob/master/go/logic/inspect.go#L208) still necessary?

PR #986, opened 2021-06-06
Experimental: use 

shlomi-noach:
Based on https://github.com/github/gh-ost/pull/935, we use `go mod` and replace `siddontang/go-mysql` with its successor, https://github.com/go-mysql-org/go-mysql

An attempt to see whether this upgrade solves https://github.com/openark/gh-ost/pull/19


shlomi-noach:
gah. intended to open downstream.

PR #988, opened 2021-06-17
Set the ServerName for TLS configuration

dbussink:
When TLS hostname validation used for the MySQL connection, the ServerName property needs to be set so that it knows which name to validate on the certificate.

Without this option and with InsecureSkipVerify set to false, validation will error here with a fatal error otherwise:

```
FATAL tls: either ServerName or InsecureSkipVerify must be specified in the tls.Config
```

I'm not sure if there's a good way to additional test this specific change, as there's not a lot of extensive tests around the TLS handling and setup methods like `UseTLS` for example. 

I think this change is also necessary to be able to complete #521.

### DIFF
--- a/go/mysql/connection.go
+++ b/go/mysql/connection.go
@@ -92,6 +92,7 @@ func (this *ConnectionConfig) UseTLS(caCertificatePath, clientCertificate, clien
 	}
 
 	this.tlsConfig = &tls.Config{
+		ServerName:         this.Key.Hostname,
 		Certificates:       certs,
 		RootCAs:            rootCertPool,
 		InsecureSkipVerify: allowInsecure,


shlomi-noach (approved):
LGTM

ryansimmen:
@shlomi-noach how can we get the `gh-ost-build-deploy-tarball` test to pass? Why didn't it kick off automatically?

timvaillancourt (approved):

shlomi-noach:
> how can we get the gh-ost-build-deploy-tarball test to pass?

Heh, probably `.ci build gh-ost/tls-set-server-name` if I remember correctly :stuck_out_tongue: 

> Why didn't it kick off automatically?

Dunno? Possibly a PR submitted by an owner kicks that specific build? Could be some new configuration.

shlomi-noach:
Rather:

```
.ci build gh-ost-build-deploy-tarball/tls-set-server-name
```

PR #995, opened 2021-06-25
Add http server for pprof

timvaillancourt:
### Description

In order to understand a performance issue with the inspector, this PR adds an http server to support [pprof profiling](https://golang.org/pkg/net/http/pprof/) of `gh-ost`

This is disabled by default and enabled by passing `--http-listen-addr 10.11.12.13:8080`

Endpoints:
- `GET /_ping` - returns `OK`
- `GET /debug/pprof` - returns [pprof profile data](https://golang.org/pkg/net/http/pprof/)

cc @shlomi-noach for any style thoughts

### DIFF
--- a/go/base/http.go
+++ b/go/base/http.go
@@ -1,0 +1,25 @@
+package base
+
+import (
+	"fmt"
+	"net/http"
+	_ "net/http/pprof"
+
+	"github.com/openark/golib/log"
+)
+
+// HttpConfig contains the configuration of the http server
+type HttpConfig struct {
+	ListenAddr string
+}
+
+// handleHttpPing handles a ping request
+func handleHttpPing(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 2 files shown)


rashiq:
@timvaillancourt don't see the implementation of `GET /debug/pprof` - is that still wip?

timvaillancourt:
@rashiq that endpoint is auto-exported by `net/http/pprof`: https://golang.org/pkg/net/http/pprof/#pkg-overview

rashiq (approved):

rashiq:
@timvaillancourt ah gotcha!

timvaillancourt:
I think I might move the server funcs under `migrationContext` so the server has access to that state in the future. Any thoughts on that @shlomi-noach?

timvaillancourt:
Closing as no longer required

PR #1003, opened 2021-07-09
Convert column to destination charset in DML applications

jbielick:
Related issue: https://github.com/github/gh-ost/issues/290

### Description

Avoid DML apply errors by `convert`ing character data when `charset` changes for a column.

### Background

When applying DML events (UPDATE, DELETE, etc) the values are sprintf'd into a prepared statement with the row snapshot (values). Due to the possibility of migrating text column data containing characters in the source table that are _invalid_ in the destination table (due to charset), a conversion step is often necessary. This conversion does not occur when applying DML events and an error occurs writing invalid byte sequences to the ghost table.

For example, when migrating a table/column from `latin1` to `utf8mb4`, the `latin1` column may contain characters that are invalid single-byte `utf8` characters. Characters in the `\x80-\xFF` range are most common. When written to `utf8mb4` column without conversion, they fail as they do not exist in the `utf8` codepage.

Converting these texts/characters to the destination charset using `convert(? using {charset})` will convert appropriately and the update/replace will succeed.
    
Note: there is currently no issue backfilling the ghost table when the characterset changes, likely because it's a insert-into-select-from and it all occurs within mysql. I only point this out because there are two tests added for this: `latin1text-to-utf8mb4` and `latin1text-to-ut8mb4-insert`—the former is a test that fails prior to this commit. The latter is a test that succeeds prior to this comment. Both are affected by the code in this commit. Please let me know if you would like these renamed or consolidated into `convert-utf8mb4`. They were helpful to do TDD.


localtests/convert-utf8mb4/create.sql
```diff
@@ -7,7 +7,7 @@ create table gh_ost_test (
   primary key(id)
 ) auto_increment=1;
 
-insert into gh_ost_test values (null, 'átesting');
+insert into gh_ost_test values (null, 'átesting', null, null);
```
jbielick, line 10: Fixes `Testing: convert-utf8mb4..ERROR 1048 (23000) at line 10: Column 'tl' cannot be null`

Mismatch didn't seem intentional; I can revert if needed.

localtests/convert-utf8mb4/create.sql
```diff
@@ -7,7 +7,7 @@ create table gh_ost_test (
   primary key(id)
 ) auto_increment=1;
 
-insert into gh_ost_test values (null, 'átesting');
+insert into gh_ost_test values (null, 'átesting', '', '');
```
jbielick, line 10: Fixes `Testing: convert-utf8mb4..ERROR 1136 (21S01) at line 10: Column count doesn't match value count at row 1`

Mismatch didn't seem intentional; I can revert if needed.

localtests/latin1text-to-utf8mb4-insert/create.sql
```diff
@@ -0,0 +1,9 @@
+drop table if exists gh_ost_test;
+create table gh_ost_test (
+  id int auto_increment,
+  t text charset latin1 collate latin1_swedish_ci,
+  primary key(id)
+) auto_increment=1 charset latin1 collate latin1_swedish_ci;
+
+insert into gh_ost_test values (null, char(189));
```
jbielick, line 8: 189 and 190 chosen specifically because they can be replaced with a utf8 character equivalent (prefixed with `\xC2`). Thus the checksum at the end of the test will pass.

shlomi-noach:
@jbielick I'm curious to see how the tests run. Would you care to push same branch (and create similar PR) in https://github.com/openark/gh-ost ? (If not, I can to the same on your behalf; but you're the author so it's best if you do it)

Notwithstanding #290, which I completely missed as I was phasing out, I'm surprised this issue only shows up now.

jbielick:
@shlomi-noach me, too. I got them running locally with mysql 5.7.25 and a couple changes to the Dockerfile and they all passed. Looking forward to a CI run. 

Rebased and opened at:

https://github.com/openark/gh-ost/pull/27

shlomi-noach:
Thank you!

jbielick:
@shlomi-noach cherry-picked commits here.

shlomi-noach:
@timvaillancourt :wave: this looks legit to me. However, it works on data transfer (conversion of character sets) and I'd like to see how this works under production tests. Would you be able to deploy this change?

timvaillancourt:
> @timvaillancourt 👋 this looks legit to me. However, it works on data transfer (conversion of character sets) and I'd like to see how this works under production tests. Would you be able to deploy this change?

@shlomi-noach 👋 absolutely! I can get this on some testing hosts next week 👍

timvaillancourt (approved):

jbielick:
Anything go awry?

jbielick:
@timvaillancourt @shlomi-noach  Hey y'all. :wave: 

I'm wondering if there is anything I can do to help here. Does this still seem like an appropriate change to merge? 

To add some more context, we're in need of this change to successfully convert large tables from `latin1` to `utf8mb4`. Backfilling goes just fine, but updates that are applied encounter errors (as they they do SQL writes/update differently than the backfill) and the whole things dies. Any issues or edge cases I can look into? This is understandably high-risk. 

If I were to build a binary from this branch to test in our prod DB, should I rebase and do so from the openark repo?

timvaillancourt:
> I'm wondering if there is anything I can do to help here. Does this still seem like an appropriate change to merge?

@jbielick / @shlomi-noach: 👋 the production tests of this PR ran into some unrelated test failures that are still under investigation. I think tests of this PR should be ran again to make sure this didn't break anything. cc'ing @dm-2 / @rashiq for help on tests

And just calling out that while the automated tests **will** prove gh-ost still works in general, it won't trigger the charset logic from this PR as the production test runs `ENGINE=InnoDB` only

Lastly, I deleted the `ignore_versions` file for the 5.5 version because we dropped support/tests for 5.5 so no need to ignore it

jbielick:
Thanks for the update. 

> it won't trigger the charset logic from this PR as the production test runs ENGINE=InnoDB only

May I ask what you mean in this statement? I was expecting this to affect InnoDB tables in test and production in the intended way, but perhaps I'm missing or misunderstanding something.

timvaillancourt:
> May I ask what you mean in this statement? I was expecting this to affect InnoDB tables in test and production in the intended way, but perhaps I'm missing or misunderstanding something.

@jbielick what I meant there is the tests on Production replicas at GitHub use `--alter="ENGINE=InnoDB"` as an ALTER statement, meaning this logic won't be triggered in that testing

That's why I suggest test scenarios are added to `localtests/` because these tests CAN run `ALTER`s that trigger the logic in this PR _(see `create.sql` and `extra_args` files for example)_

jbielick:
> > May I ask what you mean in this statement? I was expecting this to affect InnoDB tables in test and production in the intended way, but perhaps I'm missing or misunderstanding something.
> 
> @jbielick what I meant there is the tests on Production replicas at GitHub use `--alter="ENGINE=InnoDB"` as an `ALTER` statement, meaning this logic won't be triggered in that testing
> 
> That's why I suggest test scenarios are added to `localtests/` because these tests CAN run `ALTER`s that trigger the logic in this PR _(see `create.sql` and `extra_args` files for example)_

@timvaillancourt makes sense. I've got some tests added here in `localtests/` that reproduce the failure and assert success. Do those tests look sufficient or am I missing something that I can add here?

timvaillancourt:
@jbielick in light of https://github.com/github/gh-ost/pull/1158 being merged I just wanted to check if this is still needed? cc @wangzihuacool who can add some insight also

Thanks all 🙇

go/logic/inspect.go
```diff
@@ -190,7 +190,7 @@ func (this *Inspector) inspectOriginalAndGhostTables() (err error) {
 			this.migrationContext.MappedSharedColumns.SetEnumValues(column.Name, column.EnumValues)
 		}
 		if column.Name == mappedColumn.Name && column.Charset != mappedColumn.Charset {
-			this.migrationContext.SharedColumns.SetCharsetConversion(column.Name, column.Charset, mappedColumn.Charset)
+			this.migrationContext.MappedSharedColumns.SetCharsetConversion(column.Name, column.Charset, mappedColumn.Charset)
```
timvaillancourt, line 193: @jbielick curious if you can elaborate on the switch from `SharedColumns` to `MappedSharedColumns`. Thanks!

jbielick: Well, chronologically I believe I added this line before #1158 was opened. `SharedColumns` might be the correct usage for #1158 because it leverages `convertArg` and does a conversion of the value in go. I preferred the approach that was similar to the `DatetimetoTimestamp` and `EnumToTextConversion` approaches where the conversion occurs within mysql, as would normally occur if I altered the table without gh-ost.

So it's not that I changed this line, but it now appears in the diff as a change because the base changed and #1158 _also_ introduced this line with a slightly different goal.

I'm using `MappedSharedColumns` because those are the set that allow replacing the "SET x=x" arg with a mysql expression instead of manipulating a raw value itself (which is possible with SharedColumns). We don't need to modify the bytes that are being sent, we need mysql to interpret them differently (and as proposed here convert them).

Issue #1000, opened 2021-06-29
unique key columns is not a subset of shared columns in BuildDMLUpdateQuery on JSON table 


ykalfon:
Hi

we run gh-ost on MySQL 8.0.22 and got the following error when an UPDATE statement is performed on the table while ghost is running:
**ERROR unique key columns is not a subset of shared columns in BuildDMLUpdateQuery**

This is the scenario:

```
CREATE TABLE `test` (
  `main_id` varbinary(16) NOT NULL,
  `sub_id` varchar(36) CHARACTER SET latin1 COLLATE latin1_swedish_ci GENERATED ALWAYS AS (jsonbody->>'$._id') STORED NOT NULL,
  `jsonbody` json NOT NULL,
  PRIMARY KEY (`main_id`,`sub_id`)
) ENGINE=InnoDB DEFAULT CHARSET=utf8mb4 COLLATE=utf8mb4_bin ROW_FORMAT=COMPRESSED
```

```
INSERT INTO test (main_id, jsonbody) 
VALUES (x'F2C45C49175845FCB062D579DFB6CE1D','{"_id":"c1192a0c-b1b8-4b86-adb5-19f40a869f3f","name":"apple"}');
```
while gh-ost is running, run any update on the table, for example:

```
UPDATE test
  SET jsonbody=JSON_OBJECT('_id', 'c1192a0c-b1b8-4b86-adb5-19f40a869f3f', 'name', 'carrot')
WHERE main_id=x'F2C45C49175845FCB062D579DFB6CE1D'
  AND sub_id='c1192a0c-b1b8-4b86-adb5-19f40a869f3f';
```

and gh-ost throws: 
**ERROR unique key columns is not a subset of shared columns in BuildDMLUpdateQuery**

The gh-st command that was run is:

```
/usr/bin/gh-ost  
--max-load=Threads_running=500  
--critical-load=Threads_running=1000  
--initially-drop-ghost-table  
--initially-drop-old-table  
--chunk-size=600  
--max-lag-millis=3000  
--user=XXXX --password=XXXXX 
--assume-master-host=rw_master_1 
--throttle-control-replicas=ro_co-master_2 
--host=ro_2 
--database=test --table=test 
--verbose  
--alter="modify sub_id VARCHAR(36) CHARACTER SET latin1 COLLATE latin1_general_ci GENERATED ALWAYS AS (jsonbody->>'$._id') STORED NOT NULL"  
--assume-rbr  
--allow-master-master  
[...]

PR #1004, opened 2021-07-12
GENERATED column as part of PK: handling UPDATE scenario

shlomi-noach:
Resubmission of https://github.com/openark/gh-ost/pull/26 from downstream, this PR further supports `GENERATED` columns as part of the `PRIMARY KEY`. It was previously impossible to update rows based on a PK `GENERATED` value; now it is.

Related: #1000


meiji163:
I didn't realize this existed before I fixed the same thing in https://github.com/github/gh-ost/pull/1461, doh!

Issue #1030, opened 2021-09-28
If MySQL two-phase commit is time-consuming, it will cause gh-ost to lose data.


shaohk:
> We encountered a case of the gh-ost lose data. The details of this case is as follows:
1. start the gh-ost to alter online table t1.

2. when the gh-ost is finish, we find there is lose a data in new table t1, and the lose data is in the old table _t1_del.

3. we analyze the binlog and gh-ost log is as follows
 * lose data id is 17, binlog position is (mysql-bin.000009, 98240).
 * the gh-ost begin to sync binlog from position (mysql-bin.000009, 1780).
 * the gh-ost migration min and max value is (1, 16).
 * the gh-ost first status log is 
Copy: 0/16 0.0%; Applied: 0; Backlog: 0/1000; Time: 33s(total), 0s(copy); streamer: mysql-bin.000009:101907; Lag: 15.18s, HeartbeatLag: 15.08s, State: migrating; ETA: N/A

We suspect the case is because of MySQL two-phase commit.

The gh-ost command is
```
gh-ost \
--max-load=Threads_running=32 \
--critical-load=Threads_running=48 \
--chunk-size=500 \
--max-lag-millis=1000000 \
--initially-drop-old-table \
--initially-drop-socket-file \
--user="xx" \
--password="xxxx" \
--host="xxx.xxx.xxx.xxx" \
--port=xxxxx \
--database="shaohk" \
--table="t1" \
--alter="ADD INDEX idx_t1 (name);" \
--switch-to-rbr=false \
--cut-over=default \
--assume-rbr=true \
--exact-rowcount=false \
--default-retries=3 \
--critical-load-interval-millis=300000 \
--serve-socket-file=/tmp/ghost.socket.t1.sock \
--panic-flag-file=/tmp/ghost.panic.flag.shaohk.t1 \
--throttle-flag-file=/tmp/ghost.throttle.flag.shaohk.t1 \
--force-table-names=t1 \
--verbose \
--execute \
--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag.shaohk.t1 \
--allow-on-master
```

The insert transaction is 
```
mysql> begin;
Query OK, 0 rows affected (0.03 sec)

mysql> insert into t1 values (17,'test');
Query OK, 1 row affected (0.04 sec)

mysql> commit;
Query OK, 0 rows affected (30.3 sec)
```

The gh-ost log is
```
2021-09-26 23:11:50 INFO starting gh-ost
[...]

PR #1040, opened 2021-10-26
fix #1039 #1030: add shared-mode lock in query ReadMigrationRangeValues to avoid losing data

wangzihuacool:
related #1039 #1030 ..
This PR is to avoid losing data when gh-ost in MySQL with semi-sync replication enabled.

**Reason**
In mysql with a semi-sync slave, transactions may be in uncommitted state with 'Waiting for semi-sync ACK from slave' wait,   though its related binlog event has already been handled and discarded by gh-ost BinlogStreamer. If gh-ost ReadMigrationRangeValues executes before the transactions finnally committed, the uncommitted rows is lost.

**Fix**
Add a shared mode lock in the query of ReadMigrationRangeValues. If min/max value of the uniquekey was changed by another transaction that has not yet committed, query just waits until that transaction ends and then uses the latest values.
Thus ReadMigrationRangeValues returns the accurate value.

Thank you !

PR #1054, opened 2021-11-15
fix: lost data due to mysql tow-phase-commit

shaohk:
fix: lost data due to mysql tow-phase-commit

Related issue: https://github.com/github/gh-ost/issues/1030

### Description

This PR [fix lost data due to mysql tow-phase-commit]


Fanduzi:
Very clever!

timvaillancourt:
@wangzihuacool are you able to test if this change fixes the issue?

I think this approach is a bit cleaner than using a lock in https://github.com/github/gh-ost/pull/1040, assuming it all works

wangzihuacool:
Both using a shared lock in #1040 and triggering a WriteChangelog event can fix the issue. I have tested.
I think using a shared lock is a direct and effective way  to fix it, with better compatibility, just like the way rowcopy does. By contract triggering a writechangelog event seems to be a more clever way, but a bit confusing, with a small side effect that gh-ost migration may exits abnormally if WriteChangelog fails or timeout for some reason.

shaohk:
> Both using a shared lock in #1040 and triggering a WriteChangelog event can fix the issue. I have tested. I think using a shared lock is a direct and effective way to fix it, with better compatibility, just like the way rowcopy does. By contract triggering a writechangelog event seems to be a more clever way, but a bit confusing, with a small side effect that gh-ost migration may exits abnormally if WriteChangelog fails or timeout for some reason.

Yes, using share lock is also a very effective way. But we are worried about an unexpected problem caused by share lock, so I added a write heartbeat here. Even if the gh-ost fails due to the heartbeat write timeout, this is a small probability time and acceptable.

Fanduzi:
To fix or not to fix. How long has this been going on?

Fanduzi:
Any progress?

Fanduzi:
Hi @dm-2 @timvaillancourt 
https://github.com/github/gh-ost/pull/1040 and https://github.com/github/gh-ost/pull/1054  has had a PR for fixing the semi-sync data loss bug for some time now, is there any plans to merge?

timvaillancourt (approved):
LGTM. Updating the heartbeat is an extremely safe operation with very low risk 👍 

cc @dm-2 / @rashiq for review/merge

go/logic/migrator.go
```diff
@@ -388,6 +388,12 @@ func (this *Migrator) Migrate() (err error) {
 	if err := this.addDMLEventsListener(); err != nil {
 		return err
 	}
+
+	// write a heartbeat data into the ghc table, and avoid lost data due to MySQL two-phase-commit.
```
timvaillancourt, line 392: @shaohk could you add a bit more background in this comment on "why" writing the heartbeat causes us to `avoid lost data due to MySQL two-phase-commit`? That will be useful to those without context

Fanduzi: I'll try to describe the scene:

When using semi-sync and setting `rpl_semi_sync_master_wait_point=AFTER_SYNC`, if an INSERT statement is being committed but blocks due to an unmet ack count, the data inserted by the transaction is not visible to `ReadMigrationRangeValues`, so the copy of the existing data in the table does not include the new row inserted by the transaction. However, the binlog event for the transaction is already written to the binlog, so the `addDMLEventsListener` only captures the binlog event after the transaction, and thus the transaction's binlog event is not captured, resulting in data loss. 

If you add a heartbeat, and the transaction commit blocks because the ack is not met, then the heartbeat will not be able to write, so the ReadMigrationRangeValues will not be run. When the heartbeat writes successfully, the `ReadMigrationRangeValues` will read the newly inserted data, thus Avoiding data loss due to the above problem

timvaillancourt: @Fanduzi that's my understanding too 👍

I'm hoping we can add some of this context into the code-comment on line 392, so that this is easier to read for those without context

timvaillancourt: @shaohk two more questions:

1. Would it make sense to move this to be within the [`.ReadMigrationRangeValues()`](https://github.com/github/gh-ost/blob/02a26a584dadbe685837ff7d4620134a44196a13/go/logic/applier.go#L424-L433) func?
2. What do you think about updating a [`ChangeLogState`](https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L24-L29) instead of the heartbeat? Perhaps a new `ChangeLogState` like `ReadMigrationRangeValues` would work here 🤔

shaohk: > @shaohk two more questions:
> 
> 1. Would it make sense to move this to be within the [`.ReadMigrationRangeValues()`](https://github.com/github/gh-ost/blob/02a26a584dadbe685837ff7d4620134a44196a13/go/logic/applier.go#L424-L433) func? The goal with the `.Migrate()` func is to leave a lot of the details to other functions
> 2. What do you think about updating a [`ChangeLogState`](https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L24-L29) instead of the heartbeat? [`.WriteChangelogState()` in the applier](https://github.com/github/gh-ost/blob/master/go/logic/applier.go#L323-L325) is used for this. Perhaps a new `ChangeLogState` like `ReadMigrationRangeValues` would work here 🤔. [Example PR](https://github.com/github/gh-ost/pull/1134)

Sorry, I just saw these messages.
1. I complete the annotation information in the code.
2. I think the above two suggestions are very good, I will revise the PR and mention it again.

shaohk: > @shaohk two more questions:
> 
> 1. Would it make sense to move this to be within the [`.ReadMigrationRangeValues()`](https://github.com/github/gh-ost/blob/02a26a584dadbe685837ff7d4620134a44196a13/go/logic/applier.go#L424-L433) func? The goal with the `.Migrate()` func is to leave a lot of the details to other functions
> 2. What do you think about updating a [`ChangeLogState`](https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L24-L29) instead of the heartbeat? [`.WriteChangelogState()` in the applier](https://github.com/github/gh-ost/blob/master/go/logic/applier.go#L323-L325) is used for this. Perhaps a new `ChangeLogState` like `ReadMigrationRangeValues` would work here 🤔. [Example PR](https://github.com/github/gh-ost/pull/1134)

 [Use changelog state instead of the heartbeat](https://github.com/github/gh-ost/pull/1141)

shaohk: > I'll try to describe the scene:
> 
> When using semi-sync and setting `rpl_semi_sync_master_wait_point=AFTER_SYNC`, if an INSERT statement is being committed but blocks due to an unmet ack count, the data inserted by the transaction is not visible to `ReadMigrationRangeValues`, so the copy of the existing data in the table does not include the new row inserted by the transaction. However, the binlog event for the transaction is already written to the binlog, so the `addDMLEventsListener` only captures the binlog event after the transaction, and thus the transaction's binlog event is not captured, resulting in data loss.
> 
> If you add a heartbeat, and the transaction commit blocks because the ack is not met, then the heartbeat will not be able to write, so the ReadMigrationRangeValues will not be run. When the heartbeat writes successfully, the `ReadMigrationRangeValues` will read the newly inserted data, thus Avoiding data loss due to the above problem

Thanks, I added your description to the code comments.  https://github.com/github/gh-ost/pull/1141

timvaillancourt:
Closing in favour of https://github.com/github/gh-ost/pull/1141 👍

Issue #1082, opened 2022-01-28
Allow customizing gh-ost-on-status hook interval


tknodell-recurly:
We'd like to customize how often the `gh-ost-on-status` hook is called
Right now it looks like its [hardcoded to 60 seconds](https://github.com/github/gh-ost/blob/d6c83638d70dbfdad9e076ee61f8bbf062e09129/go/logic/migrator.go#L1019-L1021)

We'd like to make this a flag, so we can call our script to post status to slack at longer intervals
Happy to open a PR for this

dm-2:
PR merged 🚀

PR #1083, opened 2022-02-01
Add flag to customize the interval which the onStatus hook is called

tknodell-recurly:

Related issue: https://github.com/github/gh-ost/issues/1082

### Description

This PR adds a new flag `hooks-status-interval` which defaults to 60 seconds
Allows customization of how often the onStatus hook fires


dm-2 (approved):
Thanks @tknodell-recurly, this looks good - we'll include this in the next release 👍

PR #1087, opened 2022-02-07
Reduce the minimal chunk size from `100` to `10`.

arthurschreiber:

We'd like to set the minimum chunk size to a value lower than `100`. Copying `100` rows at a time is fine for tables that have a low average row length, but copying `100` rows at a time from tables that have a large average row lengths is causing replication lag in our clusters.

### Description

This PR lowers the minimum `chunk-size` value from `100` to `10`.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -551,8 +551,8 @@ func (this *MigrationContext) SetMaxLagMillisecondsThrottleThreshold(maxLagMilli
 }
 
 func (this *MigrationContext) SetChunkSize(chunkSize int64) {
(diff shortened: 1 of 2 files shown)


christianlang (approved):

octatone (approved):

dm-2 (changes requested):
Just one small change - please can you also update the ["allowed range" values in the help text](https://github.com/github/gh-ost/blob/master/go/cmd/gh-ost/main.go#L102) in this PR? Thanks!

dm-2 (approved):

Issue #1105, opened 2022-03-09
gh-ost doesn't have binaries for M1 mac


jecepeda:
Hi everyone,

I went to releases in order to download the binary for gh-ost and I didn't found any binary for darwin arm64 architecture.

Then, I tried to download this repository to build it myself, but it seems that the compiling process downloads a supported go version if the one we have in our machine is not supported, finding that it always downloads the go binary for amd64 (for linux and osx).

Is it planned to give support for arm64 binaries (either on linux/osx)? I'm open to help on this issue and put a PR if you like.

Thanks in advance

timvaillancourt:
@jecepeda I don't think there were plans to add this but I'm not opposed considering Mac OSX is already supported

Any thoughts @dm-2 / @rashiq?

dm-2:
I'm happy with adding `arm64` support, a PR would be very helpful 🙂

We use [`build.sh`](https://github.com/github/gh-ost/blob/master/build.sh) for generating `gh-ost` releases, it should be fairly straightforward to add `arm64` to this.

jecepeda:
Perfect!

dm-2:
I've added a macOS `arm64` binary to the [v1.1.4 release](https://github.com/github/gh-ost/releases/tag/v1.1.4), thanks for your contribution @jecepeda - much appreciated!

I'll close this as completed, but feel free to re-open if needed 🙂

PR #1108, opened 2022-03-11
Add binaries for arm64 architectures

jecepeda:

Related issue: https://github.com/github/gh-ost/issues/1105

### Description

Now that developers are using different architectures (amd64 and arm64, basically),  we need to use different binaries adapted for those architectures.

This PR changes `build.sh` script, adding binaries for linux/arm64 and darwin/arm64

I needed to rename the .tar.gz files to indicate the architecture. Now its naming is:

`./gh-ost-binary-${osshort}-${GOARCH}-${timestamp}.tar.gz`


dm-2 (approved):
Brilliant, thanks @jecepeda - looks good to me 👍

Issue #1119, opened 2022-04-18
A data-loss case occurs when chunk-key contains decimal-column  


wangzhanbing:
Hello:

We tried to recycle holes by `alter table tab engine = innodb` from a table. 
After gh-ost executed, no error coccurs, but data is lost:
```
gh-ost version is 1.1.2
mysql version is 5.6
```

gh-ost cmd is:
```
./gh-ost  \
--host=${IP} --port=${Port} --user=${user} --password=${password} --database=${db}--table=NewInstance \
--alter=engine=innodb \
--switch-to-rbr \
--max-load=Threads_running=25 \
--critical-load=Threads_running=1000 \
--chunk-size=1000 \
--max-lag-millis=1000 \
--heartbeat-interval-millis=2000 \
--throttle-control-replicas=${IP_list} \
--cut-over=default --exact-rowcount \
--concurrent-rowcount \
--default-retries=120 \
--dml-batch-size=10 \
--nice-ratio=0 \
--serve-socket-file=./tmp/gh-ost.sock \
--throttle-flag-file=./tmp/gh-ost.pause.flag \
--ok-to-drop-table \
--cut-over-exponential-backoff=true \
--exponential-backoff-max-interval=1024 \
--panic-flag-file=./tmp/gh-ost.panic.flag \
--postpone-cut-over-flag-file=flag \
--verbose \
--execute \
```

the create- table info is:
```
CREATE TABLE `NewInstance` (
  `InstanceID` decimal(32,0) NOT NULL DEFAULT '0' ,
  `InstanceStatus` tinyint(4) NOT NULL,
  `TriggerTime` datetime NOT NULL,
  `TriggerDate` date NOT NULL,
  `BeginTime` datetime DEFAULT NULL,
  `EndTime` datetime DEFAULT NULL,
  `LastRetryTime` datetime DEFAULT NULL,
  `Retry` int(11) DEFAULT NULL,
  `Parameters` text,
  `RunParameters` text,
  `Status` tinyint(4) NOT NULL,
  `TaskID` bigint(20) NOT NULL,
  `job_id` varchar(100) NOT NULL DEFAULT '0',
  PRIMARY KEY (`InstanceID`),
  KEY `NewInstace_TaskID` (`TaskID`),
  KEY `idx_InstanceStatus_Status` (`InstanceStatus`,`Status`),
  KEY `idx_job_id` (`job_id`),
  KEY `idx_TriggerTime` (`TriggerTime`)
[...]

PR #1120, opened 2022-04-19
A data-loss case occurs when chunk-key contains decimal-column #1119

wangzhanbing:
## Issue

Related issue: https://github.com/github/gh-ost/issues/1119

That's is an issue, describing a data-loss case caused by decimal-type column in chunk-key.
We also describe how to reproduce the case and how to fix it.

### Description

This PR solves the bug by adding `cast` to decimal-type-column in values-stmt\set-stmt\where-stmt 
for example:  If type of column (`InstanceID` ) is `decimal(30, 0)` , the corresponding set-stmt and where-stmt as follows:
```
# where stmt in insert-select/select/delete sql
# ---- before commit----
where (((`InstanceID` > _binary'2708201202204012300000012')) and ((`InstanceID` < _binary'2708201202204062200000015') or ((`InstanceID` = _binary'2708201202204062200000015'))))

# -- after commit ----
where (((`InstanceID` > cast(_binary'2708201202204012300000012' as decimal(30, 0)))) and ((`InstanceID` < cast(_binary'2708201202204062200000015' as decimal(30, 0))) or ((`InstanceID` = cast(_binary'2708201202204062200000015' as decimal(30, 0))))))

# set stmt in update sql
# ---- before commit ----
set `InstanceID`=_binary'2708201202204062200000015'

# -- after commit ----
set `InstanceID`=cast(_binary'2708201202204062200000015' as decimal(30, 0))

# values stmt in replace-sql
# ---- before commit ----
values(_binary'2708201202204062200000015')

# -- after commit ----
values(cast(_binary'2708201202204062200000015' as decimal(30, 0)))
```
 

- [ ok ] contributed code is using same conventions as original code
- [ ok ] `script/cibuild` returns with no formatting errors, build errors or unit test errors.

### DIFF
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -573,6 +573,7 @@ func (this *Inspector) applyColumnTypes(databaseName, tableName string, columnsL
 				continue
 			}
 
+			column.TypeDesc = columnType
 			if strings.Contains(columnType, "unsigned") {
 				column.IsUnsigned = true
 			}
@@ -601,6 +602,9 @@ func (this *Inspector) applyColumnTypes(databaseName, tableName string, columnsL
 			}
 			if charset := m.GetString("CHARACTER_SET_NAME"); charset != "" {
 				column.Charset = charset
+			}
+			if strings.HasPrefix(columnType, "decimal") {
+				column.Type = sql.DecimalColumnType
 			}
 		}
 		return nil
--- a/go/sql/builder.go
+++ b/go/sql/builder.go
@@ -42,6 +42,8 @@ func buildColumnsPreparedValues(columns *ColumnList) []string {
 			token = fmt.Sprintf("ELT(?, %s)", column.EnumValues)
 		} else if column.Type == JSONColumnType {
 			token = "convert(? using utf8mb4)"
+		} else if column.Type == DecimalColumnType {
+			token = fmt.Sprintf("cast(? as %s)", column.TypeDesc)
 		} else {
 			token = "?"
 		}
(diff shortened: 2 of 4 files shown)


timvaillancourt:
@wangzhanbing thanks for this PR! I think this fix makes sense 👍 

Is it possible to add a "localtest" for this scenario under the [`localtests/`](https://github.com/github/gh-ost/tree/master/localtests) subdirectory?

wangzhanbing:
> @wangzhanbing thanks for this PR! I think this fix makes sense 👍
> 
> Is it possible to add a "localtest" for this scenario under the [`localtests/`](https://github.com/github/gh-ost/tree/master/localtests) subdirectory?

OK， That's what I am doing

wangzhanbing:
one another question: 
do we meet some trouble on continuing to support for old version （5.5 5.6） ?

In my company, most of mysql-instances will still be 5.6 in next one or two years. maybe so do other companies.

Is there another more elegant solution than abandoning history supports ?

wangzhanbing:
Can someone help me review the code? @rashiq  @dm-2

dm-2:
> do we meet some trouble on continuing to support for old version （5.5 5.6） ?
> 
> In my company, most of mysql-instances will still be 5.6 in next one or two years. maybe so do other companies.
> 
> Is there another more elegant solution than abandoning history supports ?

We have decided to only develop and test against [currently supported versions](https://endoflife.software/applications/databases/mysql) of MySQL as we're unable to support old versions indefinitely. We'll continue to accept contributed fixes for MySQL 5.6 for the time being, but will only test that these fixes work against 5.7 and 8.0. New features will only be tested against 5.7 and 8.0.

In a future release, we will be dropping support for MySQL 5.5 entirely (including code specific to 5.5).

> Can someone help me review the code? @rashiq @dm-2

I will take a look this week :+1:

dm-2:
@wangzhanbing please could you take a look at the merge conflict? Thank you! 🙇

Issue #1123, opened 2022-05-15
optimize migrator's executeWriteFuncs


debug-LiXiwen:
- suggest an enhancement

- Migrator's executeWriteFuncs will have a infinite loop when applyEventsQueue and copyRowsQueue is empty.
- So I Optimize the priority logic to avoid the infinite loop

enhancement

PR #1124, opened 2022-05-15
optimize migrator's executeWriteFuncs

debug-LiXiwen:

- issue: [1123](https://github.com/github/gh-ost/issues/1123)

### Description
- Migrator's executeWriteFuncs will have a infinite loop when applyEventsQueue and copyRowsQueue is empty.
- So I Optimize the priority logic to avoid the infinite loop


go/logic/migrator.go
```diff
@@ -1279,30 +1279,31 @@ func (this *Migrator) executeWriteFuncs() error {
...
+			{
+				copyRowsStartTime := time.Now()
+				// Retries are handled within the copyRowsFunc
+				if err := copyRowsFunc(); err != nil {
+					return this.migrationContext.Log.Errore(err)
+				}
+				if niceRatio := this.migrationContext.GetNiceRatio(); niceRatio > 0 {
+					copyRowsDuration := time.Since(copyRowsStartTime)
+					sleepTimeNanosecondFloat64 := niceRatio * float64(copyRowsDuration.Nanoseconds())
+					sleepTime := time.Duration(time.Duration(int64(sleepTimeNanosecondFloat64)) * time.Nanosecond)
```
timvaillancourt, line 1305:
```suggestion
					sleepTime := time.Duration(time.Duration(sleepTimeNanosecondFloat64) * time.Nanosecond)
```
The `int64()` conversion is unnecessary and is making the linter upset:
```
unnecessary conversion (unconvert)
```

go/logic/migrator.go
```diff
@@ -1279,30 +1279,31 @@ func (this *Migrator) executeWriteFuncs() error {
...
 						}
 					}
 				default:
-					{
-						// Hmmmmm... nothing in the queue; no events, but also no row copy.
-						// This is possible upon load. Let's just sleep it over.
-						this.migrationContext.Log.Debugf("Getting nothing in the write queue. Sleeping...")
-						time.Sleep(time.Second)
-					}
+					break priority
```
timvaillancourt, line 1293: @debug-LiXiwen I am new to using labeled-`break`/`continue`. Sorry for the noob question

I assume `break priority` causes the program to jump-back to `priority:` [on line 1283](https://github.com/github/gh-ost/pull/1124/files#diff-bd6a81585ccffb0f0feef9d5b3d1c542432f90c29a366a7bfb1e96531d36cbbaR1283). Will [the `for` loop on line 1284](https://github.com/github/gh-ost/pull/1124/files#diff-bd6a81585ccffb0f0feef9d5b3d1c542432f90c29a366a7bfb1e96531d36cbbaR1284) on execute on the 2nd+ iteration?

I believe we want to empty `this.applyEventsQueue` once and continue to row copy, so I want to confirm

go/logic/migrator.go
```diff
@@ -1279,30 +1279,31 @@ func (this *Migrator) executeWriteFuncs() error {
...
-						// Retries are handled within the copyRowsFunc
-						if err := copyRowsFunc(); err != nil {
-							return this.migrationContext.Log.Errore(err)
-						}
-						if niceRatio := this.migrationContext.GetNiceRatio(); niceRatio > 0 {
-							copyRowsDuration := time.Since(copyRowsStartTime)
-							sleepTimeNanosecondFloat64 := niceRatio * float64(copyRowsDuration.Nanoseconds())
-							sleepTime := time.Duration(int64(sleepTimeNanosecondFloat64)) * time.Nanosecond
-							time.Sleep(sleepTime)
+						if err := this.onApplyEventStruct(eventStruct); err != nil {
```
timvaillancourt, line 1288:
```suggestion
						this.throttler.throttle(nil)
						if err := this.onApplyEventStruct(eventStruct); err != nil {
```
I believe we should check the throttler before each `this.onApplyEventStruct(eventStruct)`

timvaillancourt (changes requested):
This is an interesting idea. I left a few review comments, thanks @debug-LiXiwen! 🙏

go/logic/migrator.go
```diff
@@ -1279,30 +1279,31 @@ func (this *Migrator) executeWriteFuncs() error {
 					return err
 				}
 			}
-		default:
-			{
+		case copyRowsFunc := <-this.copyRowsQueue:
+		priority:
```
timvaillancourt, line 1283: Would moving [line 1283 to 1295](https://github.com/github/gh-ost/pull/1124/files#diff-bd6a81585ccffb0f0feef9d5b3d1c542432f90c29a366a7bfb1e96531d36cbbaR1283-R1295) to a new function such as `migrator.drainApplierEventQueue()` remove the need for `priority:` label + `break priority`?

I wonder if this is easier to read (if I understand correctly)

go/logic/migrator.go
debug-LiXiwen: It’s valid to terminate outer loops
labeled-break will jump labeled code and  move to the next clause

go/logic/migrator.go
debug-LiXiwen: yeah, this is a good idea

debug-LiXiwen: the outer time.Duration is unnecessary too

go/logic/migrator.go
debug-LiXiwen: yeah, your suggestion is right

timvaillancourt:
@debug-LiXiwen thanks for the PR changes, it is passing CI now 👍 

Now that the code looks good, I'm curious if you could explain what situations this optimizes?

If I understand correctly, this is the scenario this optimizes:
1. [`case  eventStruct <:= - this.applyEventsQueue()`](https://github.com/github/gh-ost/blob/b209349e7fdf16d2823e3ef02c3fbef1fd51598a/go/logic/migrator.go#L1276) is empty
2. Immediately after, events are added to `this.applyEventsQueue()` that were not seen by the last step
3. The `select` moves on to `case copyRowsFunc := <-this.copyRowsQueue()`, which is **NOT** empty
4. The new func `this.drainApplierEventQueue()` drains the events that appeared at step #2
    - On `master` today, the `for` loop would need to iterate again to catch the new applier events

Did I get this right and are the more scenarios? 🙏

go/logic/migrator.go
```diff
@@ -1279,36 +1279,43 @@ func (this *Migrator) executeWriteFuncs() error {
...
 	}
 }
 
+func (this *Migrator) drainApplierEventQueue() error {
+	for {
+		this.throttler.throttle(nil)
+		select {
+		case eventStruct := <-this.applyEventsQueue:
+			{
+				if err := this.onApplyEventStruct(eventStruct); err != nil {
```
timvaillancourt, line 1309:
```suggestion
	for {
		select {
		case eventStruct := <-this.applyEventsQueue:
			{
				this.throttler.throttle(nil)
				if err := this.onApplyEventStruct(eventStruct); err != nil {
```
Let's only call the throttler if there is an event in the queue

timvaillancourt (changes requested):
One more requested change I noticed now

debug-LiXiwen:
> 

1. When the copyRowsQueue is not ready, you don't need to sleep manually, the select of the go language will be handled automatically
2.  When the copyRowsQueue is ready, logically analyze the new way of processing to achieve better prioritization, but I did not verify the effect in the actual scene

go/logic/migrator.go
debug-LiXiwen: The call way I am reference [Line 1271](https://github.com/github/gh-ost/blob/b209349e7fdf16d2823e3ef02c3fbef1fd51598a/go/logic/migrator.go#L1271), the throttler is called before there is an event in the queue

timvaillancourt: @debug-LiXiwen yes, but the `for` loop in `.drainApplierEventQueue()` could potentially write many events

Currently the throttler is checked between every apply event, so I believe `.drainApplierEventQueue()` should do the same on each iteration of the `for` loop on [line 1304](https://github.com/github/gh-ost/blob/b209349e7fdf16d2823e3ef02c3fbef1fd51598a/go/logic/migrator.go#L1304)

[Your last commit](https://github.com/github/gh-ost/commit/b209349e7fdf16d2823e3ef02c3fbef1fd51598a) is already doing this [on line 1305](https://github.com/github/gh-ost/blob/b209349e7fdf16d2823e3ef02c3fbef1fd51598a/go/logic/migrator.go#L1305) but the throttler check is ran before we know if there is anything in the applier queue. I believe the throttler check should occur [on line 1309](https://github.com/github/gh-ost/blob/b209349e7fdf16d2823e3ef02c3fbef1fd51598a/go/logic/migrator.go#L1309) instead - only when there is events to apply

timvaillancourt:
> 1. you don't need to sleep manually, the select of the go language will be handled automatically

That's a good point, the `time.Sleep()` is a bit inefficient. Thanks for explaining 👍

go/logic/migrator.go
debug-LiXiwen: oh, you're right,I will modify it

dm-2:
Thanks for your contribution @debug-LiXiwen! 🙇

> 1. When the copyRowsQueue is not ready, you don't need to sleep manually, the select of the go language will be handled automatically

I think relying on `select` to wait instead of using `time.Sleep` is a good idea as it makes the code easier to read. I don't think the performance difference will be significant, but I like that it makes the code flow better.

> 2. When the copyRowsQueue is ready, logically analyze the new way of processing to achieve better prioritization, but I did not verify the effect in the actual scene

I think `drainApplierEventQueue()` is unnecessary and we should remove it. I understand the reason for this optimisation, but the benefit will be very small and it makes the code more complicated than necessary. I prefer that we keep the code as simple as possible, and only optimise when there's a demonstrated need.

timvaillancourt:
> I don't think the performance difference will be significant, but I like that it makes the code flow better.

Agreed, I think it's very unlikely this would improve performance for most workloads

> I think drainApplierEventQueue() is unnecessary and we should remove it

I would be on board with this 👍. It's is incredibly unlikely there are new events to apply immediately after the `select` already checked the apply queue

Issue #713, opened 2019-02-14
wildcards in grants are not supported


babinomec:
We have a lot of databases which are named like "something_data_<client>". If I give these grants to ghost user :

```GRANT ALL PRIVILEGES ON `something\_data\_%`.* TO 'ghost'@'%'```

It fails with :
```ERROR User has insufficient privileges for migration. Needed: SUPER|REPLICATION CLIENT, REPLICATION SLAVE and ALL on `something_data_<client>`.*```

If I connect with mysql client, the user can see the databases and make changes.

It's most probably in the way the permissions are checked in `gh-ost/go/logic/inspect.go` `validateGrants()`. I can't think of a non-ugly solution to check that though...

mattmilesi:
@babinomec have you found a solution?

We use RDS (Aurora) and have the same issue.
It fails with:
``FATAL 2021-08-06 12:57:23 ERROR User has insufficient privileges for migration. Needed: SUPER|REPLICATION CLIENT, REPLICATION SLAVE and ALL on `mydb`.*.``

babinomec:
As far as I recall, no.

We ended up needing gh-ost on lots of different databases/tables so I think we gave rights on everything and limited on the host part.

PR #1125, opened 2022-05-18
Allow wildcards in grants

Gonlo2:
Related issue: https://github.com/github/gh-ost/issues/713

### Description

This PR adds support for using wildcards in grants by converting the mysql pattern into a regex and checking if it matches the named database.

### DIFF
--- a/go/logic/inspect.go
+++ b/go/logic/inspect.go
@@ -23,6 +23,35 @@ import (
 
 const startSlavePostWaitMilliseconds = 500 * time.Millisecond
 
+var grantOnRegexp = regexp.MustCompile(" ON `(.*?)`\\.\\*")
+
+func grantMatch(grant string, databaseName string) bool {
+	matches := grantOnRegexp.FindStringSubmatch(grant)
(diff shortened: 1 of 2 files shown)


go/logic/inspect.go
```diff
@@ -278,6 +275,41 @@ func (this *Inspector) validateGrants() error {
 	return this.migrationContext.Log.Errorf("User has insufficient privileges for migration. Needed: SUPER|REPLICATION CLIENT, REPLICATION SLAVE and ALL on %s.*", sql.EscapeName(this.migrationContext.DatabaseName))
 }
 
+// grantContainsAll verifies the passed grant contain all the passed strings and
+// that the grant match the DB
+func (this *Inspector) grantContainsAll(grant string, substrings ...string) bool {
```
timvaillancourt, line 280: @Gonlo2 is it possible to add a unit test that confirms the logic in `.grantContainsAll()`?

Gonlo2: @timvaillancourt Any update about this?

Issue #1133, opened 2022-06-03
Timestamp old table with a later time value


vbalys:
We are using `--timestamp-old-table` switch for our migrations. In addition, we have background jobs that drop old tables after a certain period of time. This is done in order to let table "cool down" to avoid InnoDB locks. We are consulting timestamp in table name to decide if it is already safe to finally drop it.

The obvious problem for us is that [timestamp is of the migration start](https://github.com/github/gh-ost/blob/ed46138c06430421b44fe285a26a44398b256a1c/go/base/context.go#L316). In case of a long (multiple days) migration, this timestamp provides no information of when data from the original table was used last.

There are multiple possible solutions to this:
- Use "safe" grace period before dropping table. Not really suitable for us, though, as we have quite a lot of long-running migrations and some big tables, and we want to cleanup in a couple of days after migration, not weeks or months.
- Use other means to decide when table is safe to be dropped instead of consulting timestamp in the table name. Likely a very sensitive option, but we like current solution based on timestamp too much :grin:, and we apply it in other places, not only in `gh-ost` migrations.
- Probably many other ones ...

Another option is to add capability (probably via additional modifying switch next to `--timestamp-old-table`) to use a different timestamp, likely [one of those](https://github.com/github/gh-ost/blob/ed46138c06430421b44fe285a26a44398b256a1c/go/base/context.go#L176-L178). This way old table name would reflect not the migration start moment but a moment closer to the cut-over and the factual "deprecation" of the table. 

The obvious downside would be that these timestamps are set rather late in the process, so old table name would not be defined up to that moment. This could, for example, affect hooks that make use of [old table name env variable](https://github.com/github/gh-ost/blob/ed46138c06430421b44fe285a26a44398b256a1c/go/logic/hooks.go#L54) earlier in the process, as this env variable would not be set.

Before continuing with this idea, I wanted to check with you if this capability is worth the effort of further investigation and attempts to raise PR, having in mind its deficiencies and limitations for its usage. It would be surely useful to us, so we are rather motivated to implement it.

Thanks in advance for your input.

timvaillancourt:
> We are using `--timestamp-old-table` switch for our migrations. In addition, we have background jobs that drop old tables after a certain period of time. This is done in order to let table "cool down" to avoid InnoDB locks

@vbalys 👋 right, it seems `.GetOldTableName()` uses the "start time" instead of "now" time. It might be possible to add a feature to do that but I'm not sure how it would affect users of hooks, as you mentioned

[...]

PR #1134, opened 2022-06-04
Add `Migrated` changelog event state

timvaillancourt:
Related issue: https://github.com/github/gh-ost/issues/1133

### Description

This PR adds a `Migrated` changelog state and sets it [in the `.finalCleanup()` stage](https://github.com/github/gh-ost/blob/ec199f11858193131309e15399e15ceea54e4fc5/go/logic/migrator.go#L430-L432) of the migration. This state is informational-only and is used to indicate when a migration completed

This end time of a migration should be queryable via:
```sql
SELECT last_update FROM something_ghc WHERE hint='state' AND value='Migrated';
```


dm-2:
@timvaillancourt looks like this one has a conflict that needs resolving, would you mind rebasing? 😄

PR #1146, opened 2022-07-07
Switch to golang.org/x/term; go mod vendor

dm-2:
### Description

This PR:
- switches to `golang.org/x/term` from deprecated `golang.org/x/crypto/ssh/terminal` module
- `go mod vendor`

Follow-up from:
- https://github.com/github/gh-ost/pull/1145#issuecomment-1177731105


timvaillancourt (approved):
🚀

PR #1153, opened 2022-07-20
Tear down streamer

RainbowDashy:
Related issue: https://github.com/github/gh-ost/issues/465

### Description

The PR https://github.com/github/gh-ost/pull/479 introduces "teardown" and make it possible to use gh-ost as a library. However some cleanups are missing.

This PR closes streamer properly when tearing down.

Downstream PR is https://github.com/bytebase/gh-ost/pull/3

### DIFF
--- a/go/logic/streamer.go
+++ b/go/logic/streamer.go
@@ -219,5 +219,7 @@ func (this *EventsStreamer) Close() (err error) {
 }
 
 func (this *EventsStreamer) Teardown() {
+	this.migrationContext.Log.Debugf("Tearing down...")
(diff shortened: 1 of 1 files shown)


RainbowDashy:
After a second thought, I believe that we shouldn’t reuse the `CutOverCompleteFlag` here, which would stop streaming via `canStopStreaming()` but to have a new field called `finishedMigrating` in `EventsStreamer` and test that field in `canStopStreaming()` instead. This is more consistent with the existing patterns present in `applier` , for example.

@timvaillancourt wdyt?

Issue #1157, opened 2022-08-03
gh-ost fails to appply binlog event for latin1 character set with Error 1366


wangzihuacool:
We have a table with latin1 character set. We insert a row with multibyte data into the table using latin1 connection, while gh-ost runs.Then gh-ost fails with Error 1366 Incorrect string value: '\xEF\xBF\xBD' for column 'pad' at row 1.

Executed command
```
gh-ost --chunk-size=1000 --max-lag-millis=30000 --host='127.0.0.1' --port=3307 --user=test --password=test --database='test' --table='sbtest3' --alter="ENGINE=InnoDB;" --allow-on-master  --cut-over=atomic --assume-rbr --serve-socket-file=/tmp/59e1b122-a587-4e45-a692-ea5d2380d1e9.sock -skip-foreign-key-checks -debug --execute 1>/tmp/59e1b122-a587-4e45-a692-ea5d2380d1e9.log 2>&1 &
```
While gh-ost running, we insert a row into table. 
```
set names latin1; insert into sbtest3 values(1, '1', '1,', '威');
```
Then gh-ost applying binlog into _gho table , error 1366 happens.
```
 Error 1366: Incorrect string value: '\xEF\xBF\xBD' for column 'pad' at row 1; query=
                        replace /* gh-ost `test`.`_sbtest3_gho` */ into
                                `test`.`_sbtest3_gho`
                                        (`id`, `k`, `c`, `pad`)
                                values
                                        (?, ?, ?, ?)
```

gh-ost version
```
gh-ost --version
1.1.4
```

wangzihuacool:
After analyzing and testing, I found that the problem is encoding.NewDecoder() in types.go
```
if encoding, ok := charsetEncodingMap[this.Charset]; ok {
			arg, _ = encoding.NewDecoder().String(s)
}
```
So, I added debug before and after this code to print the bytes corresponding to the characters，output below：
```
2022/08/03 14:44:24 ariginal column data:  威
2022/08/03 14:44:24 decoded column data:  å¨�
2022/08/03 14:44:24 bytes column data before decoded:  [229 168 129]  
2022/08/03 14:44:24 bytes column data after decoded:  [195 165 194 168 239 191 189]
2022-08-03 14:44:24 ERROR Error 1366: Incorrect string value: '\xEF\xBF\xBD' for column 'pad' at row 1; query=
                        replace /* gh-ost `test`.`_sbtest3_gho` */ into
                                `test`.`_sbtest3_gho`
[...]

Issue #1112, opened 2022-03-29
gh-ost fails to insert a row with Error 1366: Incorrect string value


mmeller-wikia:
Hello :)

I tried to run gh-ost to add index to a table.

Tested versions 1.1.0 and 1.14
```
gh-ost --version
1.1.4
```
Executed command
```
gh-ost --user="$USER" --password="$DB_PASSWORD" --host="$HOST" \
--database="$DATABASE" --table="$TABLE" --alter="ADD INDEX user_registration (user_registration)" \
--allow-on-master --exact-rowcount --verbose --debug --stack --execute
```
Error with stacktrace
```
2022-03-29 11:44:57 ERROR Error 1366: Incorrect string value: '\xEF\xBF\xBD\xC3\x90\xC2...' for column 'user_name' at row 1; query=
                        replace /* gh-ost `$DATABASE`.`_user_gho` */ into
                                `test`.`_user_gho`
                                        (`**redacted**`)
                                values
                                        (**redacted**)
                ; **redacted**
goroutine 70 [running]:
runtime/debug.Stack()
        /usr/local/Cellar/go/1.17.5/libexec/src/runtime/debug/stack.go:24 +0x65
runtime/debug.PrintStack()
        /usr/local/Cellar/go/1.17.5/libexec/src/runtime/debug/stack.go:16 +0x19
github.com/github/gh-ost/vendor/github.com/outbrain/golib/log.logErrorEntry(0xc0002afa40, {0x7e5660, 0xc0002afa40})
        /Users/dm-2/git/gh-ost/.gopath/src/github.com/github/gh-ost/vendor/github.com/outbrain/golib/log/log.go:178 +0x98
github.com/github/gh-ost/vendor/github.com/outbrain/golib/log.Errore(...)
        /Users/dm-2/git/gh-ost/.gopath/src/github.com/github/gh-ost/vendor/github.com/outbrain/golib/log/log.go:224
github.com/github/gh-ost/go/base.(*simpleLogger).Errore(0xc0000aa000, {0x7e5660, 0xc0002afa40})
        /Users/dm-2/git/gh-ost/.gopath/src/github.com/github/gh-ost/go/base/default_logger.go:50 +0x28
github.com/github/gh-ost/go/logic.(*Migrator).onApplyEventStruct(0xc0000aa000, 0x0)
        /Users/dm-2/git/gh-ost/.gopath/src/github.com/github/gh-ost/go/logic/migrator.go:1239 +0x2fd
[...]

PR #1158, opened 2022-08-03
Fix: Convert column value in binlog events to bytes instead of utf8 encoded unicode

wangzihuacool:
Related issue:  #1157 #1112 #532

# Description:
This PR converts character of the column data to bytes , and modifies applier connection charset from utf8mb4 to latin1, to aviod charset conversion and thus avioding Error 1366: Incorrect string value. 

The current processing method is to convert the string parsed by the binlog event into utf8 encoded unicode, and then the applier sets the database connection string to utf8mb4 to insert the unicode characters into the table.This method has a drawback, that is, assuming that the data in the latin1 table is utf8 encoded characters, then the latin1 column may contain characters that are invalid single-byte utf8 characters. Characters in the \x80-\xFF range are most common. When written to utf8mb4 column without conversion, they fail as they do not exist in the utf8 codepage.

Since Latin1 is a single-byte encoding, and all 256 values ​​​​of 1 byte are fully occupied, theoretically, any encoded value can be stored in the Latin1 field.We convert the data of character type into bytes, and then write the single-byte characters of latin1 to the table, so there is no problem of transcoding


wangzihuacool:
Restore db connection charset to utf8mb4, no need to change charset to latin1.

timvaillancourt:
@wangzihuacool thanks for this PR! Currently it is not passing CI checks, could you remedy this when you have time? 🙇

wangzihuacool:
> @wangzihuacool thanks for this PR! Currently it is not passing CI checks, could you remedy this when you have time? 🙇

Thank you for your reply. I have made some fixes for unit tests, and tests pass on my local run.
Looking forward to a CI check.

Johnson990 (approved):

wangzihuacool:
alter-charset migration tests failed. Since we changed the column value to bytes, there is no character set conversion when applying DML events.  To support changing the character set of the column with characters in the source table that are invalid in the destination table (due to charset), we need to merge PR #1003 first.

wangzihuacool:
I've made some changes. If the column value got from binlog event is a string type and no charset conversion, it will be converted to bytes;otherwise we just fall back to the previous way of dealing with characters. 
Now the migration tests on MySQL 5.7 and 8.0 both pass on my local run. Looking forward to a CI check. Thank you.

dm-2 (approved):

PR #1164, opened 2022-08-13
Improve applier `.ReadMigrationRangeValues()` func accuracy

timvaillancourt:
### Description

This PR moves the applier `.ReadMigrationRangeValues()` func to use a single transaction for both the min and max queries

Gathering these values in the same transaction is more accurate than the current approach where 2 x different auto-commit transactions grab these values

Lastly, the min/max funcs were made "private" because they're only called by `.ReadMigrationRangeValues()`

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -439,14 +439,14 @@ func (this *Applier) ExecuteThrottleQuery() (int64, error) {
 }
 
 // ReadMigrationMinValues returns the minimum values to be iterated on rowcopy
-func (this *Applier) ReadMigrationMinValues(uniqueKey *sql.UniqueKey) error {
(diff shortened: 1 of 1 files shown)


shlomi-noach (approved):
This change is fine. Ultimately it does not matter that much: we take note of the binlog position before we evaluate the min/max range values. That means however we read the min/max values, any changes that happen in between -- we will address those changes as we iterate the binary logs; the changes will be covered by the binlog stream that we process.

timvaillancourt:
> This change is fine. 

Thanks @shlomi-noach! Yeah I think so, I'm glad this didn't break any assumptions I didn't know about 👍 

> we will address those changes as we iterate the binary logs; the changes will be covered by the binlog stream that we process.

Agreed, I think this is more of a cosmetic change that won't actually affect the correctness of the end-result

dm-2 (approved):

PR #1170, opened 2022-08-25
Enable the errcheck linter

timvaillancourt:
### Description

This PR enables the [`errcheck` ](https://github.com/kisielk/errcheck) linter in the project `golangci-lint` config:

> errcheck is a program for checking for unchecked errors in go programs.

Also a lot of missing error handling was added for problems (or `// nolint:errcheck` flags for some false-positives)

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -146,19 +146,19 @@ func (this *Applier) readTableColumns() (err error) {
 }
 
 // showTableStatus returns the output of `show table status like '...'` command
-func (this *Applier) showTableStatus(tableName string) (rowMap sqlutils.RowMap) {
+func (this *Applier) showTableStatus(tableName string) (rowMap sqlutils.RowMap, err error) {
 	query := fmt.Sprintf(`show /* gh-ost */ table status from %s like '%s'`, sql.EscapeName(this.migrationContext.DatabaseName), tableName)
-	sqlutils.QueryRowsMap(this.db, query, func(m sqlutils.RowMap) error {
+	err = sqlutils.QueryRowsMap(this.db, query, func(m sqlutils.RowMap) error {
 		rowMap = m
 		return nil
 	})
-	return rowMap
+	return rowMap, err
 }
 
 // tableExists checks if a given table exists in database
-func (this *Applier) tableExists(tableName string) (tableFound bool) {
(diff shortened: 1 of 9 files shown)


go/logic/throttler.go
```diff
@@ -441,8 +447,15 @@ func (this *Throttler) initiateThrottlerCollection(firstThrottlingCollected chan
 	go this.collectThrottleHTTPStatus(firstThrottlingCollected)
 
 	go func() {
-		this.collectGeneralThrottleMetrics()
-		firstThrottlingCollected <- true
+		for {
+			if err := this.collectGeneralThrottleMetrics(); err != nil {
```
dm-2, line 451: We're slowly replicating the entire ticker loop that follows, perhaps we should refactor `initiateThrottlerCollection` to something along the lines of the below - what do you think?

(`notifyFirstThrottlingCollected` could probably be better named!)

```go
func (this *Throttler) initiateThrottlerCollection(notifyFirstThrottlingCollected chan<- bool) {
	go this.collectReplicationLag(notifyFirstThrottlingCollected)
	go this.collectControlReplicasLag()
	go this.collectThrottleHTTPStatus(notifyFirstThrottlingCollected)
        firstMetricsCollected := false

	go func() {
		ticker := time.NewTicker(time.Second)
		defer ticker.Stop()
		for range ticker.C {
			if atomic.LoadInt64(&this.finishedMigrating) > 0 {
				return
			}

			err := this.collectGeneralThrottleMetrics()
			if err != nil {
				this.migrationContext.Log.Errorf("Failed to collect throttler metrics: %+v", err)
			} else if !firstMetricsCollected {
				notifyFirstThrottlingCollected <- true
				firstMetricsCollected = true
			}
		}
	}()
}
```

go/logic/applier.go
```diff
@@ -378,7 +384,9 @@ func (this *Applier) WriteChangelog(hint, value string) (string, error) {
 }
 
 func (this *Applier) WriteAndLogChangelog(hint, value string) (string, error) {
```
dm-2, line 386: It looks like we never actually use the `string` return value from the `WriteAndLogChangelog`, `WriteChangelogState`, and `WriteChangelog` functions - what do you think about getting rid of it and just returning `error`?

Issue #1171, opened 2022-08-26
because lock is not release,   drop cutover sentry table is hang


lmtwga:
### reason
-   when  atomicCutOver() -> this.waitForEventsUpToLock() != nil , then will drop  _del table https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L668
- There are two places to delete _del table:
  - atomicCutOver() -> defer func() -> dropCutOverSentryTableOnce.Do -> this.applier.DropAtomicCutOverSentryTableIfExists() 
     https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L647
  - AtomicCutOverMagicLock() - > dropCutOverSentryTableOnce.Do -> tx.Exec(query)  https://github.com/github/gh-ost/blob/3f44e04365725e2d52ef05bdb997077a4d07abe1/go/logic/applier.go#L980

- but, two places use Once.Do()。Who executes first depends on luck!
- Unfortunately，when 'atomicCutOver() -> defer func() -> dropCutOverSentryTableOnce.Do -> this.applier.DropAtomicCutOverSentryTableIfExists() ' was executed firstly，the lock of lock _del table has not been released, and because the db connection db.SetMaxOpenConns(3)  https://github.com/github/gh-ost/blob/master/go/mysql/utils.go#L56
so, the first place to drop _del table will use a different MySQL thread, will hang

lmtwga:
@shlomi-noach Please check when you have time
the Pr: https://github.com/github/gh-ost/pull/1173

lmtwga:
@shlomi-noach Please review when you have time, thank you very much!

lmtwga:
@timvaillancourt  @dm-2 @jsoref  Please review when you have time, thank you very much!

dm-2:
:wave: @lmtwga, thanks for the issue and PR!

Nice debugging! Just to summarise what's happening:
1. If `atomicCutOver` function exits because of an error (i.e. before `true` is sent to the `okToUnlockTable` channel), it executes its `defer` block whilst `AtomicCutOverMagicLock` function still holds the table lock that includes the sentry table
2. The `defer` block for `atomicCutOver` sends `true` to the `okToUnlockTable` channel, causing a race to the `dropCutOverSentryTableOnce`-protected drop of the sentry table in both the `atomicCutOver` and `AtomicCutOverMagicLock` functions
[...]

PR #1180, opened 2022-09-16
fix: because lock is not release, drop cutover sentry table is hanged

lmtwga:
Related issue: https://github.com/github/gh-ost/issues/1171

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -9,7 +9,6 @@ import (
 	gosql "database/sql"
 	"fmt"
 	"strings"
-	"sync"
 	"sync/atomic"
 	"time"
 
@@ -904,7 +903,7 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
 }
 
 // AtomicCutOverMagicLock
-func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocked chan<- error, okToUnlockTable <-chan bool, tableUnlocked chan<- error, dropCutOverSentryTableOnce *sync.Once) error {
+func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocked chan<- error, okToUnlockTable <-chan bool, tableUnlocked chan<- error) error {
 	tx, err := this.db.Begin()
 	if err != nil {
 		tableLocked <- err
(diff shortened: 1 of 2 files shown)


lmtwga:
@dm-2 
Please review when you have time.
Thank you very much!

lmtwga:
@dm-2 @timvaillancourt 
Please review when you have time.
Thank you very much!

dm-2:
The problem and the fix make sense to me :+1:

@timvaillancourt would you mind taking a look as well? 🙇 Your thoughts would be much appreciated, given that it's such a critical section of code!

gtowey-air:
Hey @dm-2; I just got hit with this bug as well and I would love to get this or some related fix merged -- would you mind if I took a look and did some testing on this PR to validate?

go/logic/applier.go
```diff
@@ -983,12 +983,10 @@ func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocke
...
 	)
 
-	dropCutOverSentryTableOnce.Do(func() {
-		if _, err := tx.Exec(query); err != nil {
-			this.migrationContext.Log.Errore(err)
-			// We DO NOT return here because we must `UNLOCK TABLES`!
-		}
-	})
+	if _, err := tx.Exec(query); err != nil {
+		this.migrationContext.Log.Errore(err)
```
gtowey-air, line 987: It might be better to execute this query with a timeout and exit the program if it fails.  While this PR should change ghost so that it will always successfully unlock the tables, in the event that there is some other unforseen race condition or deadlock here, then stopping program execution would be a guaranteed way to at least release all locks held by this program.

It would have the unfortunate side effect of leaving the `_del` table in existence which would need to be cleaned up later, but IMO that's operationally better than holding table locks which will cause any application actively trying to use the tables to stall.

gtowey-air:
@rashiq @dm-2 I've tested this PR and it looks like it fixes the problem.  I recommend that we go ahead and merge this!

timvaillancourt:
> @timvaillancourt would you mind taking a look as well?

Sorry for the delay, this LGTM 👍

timvaillancourt (approved):

rashiq (approved):

gtowey-air:
@rashiq Thanks for the approval!  Would you be able to merge this?

rashiq:
@gtowey-air Hey! (also nice to see you!) - sorry for the delay, merging now

Issue #1198, opened 2022-11-10
Use mysql 8.0 INSTANT DDL if supported


morgo:
Hi friends,

I would like to propose a new feature (likely on by default) where gh-ost attempts to apply the DDL change first (but with an assertion that it must be `INSTANT`). This is particularly targeted at MySQL 8.0+ users, but will likely work on some earlier versions as well (Aurora 5.7 supports instant changes, although I've not specifically tested with the instant assertion..).

I am happy to contribute a patch for it - but per your request I am opening an issue here first to discuss :-)

I hacked up something just now to show that it looks easy enough to add:
https://github.com/github/gh-ost/compare/master...morgo:gh-ost:attempt-instant-ddl?expand=1

For an example usage:
```bash
go run . \
--max-load=Threads_running=25 \
--critical-load=Threads_running=1000 \
--chunk-size=1000 \
--user="msandbox" \
--password="msandbox" \
--host="127.0.0.1" \
--port=8031 \
--allow-on-master \
--database="test" \
--table="customers" \
--verbose \
--initially-drop-old-table \
--ok-to-drop-table \
--execute  --initially-drop-ghost-table --alter="add newcol9 int"
2022-11-09 20:17:59 INFO starting gh-ost
2022-11-09 20:17:59 INFO Migrating `test`.`customers`
2022-11-09 20:17:59 INFO inspector connection validated on 127.0.0.1:8031
2022-11-09 20:17:59 INFO User has SUPER, REPLICATION SLAVE privileges, and has ALL privileges on `test`.*
2022-11-09 20:17:59 INFO binary logs validated on 127.0.0.1:8031
2022-11-09 20:17:59 INFO Restarting replication on 127.0.0.1:8031 to make sure binlog settings apply to replication thread
2022-11-09 20:17:59 INFO Inspector initiated on mtocker-macbookpro.local:8031, version 8.0.31
2022-11-09 20:17:59 INFO Table found. Engine=InnoDB
2022-11-09 20:17:59 INFO Estimated number of rows via EXPLAIN: 9
2022-11-09 20:17:59 INFO Recursively searching for replication master
2022-11-09 20:17:59 INFO Master found to be mtocker-macbookpro.local:8031
2022-11-09 20:17:59 INFO log_slave_updates validated on 127.0.0.1:8031
2022-11-09 20:17:59 INFO Attempting to execute ALTER TABLE as INSTANT DDL
[...]

PR #1201, opened 2022-11-14
Attempt INSTANT DDL first

morgo:
Fixes https://github.com/github/gh-ost/issues/1198

### Description

This adds support for attempting INSTANT DDL from MySQL 8.0 before applying gh-ost's original algorithm. It is disabled by default for now (per suggestion), although in future it makes sense to enable by default and hide behind version detection.

### DIFF
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -45,6 +45,22 @@ If you happen to _know_ your servers use RBR (Row Based Replication, i.e. `binlo
 Skipping this step means `gh-ost` would not need the `SUPER` privilege in order to operate.
 You may want to use this on Amazon RDS.
 
+### attempt-instant-ddl
+
+MySQL 8.0 support "instant DDL" for some operations. If an alter statement can be completed with instant DDL, only a metadata change is required internally, so MySQL will return _instantly_ (only requiring a metadata lock to complete). Instant operations include:
+
+- Adding a column
+- Dropping a column
+- Dropping an index
+- Extending a varchar column
+- Adding a virtual generated column
+
+It is not reliable to parse the `ALTER` statement to determine if it is instant or not. This is because the table might be in an older row format, or have some other incompatibility that is difficult to identify.
+
+The risks of attempting to instant DDL are relatively minor: Gh-ost may need to acquire a metadata lock at the start of the operation. This is not a problem for most scenarios, but it could be a problem for users that start the DDL during a period with long running transactions.
+
+gh-ost will automatically fallback to the normal DDL process if the attempt to use instant DDL is unsuccessful.
+
 ### conf
 
 `--conf=/path/to/my.cnf`: file where credentials are specified. Should be in (or contain) the following format:
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -47,7 +47,7 @@ You may want to use this on Amazon RDS.
 
 ### attempt-instant-ddl
 
-MySQL 8.0 support "instant DDL" for some operations. If an alter statement can be completed with instant DDL, only a metadata change is required internally, so MySQL will return _instantly_ (only requiring a metadata lock to complete). Instant operations include:
+MySQL 8.0 supports "instant DDL" for some operations. If an alter statement can be completed with instant DDL, only a metadata change is required internally. Instant operations include:
 
 - Adding a column
 - Dropping a column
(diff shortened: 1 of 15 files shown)


go/logic/migrator.go
```diff
@@ -351,6 +351,18 @@ func (this *Migrator) Migrate() (err error) {
 	if err := this.initiateInspector(); err != nil {
 		return err
 	}
+	// In MySQL 8.0 (and possibly earlier) some DDL statements can be applied instantly.
```
timvaillancourt, line 354: I think we should start binlog streaming before the instant DDL is attempted. This might not change integrity, but it would be more consistent with the current approach/docs

Moving this below `this.initiateStreaming()` would achieve this

go/logic/migrator.go
```diff
@@ -734,6 +746,17 @@ func (this *Migrator) initiateServer() (err error) {
 	return nil
 }
 
+// attemptInstantDDL tries to apply the DDL statement directly to the table
+// using a ALGORITHM=INSTANT assertion. If this fails, it will return an error,
+// in which case the original algorithm should be used.
+func (this *Migrator) attemptInstantDDL() (err error) {
+	this.applier = NewApplier(this.migrationContext)
```
timvaillancourt, line 753: @morgo assuming my earlier suggestion is made I'd like to avoid opening the applier twice, just in case

I'm hoping the applier setup by `[this.initiateApplier()](https://github.com/github/gh-ost/blob/master/go/logic/migrator.go#L357-L359)` in `.Migrate()` can be used here

go/logic/applier_test.go
```diff
@@ -170,3 +170,17 @@ func TestApplierBuildDMLEventQuery(t *testing.T) {
...
+	migrationContext.DatabaseName = "test"
+	migrationContext.OriginalTableName = "mytable"
+	migrationContext.AlterStatementOptions = "ADD INDEX (foo)"
+	applier := NewApplier(migrationContext)
+
+	t.Run("instantDDLstmt", func(t *testing.T) {
+		stmt := applier.generateInstantDDLQuery()
+		test.S(t).ExpectEquals(stmt, "ALTER /* gh-ost */ TABLE `test`.`mytable` ADD INDEX (foo), ALGORITHM=INSTANT")
+	})
+
```
timvaillancourt, line 185: The linter doesn't like this trailing space
```suggestion
```

timvaillancourt (commented):
Thanks @morgo! I like the flag approach 👍

Added a few comments inline

go/logic/migrator.go
morgo: Sure, i've moved it below flag file creation as well. The only downside is the log writes more info that is unnecessary. But it's a non-default flag now, and I think that's fine.

go/logic/migrator.go
```diff
@@ -360,6 +360,17 @@ func (this *Migrator) Migrate() (err error) {
...
 	}
+	// In MySQL 8.0 (and possibly earlier) some DDL statements can be applied instantly.
+	// Attempt to do this if AttemptInstantDDL is set.
+	if this.migrationContext.AttemptInstantDDL {
+		this.migrationContext.Log.Infof("Attempting to execute alter with ALGORITHM=INSTANT")
+		if err := this.attemptInstantDDL(); err == nil {
+			this.migrationContext.Log.Infof("Success! table %s.%s migrated instantly", sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.OriginalTableName))
+			return nil
+		} else {
+			this.migrationContext.Log.Infof("ALGORITHM=INSTANT failed, proceeding with original algorithm: %s", err)
```
dm-2, line 371:
```suggestion
			this.migrationContext.Log.Infof("ALGORITHM=INSTANT not supported for this operation, proceeding with original algorithm: %s", err)
```

What do you think about changing the wording here? Although the operation technically has failed, outputting "failed" may cause concern for some users even though gh-ost overall has not failed.

dm-2 (commented):
I've added some thoughts around docs/output - the actual code looks good to me though :+1:

doc/command-line-flags.md
```diff
@@ -45,6 +45,22 @@ If you happen to _know_ your servers use RBR (Row Based Replication, i.e. `binlo
 Skipping this step means `gh-ost` would not need the `SUPER` privilege in order to operate.
 You may want to use this on Amazon RDS.
 
+### attempt-instant-ddl
```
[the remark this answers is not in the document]
morgo: Will do. I can modify the 3rd paragraph to say:

```
`--attempt-instant-ddl` is disabled by default, but the risks of enabling it are relatively minor: `gh-ost` may need to acquire a metadata lock at the start of the operation. This is not a problem for most scenarios, but it could be a problem for users that start the DDL during a period with long running transactions.
```

go/logic/migrator.go
```diff
@@ -734,6 +745,13 @@ func (this *Migrator) initiateServer() (err error) {
 	return nil
 }
 
+// attemptInstantDDL tries to apply the DDL statement directly to the table
```
timvaillancourt, line 748: Last nit, I swear 😅! This func (mostly the comment) feels redundant

What do you think about `.Migrate()` calling `this.applier.AttemptInstantDDL()` directly? I think that might be more straightforward to read, but totally optional

timvaillancourt (dismissed):

timvaillancourt (approved):
LGTM 👍. Thanks again @morgo!

dm-2 (approved):
Thanks @morgo, this looks great! :rocket:

PR #1202, opened 2022-11-15
Fix migration test passing when `convert-utf8mb4` test setup was partially failing

dm-2:
I noticed in the CI for another PR that we had an error message that was being ignored:
```
ERROR 1136 (21S01) at line 10: Column count doesn't match value count at row 1
```

This was in the test setup for `convert-utf8mb4`, and was caused by an `insert` statement not having enough fields (it should have had four fields because the last two columns in the test table are `not null`):
```
insert into gh_ost_test values (null, 'átesting');
```

This PR addresses this by:
- adding error checking to migration test setup so that we can detect when tests are not setup correctly (as the actual test succeeded when run in this case)
- fixing the error in the test setup for `convert-utf8mb4` by removing the unnecessary invalid `insert` statement

I have also added a `sort` for the list of tests so that it's easier to track progress through the list of tests.

### DIFF
--- a/localtests/convert-utf8mb4/create.sql
+++ b/localtests/convert-utf8mb4/create.sql
@@ -7,9 +7,6 @@ create table gh_ost_test (
   primary key(id)
 ) auto_increment=1;
 
-insert into gh_ost_test values (null, 'átesting');
-
-
 insert into gh_ost_test values (null, 'Hello world, Καλημέρα κόσμε, コンニチハ', 'átesting0', 'initial');
 
 drop event if exists gh_ost_test;
--- a/localtests/test.sh
+++ b/localtests/test.sh
@@ -116,7 +116,8 @@ test_single() {
     gh-ost-test-mysql-replica --default-character-set=utf8mb4 test -e "set @@global.sql_mode='$(cat $tests_path/$test_name/sql_mode)'"
   fi
 
-  gh-ost-test-mysql-master --default-character-set=utf8mb4 test < $tests_path/$test_name/create.sql
+  cat $tests_path/$test_name/create.sql
+  bash -x gh-ost-test-mysql-master --default-character-set=utf8mb4 test < $tests_path/$test_name/create.sql
 
   extra_args=""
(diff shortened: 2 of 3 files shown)


dm-2:
Broken test setup for `convert-utf8mb4` failing as expected after adding error checking to test setup:

```
Testing: convert-utf8mb4..
ERROR 1136 (21S01) at line 10: Column count doesn't match value count at row 1
ERROR convert-utf8mb4 create failure. cat ./localtests/convert-utf8mb4/create.sql:
drop table if exists gh_ost_test;
create table gh_ost_test (
  id int auto_increment,
  t varchar(128) charset utf8 collate utf8_general_ci,
  tl varchar(128) charset latin1 not null,
  ta varchar(128) charset ascii not null,
  primary key(id)
) auto_increment=1;

insert into gh_ost_test values (null, 'átesting');


insert into gh_ost_test values (null, 'Hello world, Καλημέρα κόσμε, コンニチハ', 'átesting0', 'initial');

drop event if exists gh_ost_test;
delimiter ;;
create event gh_ost_test
  on schedule every 1 second
  starts current_timestamp
  ends current_timestamp + interval 60 second
  on completion not preserve
  enable
  do
begin
  insert into gh_ost_test values (null, md5(rand()), 'átesting-a', 'a');
  insert into gh_ost_test values (null, 'novo proprietário', 'átesting-b', 'b');
  insert into gh_ost_test values (null, '2H₂ + O₂ ⇌ 2H₂O, R = 4.7 kΩ, ⌀ 200 mm', 'átesting-c', 'c');
  insert into gh_ost_test values (null, 'usuário', 'átesting-x', 'x');

  delete from gh_ost_test where ta='x' order by id desc limit 1;
end ;;
+ FAIL
Error: Process completed with exit code 1.
```

dm-2:
Migration tests now passing as expected - fixed by removing invalid `insert` statement that wasn't necessary for testing.

alexisjcarr (approved):

PR #1206, opened 2022-11-24
Make migration tests CI faster by using tmfs instead of disk

dm-2:
Testing using tmpfs to speed up migration tests


dm-2:
closing as experimental - best approach to speed up migration tests CI would be to parallelise gh-ost tests

Issue #1204, opened 2022-11-17
Performance Improvements (Raw Ideas)


morgo:
We are looking to improve the performance of `gh-ost` so that we can safely operate with larger database sizes. Our use case is a bit different to GitHub’s in that we **do not use read-replicas**. Some of the DBs have very cyclic usage as well (i.e. only busy 9-5 M-F), and may have windows of free capacity.

I have a few ideas I wanted to run by you since I’m sure some have come up prior:

| Feature  | Context |
| ------------- | ------------- |
| Use 8.0 ALGORITHM=INSTANT when applicable | [Merged](https://github.com/github/gh-ost/pull/1201), thank you kindly for the review!  |
| Dynamic Chunk Size  | Gh-ost can observe the exec time of processing chunks, and dynamically increase the size if it fits below a threshold. For our environment (because we have a lot of replica tolerance) we typically run larger batch sizes, but have varying DB instance sizes. Being able to have this auto-tune is a win for us.  |
| Multi-threaded applier | Parallel replication apply is much better in MySQL 8.0 – so combined with that we don’t use read-replicas, we can probably push more changes through the binlog than `gh-ost` currently does. We think we can tolerate a few minutes of replica lag. Our limit is Aurora restricts the relay log to ~1000M, if we exceed that.. we reduce our DR capabilities. (Note: there's an [earlier issue](https://github.com/github/gh-ost/issues/193) on this. It lacks the 8.0 parallel context, and the issue @shlomi-noach probably hit when he said it is slower, is possibly [this one](https://dev.mysql.com/doc/refman/8.0/en/innodb-auto-increment-handling.html#innodb-auto-increment-lock-modes)? In any case, I've verified I can bulk-parallel insert with improved performance.)  |
| Defer Binary Log Apply |  Currently gh-ost prioritizes applying the binary log ahead of copying rows. I actually think it’s possible to track only the primary keys that were discovered in the binary log in memory + if the _last modification_ was a delete or not (bool). If this is kept in a map, then it can be applied after the copy is done. The benefit of this change is most evident in workloads that tend to update the same rows.  |
| Resume from failure | I know there is a [stale PR for this](https://github.com/github/gh-ost/pull/343). This doesn’t improve the performance, but it’s semi-related since some of our long running DDLs fail. We also like to use daily pod-cycling on our k8s clusters, so having 2 week long single processes complicates our infra. |

[...]

PR #1224, opened 2022-12-07
Support for dynamic chunk size

morgo:

Related issue: https://github.com/github/gh-ost/issues/1204

### Description

This PR dynamically adjusts the `chunkSize` to be based on execution time feedback. It is disabled by default for backward compatibility, but should be considered a default for future releases. As well as improving performance (my motivation) it is a little bit safer than a static `chunkSize` considering wide-tables with a lot of secondary indexes.

There are also some guard rails in place to make sure if the observed execution time is 5x the target, it quickly scales down the `chunkSize` (which is not a problem, as it will later scale back up if this was a temporary blip).


meiji163:
This looks promising, I'll do some testing on larger tables.

PR #1227, opened 2022-12-11
Base validate connection cleanup

timvaillancourt:
### Description

This PR simplifies the code in `base.ValidateConnection` and reduces the number of queries from 3 -> 2 in most cases by gathering the `@@global.version` and @@global.port` in the same query

Later, the `@@global.port` variable is checked for `NULL` as this can happen on some cloud MySQL flavours


timvaillancourt:
Close in favour of https://github.com/github/gh-ost/pull/1229

Issue #1235, opened 2022-12-27
add the possibility to toggle the optimizer switch during the migration session


cyrinux:
Hi,

We realized we have for some databases the need to be able to set the migration session optimizer switch, for example set `prefer_ordering_index=on`

https://dev.mysql.com/doc/refman/8.0/en/switchable-optimizations.html

I will do a PR for this.

PR #1236, opened 2022-12-27
feat(#1235): add an optimizer switch flag to migration context

cyrinux:
### Description

This add a `optimizer_switch` string flag.

`--optimizer-switch="prefer_ordering_index=on"`

I don't check the value, easier to maintains, this will just fail if its a unknown value, depending the mysql/maria/percona version.

Related issue: https://github.com/github/gh-ost/issues/1235

### DIFF
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -139,6 +139,7 @@ func main() {
 	criticalLoad := flag.String("critical-load", "", "Comma delimited status-name=threshold, same format as --max-load. When status exceeds threshold, app panics and quits")
 	flag.Int64Var(&migrationContext.CriticalLoadIntervalMilliseconds, "critical-load-interval-millis", 0, "When 0, migration immediately bails out upon meeting critical-load. When non-zero, a second check is done after given interval, and migration only bails out if 2nd check still meets critical load")
 	flag.Int64Var(&migrationContext.CriticalLoadHibernateSeconds, "critical-load-hibernate-seconds", 0, "When non-zero, critical-load does not panic and bail out; instead, gh-ost goes into hibernation for the specified duration. It will not read/write anything from/to any server")
+	flag.StringVar(&migrationContext.OptimizerSwitch, "optimizer-switch", "", "Optimizer switch params, eg: prefer_ordering_index=on")
 	quiet := flag.Bool("quiet", false, "quiet")
 	verbose := flag.Bool("verbose", false, "verbose")
 	debug := flag.Bool("debug", false, "debug mode (very verbose)")
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -139,7 +139,7 @@ func main() {
 	criticalLoad := flag.String("critical-load", "", "Comma delimited status-name=threshold, same format as --max-load. When status exceeds threshold, app panics and quits")
 	flag.Int64Var(&migrationContext.CriticalLoadIntervalMilliseconds, "critical-load-interval-millis", 0, "When 0, migration immediately bails out upon meeting critical-load. When non-zero, a second check is done after given interval, and migration only bails out if 2nd check still meets critical load")
 	flag.Int64Var(&migrationContext.CriticalLoadHibernateSeconds, "critical-load-hibernate-seconds", 0, "When non-zero, critical-load does not panic and bail out; instead, gh-ost goes into hibernation for the specified duration. It will not read/write anything from/to any server")
-	flag.StringVar(&migrationContext.OptimizerSwitch, "optimizer-switch", "", "Optimizer switch params, eg: prefer_ordering_index=on")
+	flag.StringVar(&migrationContext.OptimizerSwitch, "optimizer-switch", "prefer_ordering_index=on", "optimizer_switch param")
 	quiet := flag.Bool("quiet", false, "quiet")
 	verbose := flag.Bool("verbose", false, "verbose")
 	debug := flag.Bool("debug", false, "debug mode (very verbose)")
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -49,12 +49,12 @@ func newDmlBuildResultError(err error) *dmlBuildResult {
 	}
 }
 
-func (this *Applier) setOptimizerSwitch(tx *gosql.Tx) error {
+func (this *Applier) setOptimizerSwitch() error {
 	if this.migrationContext.OptimizerSwitch == "" {
 		return nil
 	}
 	optimizerString := fmt.Sprintf("SET SESSION optimizer_switch=%q", this.migrationContext.OptimizerSwitch)
-	_, err := tx.Query(optimizerString)
+	_, err := this.db.Query(optimizerString)
 	return err
 }
 
@@ -108,6 +108,10 @@ func (this *Applier) InitDBConnections() (err error) {
 		} else {
 			this.connectionConfig.ImpliedKey = impliedKey
 		}
+	}
+	err = this.setOptimizerSwitch()
+	if err != nil {
+		return err
 	}
 	if err := this.readTableColumns(); err != nil {
 		return err
@@ -555,11 +559,6 @@ func (this *Applier) ReadMigrationRangeValues() error {
 		return err
 	}
 	defer tx.Rollback()
-
-	err = this.setOptimizerSwitch(tx)
-	if err != nil {
-		return err
-	}
 
 	if err := this.readMigrationMinValues(tx, this.migrationContext.UniqueKey); err != nil {
 		return err
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -49,6 +49,15 @@ func newDmlBuildResultError(err error) *dmlBuildResult {
 	}
 }
 
+func (this *Applier) setOptimizerSwitch(tx *gosql.Tx) error {
+	if this.migrationContext.OptimizerSwitch == "" {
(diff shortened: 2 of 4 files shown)


go/logic/applier.go
```diff
@@ -49,6 +49,15 @@ func newDmlBuildResultError(err error) *dmlBuildResult {
 	}
 }
 
+func (this *Applier) setOptimizer(tx *gosql.Tx) error {
```
timvaillancourt, line 52: @cyrinux the `setOptimizer` func-name feels a bit unclear about what it performs I think `setOptimizerSwitch` is a more descriptive func name. What do you think?

go/logic/applier.go
```diff
@@ -49,6 +49,15 @@ func newDmlBuildResultError(err error) *dmlBuildResult {
...
+	if this.migrationContext.OptimizerSwitch == "" {
+		return nil
+	}
+	optimizerString := fmt.Sprintf("set session optimizer_switch=%q", this.migrationContext.OptimizerSwitch)
```
timvaillancourt, line 56:
```suggestion
	optimizerString := fmt.Sprintf("SET SESSION optimizer_switch=%q", this.migrationContext.OptimizerSwitch)
```
I suggest we use capital `SET SESSION` to match the rest of `applier.go`

go/logic/applier.go
```diff
@@ -502,6 +516,11 @@ func (this *Applier) readMigrationMaxValues(tx *gosql.Tx, uniqueKey *sql.UniqueK
 		return err
 	}
 
+	err = this.setOptimizer(tx)
```
timvaillancourt, line 519: @cyrinux looks like `setOptimizer` is ran in both `readMigrationMinValues` and `readMigrationMaxValues`, which means twice in the same SQL transaction

These 2 x funcs are exclusively called [by `ReadMigrationRangeValues`](https://github.com/github/gh-ost/blob/71df3e8d8affca587ee5237391c89533420a5ba3/go/logic/applier.go#L541-L558) which passes both funcs the same transaction

Can we run `setOptimizer` only once in `ReadMigrationRangeValues` [after this line here](https://github.com/github/gh-ost/blob/71df3e8d8affca587ee5237391c89533420a5ba3/go/logic/applier.go#L567)?

timvaillancourt (changes requested):
@cyrinux thanks for the PR! I have left a few comments/nits in line

Two more questions:
1. For context/curiosity, in what situations or environments are you needing to override the `optimizer_switch`? This might be useful to know or document
2. Can you please update [`doc/command-line-flags.md`](https://github.com/github/gh-ost/blob/master/doc/command-line-flags.md) with a description of this new flag?

🙇

go/logic/applier.go
cyrinux: Sure, thanks for the review !

cyrinux:
> @cyrinux thanks for the PR! I have left a few comments/nits in line
> 
> Two more questions:
> 
>     1. For context/curiosity, in what situations or environments are you needing to override the `optimizer_switch`? This might be useful to know or document
> 
>     2. Can you please update [`doc/command-line-flags.md`](https://github.com/github/gh-ost/blob/master/doc/command-line-flags.md?rgh-link-date=2022-12-27T18%3A27%3A27Z) with a description of this new flag?
> 
> 
> 🙇

I'm not DBA myself, but I understand sometimes we set those `optmizer_switch` on `GLOBAL` and then depending the value, this can slow by 10 in my case a simple `count(*)`. It was the case here, unable to use the tools after some months and tweaking.

go/cmd/gh-ost/main.go
```diff
@@ -139,6 +139,7 @@ func main() {
 	criticalLoad := flag.String("critical-load", "", "Comma delimited status-name=threshold, same format as --max-load. When status exceeds threshold, app panics and quits")
 	flag.Int64Var(&migrationContext.CriticalLoadIntervalMilliseconds, "critical-load-interval-millis", 0, "When 0, migration immediately bails out upon meeting critical-load. When non-zero, a second check is done after given interval, and migration only bails out if 2nd check still meets critical load")
 	flag.Int64Var(&migrationContext.CriticalLoadHibernateSeconds, "critical-load-hibernate-seconds", 0, "When non-zero, critical-load does not panic and bail out; instead, gh-ost goes into hibernation for the specified duration. It will not read/write anything from/to any server")
+	flag.StringVar(&migrationContext.OptimizerSwitch, "optimizer-switch", "", "Optimizer switch params, eg: prefer_ordering_index=on")
```
timvaillancourt, line 142:
```suggestion
	flag.StringVar(&migrationContext.OptimizerSwitch, "optimizer-switch", "prefer_ordering_index=on", "optimizer_switch param")
```
I think we can make `prefer_ordering_index=on` a default setting as it is already the MySQL default and could make results more correct

For example, the min/max value queries for the migration rely on the index being ordered. "prefer" is good, but this might highlight a need for a `FORCE INDEX` when we get the min/max values longer term 🤔 

Any thoughts here @dm-2 / @rashiq / @morgo?

go/logic/applier.go
```diff
@@ -547,6 +556,11 @@ func (this *Applier) ReadMigrationRangeValues() error {
 	}
 	defer tx.Rollback()
 
+	err = this.setOptimizerSwitch(tx)
```
timvaillancourt, line 559: @cyrinux on 2nd thought, this applies to the whole applier connection/session and it would be less confusing to locate this call [within `.InitDBConnections()`](https://github.com/github/gh-ost/blob/eec7aa5feed3f18f65c98525a968220cb254e237/go/logic/applier.go#L84) when the connection is first opened vs in the middle

The only difference is `setOptimizerSwitch(...)` wouldn't use a transaction, the `this.db` handle _(`*sql.DB` not `*sql.Tx`)_ can be used directly instead. `.Query(...)` will work the same on the `this.db` handle

timvaillancourt (changes requested):
@cyrinux thanks for the changes! Two more suggestions/questions added on 2nd thought, my apologies.

go/cmd/gh-ost/main.go
morgo: `Prefer_ordering_index` is `ON` by default, but my opinion is that this is only for backward compatibility. @cyrinux are you sure you don't need to turn it `OFF` for your use-case?

I don't like `Prefer_ordering_index` as an optimization, since my experience has been it often makes mistakes (leading to very bad performance). We are talking about the same issue [written about here](https://blog.jcole.us/2019/09/30/reconsidering-access-paths-for-index-ordering-a-dangerous-optimization-and-a-fix/).

We have ran into this problem on my team (not for gh-ost, for other workloads), but our workaround has been to use `FORCE INDEX`. This is only because the switch shipped after 5.7-GA so it's not in Aurora v2.

cyrinux: @morgo I'm sure we toggle `Prefer_ordering_index=on` 😉 
And yes, its the default behavior but not for me. @morgo @timvaillancourt should I so add FORCE INDEX to the `SELECT COUNT(*)` as this make sense to always use it for this kind of query ?

dland:
For people who stumble upon this later on and want to know the reason from a DBA point of view (I worked with Cyril to put this together)...

The backstory behind this is that we have a stable application that has been running for over 15 years on Mysql. When we upgraded from 5.7 to 8.0, we observed some severe performance impacts on a certain set of queries. It would have been particularly hard to rewrite them (since they are generated by a query builder that no-one is particularly keen to hack on in case the change introduces other side effects).

We worked out that we could recover the original performance (at the expense of marginally slowing a few other types of queries, which was deemed to be an acceptable tradeoff) by setting `optimizer_switch='prefer_ordering_index=off'`. And then we were able to completely forget about the issue.

Fast forward to a recent gh-ost session where an execute would repeatedly crash with
```
[mysql] 2022/12/21 16:32:48 packets.go:37: unexpected EOF
```

Why? Consider first, a source table:
```
CREATE TABLE thing (
    thing_id int not null auto_increment primary key
);
```
The table I was trying to ghost was, roughly:
```
CREATE TABLE thing_ref (
    thing_id int not null,
    ref_id int not null,
    primary key (thing_id, ref_id),
    key (ref_id)
);
```
In production, the `thing` table contains 165 million rows and the `thing_ref` table contains 1.43 billion rows (essentially, each thing can have a set of many refs, and we want to answer efficiently how many things have a particular ref).

When explaining the query that gh-ost issued to find the min/max migration values, it became apparent that it was using the index on `ref_id` instead of `thing_id, ref_id` which meant that, instead of being more or less instantaneous, mysql was doing a table scan over 1.43 billion rows to locate the min/max of thing_id, far too long to be acceptable. Reverting our local `optimizer_switch` hack (hence the original patch) restored the speed of the query back to something that was acceptable, and gh-ost was able to complete the migration of the table (taking 27 hours to do so!)

So yeah tl;dr issuing a `force index` should result in the desired behavior.

timvaillancourt:
@cyrinux / @dland the "force index" fix was merged in this PR https://github.com/github/gh-ost/pull/1237

I think this is still a good change however 👍

timvaillancourt:
@cyrinux this fails on 5.7 CI on:
```
2023-12-09 02:26:19 FATAL Error 1231: Variable 'optimizer_switch' can't be set to the value of 'prefer_ordering_index=on'
```

morgo:
As I mentioned above, this was introduced post-GA. So it won't work with
Aurora 2.0. I think this is a bit of a problem because 5.7 is no longer
supported, but Aurora 2.0 is.

PR #1237, opened 2022-12-28
Force index in unique key range queries

timvaillancourt:
### Description

This PR forces the unique-key index to be used when gathering the min/max range of the unique key. This is to avoid possibilities of the optimizer picking a different index than we planned, [discussed in more detail here](https://github.com/github/gh-ost/pull/1236#discussion_r1058031527)


morgo:
LGTM

cyrinux:
LGTM too 👍🏻

cyrinux (approved):

meiji163:
LGTM 🚀

meiji163 (approved):

Issue #1242, opened 2023-01-06
force index in Iteration Range End Value queries to fix execution plans


wangzihuacool:
I recently found that gh-ost occasionally executes slowly in MySQL 8.0.
When copying data to the _gho table, querys for Iteration Range End Values occationally use the wrong index due to inappropriate optimizer selection.

I think we can force the unique-key index to avoid inappropriate optimizer selection, in BuildUniqueKeyRangeEndPreparedQueryViaOffset and BuildUniqueKeyRangeEndPreparedQueryViaTemptable. Just like unique key range queries do in #1237 .

I'd like to contribute the patch, if you agree.

timvaillancourt:
@wangzihuacool thanks for this report. I believe https://github.com/github/gh-ost/pull/1237 will resolve this issue, can you please confirm?

timvaillancourt:
>  in BuildUniqueKeyRangeEndPreparedQueryViaOffset and BuildUniqueKeyRangeEndPreparedQueryViaTemptable

@wangzihuacool sorry I misread your issue summary, this is similar but different as you mentioned 👍 

> I'd like to contribute the patch, if you agree.

A PR would be appreciated thanks!

wangzihuacool:
> 

I have created a PR #1245 , hope you take the time to review it.

timvaillancourt:
@wangzihuacool I believe this is resolved by https://github.com/github/gh-ost/pull/1237. Please test and re-open if this is not the case. Thanks!

PR #1245, opened 2023-01-09
force index in iteration range end value queries

wangzihuacool:
Description

This PR forces the unique-key index to be used in iteration range end value queries when copying data to the _gho table.
Force the unique-key index to avoid inappropriate optimizer selection, as discussed in #1242 .

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -574,7 +574,7 @@ func (this *Applier) CalculateNextIterationRangeEndValues() (hasFurtherRange boo
 		query, explodedArgs, err := buildFunc(
 			this.migrationContext.DatabaseName,
 			this.migrationContext.OriginalTableName,
-			&this.migrationContext.UniqueKey.Columns,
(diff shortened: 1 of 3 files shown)


CommanderRoot:
Probably nitpicking but the spacing / formatting is off on the SQL queries, at least on Github. The other lines use tabs + spaces and you only used spaces.

wangzihuacool:
> Probably nitpicking but the spacing / formatting is off on the SQL queries, at least on Github. The other lines use tabs + spaces and you only used spaces.

thank you for your advise, I'll modify it to keep consistent.

wangzihuacool:
@timvaillancourt  would you please take the time to reivew it?

timvaillancourt:
@wangzihuacool I believe this is resolved by a similar PR that is on `master`: https://github.com/github/gh-ost/pull/1237

Please test and re-open if this is not the case. Thanks!

PR #1246, opened 2023-01-11
Cleanup whitespace and comments in SQL query text

timvaillancourt:
### Description

This PR cleans up SQL query text in several files where whitespace, line separation and the use of comments were inconsistent

The rules I followed were:
1. All queries include a `/* gh-ost */` or `/* gh-ost database.table */` comment
2. Single-tabs (`/t`) are used for indentation
3. Queries are defined as the variable `query :=` using backticks
4. SQL begins on the line following `query` after a single tab
    - Example:
        ```golang
        query := `
                select /* gh-ost */
                ...`
        ```
    - Exception: simple var-only queries without a `FROM`, example: `select /* gh-ost */ @@global.log_slave_updates`
6. SQL is further indented when these terms are encountered:
    - `FROM`/`INTO`
    - `SET`/`ON DUPLICATE KEY UPDATE`
    - `VALUES`
    - `WHERE`
    - `(` or `)` SQL-block brackets, example: a sub-query
    - `INNER JOIN`
    - `ON`
    - `CASE`
7. No newline at end of `query` var / close backtick after last word

What this PR does NOT address is the inconsistency in capitalization in SQL. I would like to leave this to a vote from the community in an up-coming GitHub Discussion, and I will follow-up on this separately


meiji163:
looks nice!

meiji163 (approved):

meiji163 (approved):

go/logic/applier.go
```diff
@@ -1150,7 +1151,7 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
 			return err
 		}
 
-		sessionQuery := "SET SESSION time_zone = '+00:00'"
+		sessionQuery := "SET /* gh-ost */ SESSION time_zone = '+00:00'"
```
shaohk, line 1154: @timvaillancourt Does the time zone set here need to be configured as this.migrationContext.ApplierTimeZone? I noticed that other sessions are using this.migrationContext.ApplierTimeZone for setting the time zone, but here it is set to +00:00. Should this be adjusted?

Issue #887, opened 2020-10-13
When the _xxxx_gho table is queried, data may be lost.


shatianjiang:
> gh-ost Refer to this document for switching order ：https://github.com/github/gh-ost/issues/82

Suppose now operating DDL on table `a`, three temporary tables `_a_del`, `_a_gho`, and`_a_ghc` will be generated. 

rename operation is:  **rename `a` to `_a_del`, `_a_gho` to `a`,**
lock table statement is: **lock table `a` write, `_a_del` write**
According to the order of acquiring the lock: 
> first acquire the `_a_del` table lock, 
> then `_a_gho` table lock , 
> and finally the `a` table lock. 

![knk_2020-10-13  12-14-13](https://user-images.githubusercontent.com/39261302/95815590-90550300-0d4f-11eb-8c78-465d66e6f62f.jpg)

Thank you!

shlomi-noach:
There seems to be a corruption in the original comment. On my mobile GitHub app, I can actually see what the question is. Copy+pasting here (as is, including typos):

> If there is a query transaction on the _a_gho table lock that causes the acquisition if the lock to be delayed, other threads may be before rename First obtain the lock of table a, write the data in table, and then perform table rename operation, at this time the data is lost

This situation does not happen. First, gh-ost locks the original table, then, it _waits_ until the `_gho` table has received all the writes from the original table up to the lock. From that point, there can be no mor ewrites on `_gho` table. `gh-ost` then proceeds to lock `_gho` table.

wukongHH:
How to repeat :

session 1 | session 2 | session 3 | session 4
-- | -- | -- | --
lock table a write, _a_del write; |   |   |  
  | insert into a values (1); |   |  
  |   |   | begin;select * from _a_gho;
  |   | rename table a to _a_del,_a_gho to a; |  
drop table _a_del;unlock tables; |   |   |  
  |   |   | commit;

wukongHH:
https://dev.mysql.com/doc/refman/8.0/en/metadata-locking.html

DDL statements, LOCK TABLES, and other similar statements try to reduce the number of possible deadlocks between concurrent DDL statements by acquiring locks on explicitly named tables in name order.

shatianjiang:
> There seems to be a corruption in the original comment. On my mobile GitHub app, I can actually see what the question is. Copy+pasting here (as is, including typos):
> 
[...]

PR #1269, opened 2023-04-20
fix issue 887

cenkore:
Related issue: https://github.com/github/gh-ost/issues/887

### Description

This PR adds a detection of whether rename session is acquiring the lock of the original table, to avoid data loss caused by directly unlock tables after drop magic table while rename session is still acquiring the lock of the ghost table in the atomic cut-over phase. The detection method via enable the instrument [wait/lock/metadata/sql/mdl](https://dev.mysql.com/doc/refman/5.7/en/performance-schema-metadata-locks-table.html).

Here add a parameter `allow-setup-metadata-lock-instruments` to enable wait/lock/metadata/sql/mdl instrument. In atomic cut-over phase, add two channels `okToDropSentryTable`, `dropSentryTableDone`,   the channel `okToUnlockTable` only do one thing that releasing the locks,  after `migrator` receiving `dropSentryTableDone` channel, then detect whether rename session is ready (acquire original table lock) or not.  If not, you will get a message like the follows, __Expect rename session xxxxx acquiring table metadata lock is `schema`.`table`, but got `schema`.`_table_gho`__,  then cancel the current cut-over, and try again. 

Thank you!

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -232,6 +232,8 @@ type MigrationContext struct {
 
 	recentBinlogCoordinates mysql.BinlogCoordinates
 
+	AllowSetupMetadataLockInstruments bool
+
 	Log Logger
 }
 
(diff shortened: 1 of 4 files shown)


shlomi-noach:
Can you please rephrase what this PR does? Or, if you could please illustrate by a sequence of events that cause the bug?

shlomi-noach:
Ah, sorry, I missed the link to https://github.com/github/gh-ost/issues/887

blackhat12:
Klinux??

Pada tanggal Sel, 25 Apr 2023 15:57, Shlomi Noach ***@***.***>
menulis:

PR #1322, opened 2023-10-08
Parse binlog timestamp in UTC

meiji163:
### Description

This PR addresses a situation where `TIMESTAMP` fields can be corrupted. Quoting @dm-2:

> gh-ost is incorrectly modifying the timezone of timestamp fields of rows when the following criteria are met:
> - database is running in non-UTC timezone, and
> - row is changed whilst a migration is running, and
> - field is timestamp, and
> - field contains a non-UTC timestamp

This is caused by how `go-mysql/replication` package parses binlog timestamps based on timezone. In the latest release this behaviour was disabled by editing the vendored package https://github.com/github/gh-ost/commit/7e0bdbe72a7f81d7a5546bc74279ade9f79baec5 , but in the master branch it is enabled.

### DIFF
--- a/go/binlog/gomysql_reader.go
+++ b/go/binlog/gomysql_reader.go
@@ -13,6 +13,8 @@ import (
 	"github.com/github/gh-ost/go/mysql"
 	"github.com/github/gh-ost/go/sql"
 
+	"time"
+
 	gomysql "github.com/go-mysql-org/go-mysql/mysql"
 	"github.com/go-mysql-org/go-mysql/replication"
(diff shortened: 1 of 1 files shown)


rashiq (approved):

meiji163:
This has run successfully on testing replicas on 175 tables total, so I am fairly confident to merge.

timvaillancourt:
@meiji163 I'm so glad someone finally figured this out and it's great it didn't require modifying vendored code. Thank you!

Issue #1346, opened 2023-12-07
Add git commit hash to version info


timvaillancourt:
Both the `--version` flag and `starting gh-ost %+v` logs do not include the git hash of the build, which would make it easier for someone to understand exactly what code they're running

This is a feature request to add the git commit hash to these areas of the code

PR #1347, opened 2023-12-07
Include git commit in version

timvaillancourt:

Related issue: https://github.com/github/gh-ost/issues/1346

### Description

This PR adds the git commit hash to the version info `gh-ost` returns using the `--version` flag and in the logs at startup time


meiji163 (approved):

timvaillancourt:
@meiji163 thanks for the review. I made one small change to make the `--version` output change less-breaking in case someone is parsing that output in script, etc

meiji163 (approved):

Issue #1348, opened 2023-12-07
`RELEASE_VERSION` file is stale


timvaillancourt:
The `RELEASE_VERSION` file in the root of the repo has a stale version: `1.1.2` and it has been out-of-sync in the past

It seems like this file isn't actually required as `build.sh` primarily [gathers the release version from `git` here](https://github.com/github/gh-ost/blob/master/build.sh#L51-L53)

PR #1349, opened 2023-12-07
Remove stale `RELEASE_VERSION` file

timvaillancourt:

Resolves issue: https://github.com/github/gh-ost/issues/1348

### Description

This PR removes the `RELEASE_VERSION` file as it is often stale and not really required. I feel it is better to remove this and add it back later if it's truly needed - for now it's just leading to confusion


meiji163 (approved):

qikquack:
Stop

qikquack:
Stop

Issue #1366, opened 2024-01-11
When the binlog file is larger than 4G, data loss occurs


dongwenpeng:
Hi, When the binlog file is larger than 4G, data loss occurs. 

MySQL allows a single binlog file to be larger than 4G, and in the MySQL source code, the binlog event end_log_pos field type is defined as uint32, with a maximum storage of 4G. Therefore, when a single binlog file is larger than 4G, the end_log_pos field storage will overflow and be used cyclically.

Try to use mysqlbinlog to parse a binlog file larger than 4G, example:

```
#202409 20:06:42 server id 100  end_log_pos 4294954865 CRC32 0x5e0195d9         Update_rows: table id 266
# at 4294954865
#202409 20:06:42 server id 100  end_log_pos 4294962881 CRC32 0x2f0b79cc         Update_rows: table id 266
# at 4294962881
#202409 20:06:42 server id 100  end_log_pos 3601 CRC32 0x2f367261       Update_rows: table id 266
# at 4294970897
#202409 20:06:42 server id 100  end_log_pos 11617 CRC32 0xb1ac6949      Update_rows: table id 266
```

When end_log_pos reaches 4294962881, the next event end_log_pos overflows to 3601 and becomes smaller.

### Why data is lost

In the handleRowsEvent method of the gh-ost code, which is used to process a DML event, there is a logic used to judge: if the currently received event end_log_pos value is less than or equal to the last executed event end_log_pos value, the current event will be ignored. When the binlog file is larger than 4G, all events larger than 4G will be discarded due to event end_log_pos overflow storage.

```
// StreamEvents
func (this *GoMySQLReader) handleRowsEvent(ev *replication.BinlogEvent, rowsEvent *replication.RowsEvent, entriesChannel chan<- *BinlogEntry) error {
	if this.currentCoordinates.SmallerThanOrEquals(&this.LastAppliedRowsEventHint) {
		this.migrationContext.Log.Debugf("Skipping handled query at %+v", this.currentCoordinates)
		return nil
	}
        ...
        // Execute event
	...
        // Record the position information of the last executed event
	this.LastAppliedRowsEventHint = this.currentCoordinates
	return nil
}
```

### The reason why binlog is larger than 4G

When there is a big row in the data table, too many chunks at a time may cause a large transaction (assuming the transaction is larger than 4G), or the business itself has a large transaction. A transaction is written to a single binlog file.

### Group commit and binlog

[...]

PR #1367, opened 2024-01-17
Fix binlog end log pos over lost data

shaohk:
Related issue: https://github.com/github/gh-ost/issues/1366

### Description

This PR [Fix binlog end log pos over lost data]


dikang123:
when will this PR be merged?

timvaillancourt:
@shaohk thanks a lot for this PR, this sounds like a difficult problem to debug!

Can you think of any ways that this check can be tested in go unit tests? I wonder if a test could:
1. Test that a normal `*replication.RowsEvent` event will be accepted
2. Test that an oversized `*replication.RowsEvent` event it not accepted

Luckily, `handleRowEvents` doesn't need a real MySQL server as it pushes events to `entriesChannel chan<- *BinlogEntry`, which should make this relatively simple to test

timvaillancourt (changes requested):
Looks good! Added one comment requesting unit tests 🙇

shaohk:
> @shaohk thanks a lot for this PR, this sounds like a difficult problem to debug!
> 
> Can you think of any ways that this check can be tested in go unit tests? I wonder if a test could:
> 
> 1. Test that a normal `*replication.RowsEvent` event will be accepted
> 2. Test that an oversized `*replication.RowsEvent` event it not accepted
> 
> Luckily, `handleRowEvents` doesn't need a real MySQL server as it pushes events to `entriesChannel chan<- *BinlogEntry`, which should make this relatively simple to test

@timvaillancourt Thank you for reviewing this PR. I have added unit tests as requested. Could you please take a look again? Thank you for your hard work.

go/mysql/binlog.go
```diff
@@ -74,3 +75,23 @@ func (this *BinlogCoordinates) SmallerThanOrEquals(other *BinlogCoordinates) boo
 	}
 	return this.LogFile == other.LogFile && this.LogPos == other.LogPos
 }
+
+// IsLogPosOverflowBeyond4Bytes returns true if the coordinate endpos is overflow beyond 4 bytes.
```
timvaillancourt, line 79: @shaohk last request, can we add a reference to why a value over 4 bytes is an overflow?

shaohk: OK, I have added. Please take a look again, Thks.

meiji163:
@shaohk Can you take a look at the failing tests and linter?

shaohk:
> @shaohk Can you take a look at the failing tests and linter?

@meiji163 
OK, Please take a look again, Thks.

meiji163 (approved):

meiji163:
LGTM, can you take another look @timvaillancourt?

timvaillancourt:
> LGTM, can you take another look @timvaillancourt?

@meiji163 / @shaohk: I think erroring here is the easiest next step, so LGTM 👍

But longer term, what do we "want" `gh-ost` to do here? Or to put it differently, what does the MySQL replica do in this scenario? I'm assuming it doesn't error right? 🤔

timvaillancourt (approved):
Approved with longer-term question

meiji163:
I'm not sure how MySQL handles it
I found this bug report from last year https://bugs.mysql.com/bug.php?id=112189

shaohk:
> > LGTM, can you take another look @timvaillancourt?
> 
> @meiji163 / @shaohk: I think erroring here is the easiest next step, so LGTM 👍
> 
> But longer term, what do we "want" `gh-ost` to do here? Or to put it differently, what does the MySQL replica do in this scenario? I'm assuming it doesn't error right? 🤔

https://dev.mysql.com/doc/refman/8.0/en/replication-options-binary-log.html#sysvar_max_binlog_cache_size.

`When [gtid_mode](https://dev.mysql.com/doc/refman/8.0/en/replication-options-gtids.html#sysvar_gtid_mode) is not ON, the maximum recommended value is 4GB, due to the fact that, in this case, MySQL cannot work with binary log positions greater than 4GB; when gtid_mode is ON, this limitation does not apply, and the server can work with binary log positions of arbitrary size.`

So I believe that MySQL master-slave replication is based on the GTID mode, so there is no problem with MySQL master-slave replication. The gh-ost tool for synchronizing binlogs uses a non-GTID mode, which is why this issue arises.

Issue #1373, opened 2024-01-31
Status message is not properly formatted and contains MISSING on STDERR


icyflame:
We recently upgraded from gh-ost v1.0.48 to v1.1.6. Starting in this version, we started seeing the string `MISSING` in our status messages:

```
2024-01-31 12:49:54 INFO Copy: 0/0 100.0%!;(MISSING) Applied: 0; Backlog: 0/1000; Time: 0s(total), 0s(copy); streamer: mysql_bin.000002:79674; Lag: 0.00s, HeartbeatLag: 0.02s, State: migrating; ETA: due
```

The string `MISSING` is generally shown when `fmt.Sprintf` is used with a formatting directive, but the appropriate variable value to replace the formatting directive is missing.

Using `git bisect`, I found that this particular issue was introduced in the commit https://github.com/github/gh-ost/commit/515aa72d3d9b756e454b0168b4e57bc599b45e36 as part of the PR https://github.com/github/gh-ost/pull/1194. The migration context logger prints information that is being printed to STDOUT to STDERR as well, with the local time as a prefix. The motivation for this is unclear from the PR's description.

> If this is a bug report, please provide a test case (e.g., your table definition and gh-ost command) and the error output.

Table definition:

```sql
root@localhost [test] > show create table users \G
*************************** 1. row ***************************
       Table: users
Create Table: CREATE TABLE `users` (
  `id` int(10) unsigned NOT NULL AUTO_INCREMENT,
  `name` varchar(32) NOT NULL,
  `email` varchar(255) DEFAULT NULL,
  PRIMARY KEY (`id`)
) ENGINE=InnoDB AUTO_INCREMENT=101 DEFAULT CHARSET=utf8
1 row in set (0.00 sec)
```

gh-ost command:

```sh
$ ./gh-ost --host=127.0.0.1 --conf=/gh-ost.cnf --database=test --table=users --alter="engine=innodb" --chunk-size=2000 --max-load=Threads_connected=20 --initially-drop-ghost-table --initially-drop-old-table --test-on-replica --switch-to-rbr --verbose
```

icyflame:
This issue was fixed after the above PR was merged ✅

PR #1194, opened 2022-10-27
Print status to migrationContext logger

SocalNick:
**NOTE:** This is a simple alternative to #1193 

## A Pull Request should be associated with an Issue.

> We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
> If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
> and potentially we'll be able to point development in a particular direction.

Related issue: https://github.com/github/gh-ost/issues/1026

> Further notes in https://github.com/github/gh-ost/blob/master/.github/CONTRIBUTING.md
> Thank you! We are open to PRs, but please understand if for technical reasons we are unable to accept each and any PR

### Description

Status is currently only sent to stdout. We'd like the status to also be sent to the migrationContext logger.

> In case this PR introduced Go code changes:

- [x] contributed code is using same conventions as original code
- [x] `script/cibuild` returns with no formatting errors, build errors or unit test errors.

PR #1374, opened 2024-01-31
Print status to migration context logging after escaping % character

icyflame:
### Description

The string `status` has the character `%` after the progress percentage number. When printed once again using the `Infof` function, this `%` is read as a directive for formatting and ends up showing `MISSING` in the output as shown below:

	2024-01-31 12:49:54 INFO Copy: 0/0 100.0%!;(MISSING) Applied: 0; Backlog: 0/1000; Time: 0s(total), 0s(copy); streamer: mysql_bin.000002:79674; Lag: 0.00s, HeartbeatLag: 0.02s, State: migrating; ETA: due

Simply changing `Infof` to `Info` does not work here because `this.migrationContext.Log` is an object of the type `DefaultLogger` in this repository, which uses the library https://github.com/outbrain/golib/log. That library uses formatting __even__ when the `Info`function is called:
https://github.com/outbrain/golib/blob/2531e5dbcc71b6f8a4ccf1205c209ae89b7529fc/log/log.go#L191-L193

If this patch is not acceptible, then I can also just remove this line. The migration context logger prints the messages which are already printed to STDOUT once again to STDERR with the current time as a prefix. This change was introduced in the commit https://github.com/github/gh-ost/blob/515aa72d3d9b756e454b0168b4e57bc599b45e36/go/logic/migrator.go#L1039, introduced in the PR https://github.com/github/gh-ost/pull/1194.

Related issue: https://github.com/github/gh-ost/issues/1373

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1042,7 +1042,14 @@ func (this *Migrator) printStatus(rule PrintStatusRule, writers ...io.Writer) {
 	)
 	w := io.MultiWriter(writers...)
 	fmt.Fprintln(w, status)
-	this.migrationContext.Log.Infof(status)
+
+	// This "hack" is required here because the underlying logging library
+	// github.com/outbrain/golib/log provides two functions Info and Infof; but the arguments of
+	// both these functions are eventually redirected to the same function, which internally calls
+	// fmt.Sprintf. So, the argument of every function called on the DefaultLogger object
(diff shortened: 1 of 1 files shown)


meiji163 (approved):

timvaillancourt:
Awesome thanks! This reminds me we should move to a modern logger at some point

PR #1378, opened 2024-02-05
feat binlog apply optimization

shaohk:
### Description

1. Support for ignoring binlog events that exceed chunk boundary values.
If the value corresponding to the unique key of the binlog exceeds the maximum value of chunk iteration and is less than the maximum boundary value of the copy, it is ignored.

2. Support for binlog merge processing.
When the columns of the unique key selected by chunk are int or float, the binlog is processed by map merging, and all delete operations are merged into one `delete sql`.All insert and update operations are merged into one `replace sql`. Then execute these sql in db transaction.

### DIFF
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -129,6 +129,10 @@ Noteworthy is that setting `--dml-batch-size` to higher value _does not_ mean `g
 When binlog unique key value is over `MigrationIterationRangeMaxValues`, and less than `MigrationRangeMaxValues`, the binlog will be ignored. Because the data will be synced by copy chunk.
 When binlog unique key value is over `MigrationRangeMaxValues` or less than `MigrationIterationRangeMaxValues`, the binlog will be applied.
 
+### is-merge-dml-event
+When is-merge-dml-event is `true`, and the chunk unique key is int or float type. the sync binlog event while be merged into map, and the key is unique key value.
+Then traverse the map, merge all delete operations into one `delete sql`, merge all insert and update operations into `replace sql`, and then execute.
+
 ### exact-rowcount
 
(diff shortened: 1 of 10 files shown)


dnovitski:
## Correctness Issues Found + Fix PR

I reviewed this PR in detail and found several correctness bugs in the merge DML implementation. I've submitted a fix as PR #1687 (built on top of this branch).

### Critical Issues

1. **`generateReplaceQuery` reads live source state** — `REPLACE INTO ghost SELECT ... FROM source WHERE pk IN (...)` reads the **current** source row, not the binlog event's row image. This is incorrect when:
   - Events are buffered during throttle (source may have changed since event time)
   - gh-ost is killed and restarted (replay uses current state, not original)
   - Cutover happens before all events drain

2. **`isIgnoreOverMaxChunkRangeEvent` returns `true` when nil** — Before the first row-copy chunk, `MigrationIterationRangeMaxValues` is nil, causing the comparison function to return `true` (ignore event). Early events are silently dropped.

3. **Missing column conversion tokens** — If using plain `?` placeholders, timezone (`convert_tz`), enum (`ELT`), and JSON (`convert(? using utf8mb4)`) columns would be inserted without their required conversions.

### Fix (PR #1687)

- Uses `REPLACE INTO ghost (cols) VALUES (?, ?)` with binlog after-image values
- Handles column conversions via `BuildColumnsPreparedValues`  
- Fixes nil check to return `false` (don't ignore)
- Introduces `mergedEntry` struct for safe map value handling

### Performance Results

The fix maintains the performance benefits:
- 10% faster general workload, 39% faster hot-spot dedup, 2.25× sustained throughput vs non-merge mode

See #1687 for full details and benchmarks.

PR #1383, opened 2024-03-02
[parser] support create-table flag

najisawas:
### Description

This PR adds a new flag called "--create-table", an alternative to the "--alter" flag. With "--create-table" you can redefine the table that you want gh-ost to migrate to, effectively supporting what would have required multiple ALTERs.

For example, If you need to restructure the secondary indexes on a MySQL table, you would need to do the following:

_Assuming a and b are already columns in the table_...

- ADD COLUMN `x`INT, ADD COLUMN `y` VARCHAR(24),
   ADD INDEX `a_x_idx` (`a`, `x`),
   ADD INDEX `y_idx` (y),
   DROP INDEX `a_b_idx`

Now you can simply provide a "--create-table" flag like so
```
--create-table="( ..., x INT, y VARCHAR(24), ... KEY a_x_idx (a,x), KEY y_idx (y) ...)"
```

The PR is a bit large as it includes some refactors to maintain code cleanliness. I can separate out some of the refactors into separate PRs for ease of review

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -77,10 +77,12 @@ func NewThrottleCheckResult(throttle bool, reason string, reasonHint ThrottleRea
 type MigrationContext struct {
 	Uuid string
 
-	DatabaseName          string
-	OriginalTableName     string
-	AlterStatement        string
-	AlterStatementOptions string // anything following the 'ALTER TABLE [schema.]table' from AlterStatement
+	DatabaseName             string
+	OriginalTableName        string
+	AlterStatement           string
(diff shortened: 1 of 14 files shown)


najisawas:
#1384

najisawas:
tests still need to be expanded, but wanted to publish an initial revision for visibility

meiji163:
I hesitate to continue expanding the regex-based SQL parsing. If you need to compute the `ALTER` statement from two `CREATE TABLE` statements I recommend using [schemadiff](https://github.com/planetscale/schemadiff) or [skeema](https://www.skeema.io/)

PR #1398, opened 2024-03-15
Feat parallel chunk data

shaohk:
### Description

This PR Supports concurrent copying of chunk blocks during each time slice of copying the existing data.


shaohk:
MySQL insert ... select ... does not permit concurrent.

https://dev.mysql.com/doc/refman/5.7/en/insert-select.html

shaohk:
When the unique index used for copying blocks includes an auto-increment column, concurrency is blocked on the auto-inc lock. If the unique index used does not include an auto-increment column, concurrency can proceed.

xiaoxuanfeng:
May I ask when this feature will be reflected in the new version

PR #1406, opened 2024-04-11
Refine `wait_timeout` override to be at cut-over only

timvaillancourt:
Related issue: #1407 

### Description

This PR refines https://github.com/github/gh-ost/pull/1401 by overriding the session [`wait_timeout`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) only where it is needed - at the cut-over time where an idle connection could lead to potentially-long table lock if the `gh-ost` process (or host running it) "freezes"/"stalls" at the cut-over stage

The change (at cut-over only):
1. Before cut-over:
    - The server [`wait_timeout`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) is fetched _(via an existing `select` that fetched `time_zone`)_
    - The applier session `wait_timeout` is set to be 3 x the lock-wait timeout
        - This is to ensure the lock is released if `gh-ost` stalls with a still-active connection here
4. The cut-over proceeds as normal
3. After cut-over, the original session [`wait_timeout`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) is restored to what it was set to pre-cut-over

The `--mysql-wait-timeout` flag added in #1401 is removed because it is no longer needed. No release has been cut since #1401, so this isn't necessarily a breaking change

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -164,6 +164,7 @@ type MigrationContext struct {
 	Hostname                               string
 	AssumeMasterHostname                   string
 	ApplierTimeZone                        string
+	ApplierWaitTimeout                     int64
 	TableEngine                            string
 	RowsEstimate                           int64
 	RowsDeltaEstimate                      int64
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -51,7 +51,6 @@ func main() {
 	flag.StringVar(&migrationContext.AssumeMasterHostname, "assume-master-host", "", "(optional) explicitly tell gh-ost the identity of the master. Format: some.host.com[:port] This is useful in master-master setups where you wish to pick an explicit master, or in a tungsten-replicator where gh-ost is unable to determine the master")
 	flag.IntVar(&migrationContext.InspectorConnectionConfig.Key.Port, "port", 3306, "MySQL port (preferably a replica, not the master)")
 	flag.Float64Var(&migrationContext.InspectorConnectionConfig.Timeout, "mysql-timeout", 0.0, "Connect, read and write timeout for MySQL")
-	flag.Float64Var(&migrationContext.InspectorConnectionConfig.WaitTimeout, "mysql-wait-timeout", 0.0, "wait_timeout for MySQL sessions")
 	flag.StringVar(&migrationContext.CliUser, "user", "", "MySQL user")
 	flag.StringVar(&migrationContext.CliPassword, "password", "", "MySQL password")
 	flag.StringVar(&migrationContext.CliMasterUser, "master-user", "", "MySQL user on master, if different from that on replica. Requires --assume-master-host")
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -89,7 +89,7 @@ func (this *Applier) InitDBConnections() (err error) {
 		return err
 	}
 	this.migrationContext.ApplierMySQLVersion = version
-	if err := this.validateAndReadTimeZone(); err != nil {
+	if err := this.validateAndReadGlobalVariables(); err != nil {
 		return err
 	}
 	if !this.migrationContext.AliyunRDS && !this.migrationContext.GoogleCloudPlatform && !this.migrationContext.AzureMySQL {
@@ -106,10 +106,13 @@ func (this *Applier) InitDBConnections() (err error) {
 	return nil
 }
 
-// validateAndReadTimeZone potentially reads server time-zone
-func (this *Applier) validateAndReadTimeZone() error {
-	query := `select /* gh-ost */ @@global.time_zone`
-	if err := this.db.QueryRow(query).Scan(&this.migrationContext.ApplierTimeZone); err != nil {
+// validateAndReadGlobalVariables potentially reads server global variables, such as the time_zone and wait_timeout.
+func (this *Applier) validateAndReadGlobalVariables() error {
+	query := `select /* gh-ost */ @@global.time_zone, @@global.wait_timeout`
+	if err := this.db.QueryRow(query).Scan(
+		&this.migrationContext.ApplierTimeZone,
+		&this.migrationContext.ApplierWaitTimeout,
+	); err != nil {
 		return err
 	}
 
@@ -934,6 +937,27 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
 	return nil
 }
 
+// InitAtomicCutOverWaitTimeout sets the cut-over session wait_timeout in order to reduce the
+// time an unresponsive (but still connected) gh-ost process can hold the cut-over lock.
+func (this *Applier) InitAtomicCutOverWaitTimeout(tx *gosql.Tx) error {
+	cutOverWaitTimeoutSeconds := this.migrationContext.CutOverLockTimeoutSeconds * 3
+	this.migrationContext.Log.Infof("Setting cut-over idle timeout as %d seconds", cutOverWaitTimeoutSeconds)
+	query := fmt.Sprintf(`set /* gh-ost */ session wait_timeout:=%d`, cutOverWaitTimeoutSeconds)
+	_, err := tx.Exec(query)
+	return err
+}
+
+// RevertAtomicCutOverWaitTimeout restores the original wait_timeout for the applier session post-cut-over.
+func (this *Applier) RevertAtomicCutOverWaitTimeout() {
+	this.migrationContext.Log.Infof("Reverting cut-over idle timeout to %d seconds", this.migrationContext.ApplierWaitTimeout)
+	query := fmt.Sprintf(`set /* gh-ost */ session wait_timeout:=%d`, this.migrationContext.ApplierWaitTimeout)
+	if _, err := sqlutils.ExecNoPrepare(this.db, query); err != nil {
+		this.migrationContext.Log.Errorf("Failed to restore applier wait_timeout to %d seconds: %v",
+			this.migrationContext.ApplierWaitTimeout, err,
+		)
+	}
+}
+
 // AtomicCutOverMagicLock
 func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocked chan<- error, okToUnlockTable <-chan bool, tableUnlocked chan<- error) error {
 	tx, err := this.db.Begin()
@@ -978,6 +1002,12 @@ func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocke
 		tableLocked <- err
 		return err
 	}
+
+	if err := this.InitAtomicCutOverWaitTimeout(tx); err != nil {
+		tableLocked <- err
+		return err
+	}
+	defer this.RevertAtomicCutOverWaitTimeout()
 
 	query = fmt.Sprintf(`lock /* gh-ost */ tables %s.%s write, %s.%s write`,
 		sql.EscapeName(this.migrationContext.DatabaseName),
--- a/doc/command-line-flags.md
+++ b/doc/command-line-flags.md
@@ -202,10 +202,6 @@ List of metrics and threshold values; topping the threshold of any will cause th
 
 Typically `gh-ost` is used to migrate tables on a master. If you wish to only perform the migration in full on a replica, connect `gh-ost` to said replica and pass `--migrate-on-replica`. `gh-ost` will briefly connect to the master but otherwise will make no changes on the master. Migration will be fully executed on the replica, while making sure to maintain a small replication lag.
 
-### mysql-wait-timeout
-
-If set to a value greater than zero, causes `gh-ost` to set a provided [MySQL `wait_timeout`](https://dev.mysql.com/doc/refman/8.0/en/server-system-variables.html#sysvar_wait_timeout) for MySQL sessions opened by `gh-ost`, specified in seconds.
-
 ### postpone-cut-over-flag-file
 
 Indicate a file name, such that the final [cut-over](cut-over.md) step does not take place as long as the file exists.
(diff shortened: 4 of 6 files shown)


go/base/context.go
```diff
@@ -135,6 +135,7 @@ type MigrationContext struct {
 	CriticalLoadHibernateSeconds        int64
 	PostponeCutOverFlagFile             string
 	CutOverLockTimeoutSeconds           int64
+	CutOverIdleTimeoutSeconds           int64
```
shlomi-noach, line 138: Since this value must be greater than `CutOverLockTimeoutSeconds` (or else cut-over may timeout unjustifiably), I suggest removing it altogether and setting something like `CutOverLockTimeoutSeconds * 2`.

go/cmd/gh-ost/main.go
```diff
@@ -51,7 +51,6 @@ func main() {
 	flag.StringVar(&migrationContext.AssumeMasterHostname, "assume-master-host", "", "(optional) explicitly tell gh-ost the identity of the master. Format: some.host.com[:port] This is useful in master-master setups where you wish to pick an explicit master, or in a tungsten-replicator where gh-ost is unable to determine the master")
 	flag.IntVar(&migrationContext.InspectorConnectionConfig.Key.Port, "port", 3306, "MySQL port (preferably a replica, not the master)")
 	flag.Float64Var(&migrationContext.InspectorConnectionConfig.Timeout, "mysql-timeout", 0.0, "Connect, read and write timeout for MySQL")
-	flag.Float64Var(&migrationContext.InspectorConnectionConfig.WaitTimeout, "mysql-wait-timeout", 0.0, "wait_timeout for MySQL sessions")
```
shlomi-noach, line 54: Removing a flag makes this PR backwards incompatible, and will break `gh-ost` for any user supplying the flag.

go/logic/applier.go
```diff
@@ -934,6 +937,42 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
 	return nil
 }
 
+// WaitForAtomicCutOverRename waits for the cut-over RENAME operation while periodically
+// pinging the applier connection to avoid the connection becoming idle.
+func (this *Applier) WaitForAtomicCutOverRename(tx *gosql.Tx, okToUnlockTable <-chan bool) {
```
shlomi-noach, line 942: I'm not sure this function is required, and it adds more complexity. If we set `@@session.wait_timeout` to be somewhat bigger than lock-wait-timeout, then we don't need to do a connection keep-alive. I see the idea behind this, but I think it's simpler to just set an appropriate `wait_timeout` and let MySQL handle it.

go/logic/applier.go
```diff
@@ -934,6 +937,42 @@ func (this *Applier) CreateAtomicCutOverSentryTable() error {
...
+	}
+}
+
+// InitAtomicCutOverIdleTimeout sets the cut-over session wait_timeout, returning a func() to rollback
+// to the original wait_timeout and an error, if any.
+func (this *Applier) InitAtomicCutOverIdleTimeout(tx *gosql.Tx) (restoreTimeoutFunc func(), err error) {
+	this.migrationContext.Log.Infof("Setting cut-over idle timeout as %d seconds", this.migrationContext.CutOverIdleTimeoutSeconds)
+	query := fmt.Sprintf(`set /* gh-ost */ session wait_timeout:=%d`, this.migrationContext.CutOverIdleTimeoutSeconds)
+	_, err = tx.Exec(query)
+	return func() {
```
shlomi-noach, line 965: The generated function doesn't use anything special coming from `InitAtomicCutOverIdleTimeout`, ie no local variables etc. As such, it will be simpler for `InitAtomicCutOverIdleTimeout` to only return `error`, and for the calling code (`AtomicCutOverMagicLock()`, below) to just `defer sqlutils.ExecNoPrepare(this.db, <restore wait time>)`)

go/logic/applier.go
```diff
@@ -974,6 +1013,15 @@ func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocke
 		return err
 	}
 
+	if this.migrationContext.CutOverIdleTimeoutSeconds >= 1 {
+		restoreIdleTimeoutFunc, err := this.InitAtomicCutOverIdleTimeout(tx)
```
shlomi-noach, line 1017: IMO should be replaced with:
```suggestion
		defer func() {
		if _, err := sqlutils.ExecNoPrepare(this.db, query); err != nil {
 			this.migrationContext.Log.Errorf("Failed to restore applier wait_timeout to %d seconds: %v",
 				this.migrationContext.ApplierWaitTimeout, err,
 			)
 		}
 		}
		err := this.InitAtomicCutOverIdleTimeout(tx)
```

shlomi-noach (commented):
Nice work and in particular great analysis of this stalemate situation! This looks generally correct, but see inline comments:

- It break `gh-ost` for existing users
- It adds variables where I think we can do without
- It adds code complexity which we can delegate to MySQL.

go/base/context.go
timvaillancourt: @shlomi-noach a `CutOverLockTimeoutSeconds * 2` approach is what I took initially, and I think it will probably work in practice, but I may have got a bit paranoid about one detail that I hope you can validate

The `CutOverLockTimeoutSeconds` (== `lock_wait_timeout`) tells MySQL how long the lock can be held once it is taken (3 seconds by default), but nothing is controlling how long the "whole operation" takes to issue. So if we use `CutOverLockTimeoutSeconds * 2` as the `wait_timeout`, let's call this `6`, we are assuming the `RENAME` will always take < `6` seconds, `3` seconds max locking, `3` seconds max in "processing"/network time

So what if, for whatever crazy reason (we are in this kind of territory now with this PR 😄) the server (or network) is slow to process the `RENAME` operation in the 2nd connection. The connection holding the table locks (with `wait_timeout=6`) may close at `6` seconds and in some extremely-rare edge cases this may be too early. Can you help validate this idea?

This paranoia led me to make the `wait_timeout` configurable (still doesn't solve every case) and to make sure the "waiting" connection keeps it's connection "warm" - in case the `RENAME` is slow

go/cmd/gh-ost/main.go
timvaillancourt: @shlomi-noach this flag was introduced just a few days prior in #1401 to solve the same root cause and no release has been cut yet with it

I later decided to isolate the `wait_timeout` change to cut-over time only (in this PR) to address the problem more closely and to avoid a larger change

That said, I think we could keep this flag around for more generic reasons

go/logic/applier.go
timvaillancourt: @shlomi-noach see earlier comment. This is born by paranoia that the `RENAME` in the other session could take longer than `lock_wait_timeout`, which I believe is possible (eg: server is slow, network, process scheduling, hand-wavy things, etc)

To be fair, these edge cases are extremely "edge"

go/logic/applier.go
timvaillancourt: @shlomi-noach makes sense 👍. I added a few questions to your responses; I'll make this change once we've reached agreement. Thanks for checking 🙇

timvaillancourt:
@shlomi-noach I'm curious about your feedback on one consequence of this change

Following this PR, it is possible for the session holding the `lock tables` to timeout before [the "magic table" is dropped here](https://github.com/github/gh-ost/blob/master/go/logic/applier.go#L1010-L1021)

If understand this right, the lock-session hitting `wait_timeout` will cause locks to be released, which allows the `rename tables` to succeed. That all sounds better than before, but the "magic table" will be left behind. The impact of this I don't fully understand

```bash
-initially-drop-old-table
    	Drop a possibly existing OLD table (remains from a previous run?) before beginning operation. Default is to panic and abort if such table exists
```
Would `gh-ost` just-fix this scenario for users with `-initially-drop-old-table` 🤔? 🙇

go/base/context.go
Milisya: كاني نايك امك

go/cmd/gh-ost/main.go
shlomi-noach: Ah, cool, I did not realize the flag was only added recently. Remove it, then!

go/logic/applier.go
shlomi-noach: I'd say I still think this function adds complexity, and still better to use `@@wait_timeout` based on some (same) formula such as the one presented above.

shlomi-noach:
> If understand this right, the lock-session hitting wait_timeout will cause the rename tables to succeed. 

No, actually. The `RENAME` will not succeed, because the magic table is still in place. The `RENAME` statement attempts to  rename original-table into magic-table. But since magic-table is there, the `RENAME` will fail.

The next cut-over attempt will first, before placing any locks, attempt to `DropAtomicCutOverSentryTableIfExists()` before re-creating it.

This should be safe.

Fe1777:
Certainly! It appears you are working on a GitHub issue related to the `gh-ost` repository. The task involves refining a `wait_timeout` override to ensure it only applies during the cut-over phase. The issue and relevant branch details are:

- Issue: #1406
- Branch: `mysql-wait-timeout-applier-only`
- Base: `master`

Here are the steps to address this issue and submit a pull request:

### Step-by-Step Guide

#### 1. Clone the Repository

If you haven’t already cloned the repository, you can do so with:

```bash
git clone https://github.com/github/gh-ost.git
cd gh-ost
```

#### 2. Create a New Branch

Create a new branch based on `master`:

```bash
git checkout -b refine-wait-timeout-cutover
```

#### 3. Make the Required Changes

Update the code to refine the `wait_timeout` override so it only applies during the cut-over phase. The specific changes depend on where and how the `wait_timeout` is being overridden.

##### Example Changes
This example assumes you are modifying the MySQL connection settings used in the cut-over phase.

1. Locate the code where `wait_timeout` is being set.
2. Ensure that the `wait_timeout` changes only take effect during the cut-over phase.

Here is a pseudo-code representation of what you might need to do:

```go
func OverrideWaitTimeoutDuringCutover(connection *sql.DB) {
    // Set wait_timeout specifically during the cut-over process
    _, err := connection.Exec("SET SESSION wait_timeout=3600")
    if err != nil {
        log.Fatalf("Failed to set wait_timeout: %v", err)
    }
}

// Ensure this function is called only during the cut-over phase
func PerformCutover() {
    connection := establishDBConnection() // Hypothetical function to establish a DB connection 
    OverrideWaitTimeoutDuringCutover(connection)

    // Proceed with the cut-over operations ...
}
```

#### 4. Save and Test Your Changes

Make sure to save your changes and properly test them. Testing is crucial to ensure that your changes work as expected and do not introduce any bugs.

#### 5. Commit Your Changes

Once you’re satisfied with your changes, commit them:

```bash
git add .
git commit -m "Refine wait_timeout override to only apply during cut-over phase"
```

#### 6. Push the Changes to GitHub

Push your changes to the newly created branch:

```bash
git push origin refine-wait-timeout-cutover
```

#### 7. Create a Pull Request

Now, create a pull request on GitHub from your branch (`refine-wait-timeout-cutover`) to the `master` branch.

1. Navigate to the repository on GitHub.
2. You should see a prompt to create a pull request for your recently pushed branch.
3. Click `Compare & pull request`.
4. Ensure the base branch is `master` and the compare branch is `refine-wait-timeout-cutover`.
5. Provide a descriptive title and detailed comment regarding what this pull request does.

### Pull Request Example

**Title**: Refine wait_timeout override to only apply during cut-over phase

**Description**:
This pull request addresses issue #1406. It refines the `wait_timeout` override logic to ensure that it only applies during the cut-over phase, thereby reducing the risk of unintentional timeouts in other phases.

### Review and Merge

After creating the pull request, it will need to be reviewed and approved by other collaborators. Address any feedback you receive.

Once approved, merge the pull request into `master`.

If you have the required permissions, you can merge the pull request through the GitHub interface by clicking `Merge pull request`.

Alternatively, you can merge via the command line:

```bash
git checkout master
git pull origin master
git merge refine-wait-timeout-cutover
```

After merging, you can delete your feature branch:

```bash
git branch -d refine-wait-timeout-cutover
git push origin --delete refine-wait-timeout-cutover
```

By following these steps, you will have successfully refined the `wait_timeout` override to ensure it only applies during the cut-over phase in the `gh-ost` repository. If further assistance is needed, feel free to ask!

go/base/context.go
timvaillancourt: @shlomi-noach I agree the simpler approach would be better, at least for now. Thanks for validating this highly-theoretical risk! I'll make this update

timvaillancourt:
> > If understand this right, the lock-session hitting wait_timeout will cause the rename tables to succeed.
> 
> No, actually. The `RENAME` will not succeed, because the magic table is still in place. The `RENAME` statement attempts to rename original-table into magic-table. But since magic-table is there, the `RENAME` will fail.
> 
> The next cut-over attempt will first, before placing any locks, attempt to `DropAtomicCutOverSentryTableIfExists()` before re-creating it.
> 
> This should be safe.

@shlomi-noach that makes sense (eventually)! Thanks for the validations and explanations

meiji163 (dismissed):

go/base/context.go
timvaillancourt: Removed `CutOverIdleTimeoutSeconds` 👍

go/logic/applier.go
timvaillancourt: Makes sense. Updated 👍

go/logic/applier.go
timvaillancourt: Updated 👍

timvaillancourt:
@shlomi-noach / @meiji163: I believe I've addressed the PR suggestions and this is ready for another review 🙇

meiji163 (approved):

timvaillancourt:
Merging. @shlomi-noach let me know if I missed something and I'll make a follow-up PR 👍

Issue #1418, opened 2024-05-21
collect table statistics before renaming table


wangzihuacool:
I encountered an extreme case last week. After alter table using gh-ost, the table statistics was not updated until half an hour later, causing a lot of slow queries. 
The root cause of this issue is that the statistical event for this table was lost while being placed in the asynchronous queue waiting waiting to be consumed by background statistics update thread. 

This may be an extreme case, but since the collection of statististics is performed asynchronously in the background, it is very likely that there will be situations where the statistics are not updated for a period of time after gh-ost modification, especially when there are a large number of tables in the database.

I wrote a script to test gh-ost schema changes thousands of times, and it often happens that the table's statistics were updated 2-3 seconds after gh-ost renames the table. I'm thinking if gh-ost could collect statistics once before renaming table, in order to avoid the issue of statistics not being updated in time after the schema change.

part of test info below:
`*************************** 1. row ***************************
                          table_name: sbtest1

        table_stats_last_update_time: 2024-05-19 02:07:06

        table_stats_last_update_rows: 134208

            gh_ost_table_create_time: 2024-05-19 10:47:47

            gh_ost_table_rename_time: 2024-05-19 10:47:56

                  gh_ost_finish_time: 2024-05-19 10:47:47

table_stats_after_gh_ost_finish_rows: 0

       table_stats_final_update_time: 2024-05-19 10:47:58
 
              table_stats_final_rows: 134208`

VandhanaSelvaprakash-at:
Hi, following up on this, since we've hit this exact issue at Airtable, and have deployed a fix in our fork.

Our version adds the opt-in flag (--analyze-ghost-table-before-cutover, default off) and the docs, and addresses two things we hit in practice:
- it runs the ANALYZE after the cut-over postpone gate releases, so a postponed cut-over doesn't re-stale its statistics before the swap
- a failed ANALYZE aborts the migration (fatal) rather than swapping in a table with stale stats — ANALYZE TABLE surfaces table-level failures as Msg_type=Error rows while the statement itself succeeds, so we inspect the result rows and fail closed.
- It's flag-gated, so the partition-count/replication-lag concern @shaohk raised only applies to opt-in users. 

Happy to open a PR that supersedes [this one](https://github.com/github/gh-ost/pull/1419) and credits @wangzihuacool as the original author — @timvaillancourt does that work for you?

ericyan:
[...]

PR #1419, opened 2024-05-23
analyze table before cutover

wangzihuacool:
Related issue: #1418 

Description:
analyze ghost table before cutover to avoid table statistics not timely updated.

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -1196,6 +1196,23 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent)) 
 	return nil
 }
 
+// AnalyzeTable actively analyze table to ensure that the ghost table's statistics are timely updated
+func (this *Applier) AnalyzeTable() {
+	query := fmt.Sprintf(`analyze table /* gh-ost */ %s.%s`,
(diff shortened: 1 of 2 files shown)


meiji163:
LGTM

meiji163 (approved):

shaohk:
When the modified table is a partition table, the execution time of the ANALYZE command will increase as the number of partitions increases. When the ANALYZE statement is synchronized to the replica, it may cause replication lag.

meiji163:
> When the ANALYZE statement is synchronized to the replica, it may cause replication lag.

Maybe make this feature a optional flag then? What do you think @timvaillancourt ?

timvaillancourt (approved):
👍

timvaillancourt:
> When the modified table is a partition table, the execution time of the ANALYZE command will increase as the number of partitions increases. When the ANALYZE statement is synchronized to the replica, it may cause replication lag.

@meiji163 considering this (I didn't think of it) I suggest this is an optional flag for now and we can have opt-in users report on performance

@wangzihuacool can you add an optional flag and `doc/command-line-flags.md` update for this one?

Zorgon66:
> 👍

Is it all about me? Harun

PR #715, opened 2019-02-17
MySQL 8.0.13: rename table

shlomi-noach:
Solves https://github.com/github/gh-ost/issues/714

Possibly solves #687 

This PR utilizes `8.0.13` `RENAME TABLE` which supports renaming while holding `WRITE` lock on table. This allows us to use simpler cut-over logic.

PR #1434, opened 2024-07-19
refactor cutover for MySQL 8.x rename feature && support OceanBase

### LINKED ISSUE #1427: skip delete events of changelog table in binlog listener
We are building a system for OceanBase database that is compatible with MySQL's SQL protocol and Replication protocol, called OceanBase Binlog Service. For some reason, an upsert event (on duplicated key update) in this system is converted into a delete event and an insert event in the binlog. In order to make it work without affecting the original logic, I want to check and skip the delete event of the changelog in the onChangelogEvent function.

https://github.com/github/gh-ost/blob/59db6fab40966c2c491de935e6a71b8c23aafa86/go/logic/migrator.go#L202

### LINKED ISSUE #1433: Add atomic cut over function that renames locked tables in the same session
https://github.com/github/gh-ost/pull/715 once tried to introduce a simpler cut over logic, but it was not merged. Now our team is developing a MySQL compatible binlog service of OceanBase database called OceanBase Binlog Service, which also supports the renaming locked table feature that was introduced to MySQL from 8.0.13. But the OceanBase database does not support parallel DDL now, that is, the rename process in the original atomic cut over function cannot be executed normally. Therefore, in order to support OceanBase Binlog Service, I want to introduce the new cut over function again.

I want to add a function named `atomicCutOverRenameWithLock`, which will lock the origin table and the ghost table firstly, and rename these two tables in the same session. This function should be called when connecting to OceanBase or MySQL version 8.0.13 or later.

whhe:

Close #1433  Close #1427

### Description

This PR is based on #715 by @shlomi-noach. There is a new option `--oceanbase` which is introduced for OceanBase Binlog Service, just like the `azure` and `gcp` options. This PR would also change the behavior of the atomic cut over on MySQL 8.0.13 and later versions, please let me know if anyone have concerns about it.

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -103,6 +103,7 @@ type MigrationContext struct {
 	GoogleCloudPlatform      bool
 	AzureMySQL               bool
 	AttemptInstantDDL        bool
+	OceanBase                bool
 
 	// SkipPortValidation allows skipping the port validation in `ValidateConnection`
 	// This is useful when connecting to a MySQL instance where the external port
--- a/go/base/utils.go
+++ b/go/base/utils.go
@@ -6,6 +6,7 @@
 package base
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"regexp"
@@ -62,6 +63,10 @@ func StringContainsAll(s string, substrings ...string) bool {
 }
 
 func ValidateConnection(db *gosql.DB, connectionConfig *mysql.ConnectionConfig, migrationContext *MigrationContext, name string) (string, error) {
+	if err := validateOceanBaseConnection(db, migrationContext); err != nil {
+		return "", err
+	}
+
 	versionQuery := `select @@global.version`
 
 	var version string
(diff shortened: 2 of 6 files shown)


meiji163:
Is this comment by Schlomi still relevant @whhe?
> This complicates matters. I can't just place a lock on both tables, since I want to be able to keep writing to the ghost table what remains in the binary logs after having locked the original table. The only way to be able to do so is if the same connection which places the lock would be the one to apply the remaining events; this is doable but non trivial.

It looks like `atomicCutOverMySQL8` still waits for the applier to finish the DML events up to the lock, while the table is locked with `singletonDB`

whhe:
@meiji163 I added `LockGhostTable` after `waitForEventsUpToLock`, so it should lock the ghost table with `singletonDB` after all events are applied to the ghost table.

I removed the command-line flag `oceanbase`, it could be replaced by the new `validateOceanBaseConnection` function.

Issue #1262, opened 2023-03-15
Regression in master: SET transaction_isolation = 'repeatable_read';


morgo:
> This is the place to report a bug, ask a question, or suggest an enhancement.

In this PR https://github.com/github/gh-ost/commit/da3514253f5cd848c1cbcaad63ad794427ed1833

`GetDBUri()` will execute a statement such as the following:

```
set transaction_isolation = 'repeatable-read';
```

This variable was only [introduced in MySQL 5.7.20](https://dev.mysql.com/doc/refman/5.7/en/server-system-variables.html#sysvar_transaction_isolation), so if you are using an earlier version such as what Aurora 5.7 is, it will result in an error:

```
mysql> set transaction_isolation = 'repeatable-read';
ERROR 1193 (HY000): Unknown system variable 'transaction_isolation'
```

Instead the statement needs to be `set tx_isolation = 'repeatable-read'`, but this statement is not supported by MySQL 8.0. So some version-specific logic is probably required.

Uplink03:
I encountered this because I use MariaDB 10.5 (not my choice, and I know it's not officially supported by gh-ost).

Thought I'd mention that the commit that introduced this change is actually [3f44e043](https://github.com/github/gh-ost/commit/3f44e04365725e2d52ef05bdb997077a4d07abe1), not the one mentioned by OP (although it's the same line OP saw).

For my needs, I edited [`go/mysql/connection.go`](https://github.com/github/gh-ost/blob/v1.1.6/go/mysql/connection.go#L130), replaced `transaction_isolation` with `tx_isolation`, and rebuilt gh-ost from source.

sj26:
For those trapped on MySQL 5.7, downgrading to gh-ost 1.1.5 works:

https://github.com/github/gh-ost/releases/tag/v1.1.5

suwalski:
What is the solution to this? Version 1.1.6 is unusable on MariaDB, and 1.1.5 can't convert tables to UTF-8 space, per #1158...

suwalski:
> What is the solution to this? Version 1.1.6 is unusable on MariaDB, and 1.1.5 can't convert tables to UTF-8 space, per #1158...

My solution was to `hexedit` the gh-ost binary like so:

```
0038E800   73 65 74 3D  75 74 66 2D  38 74 72 61  6E 73 61 63  set=utf-8transac
[...]

PR #1441, opened 2024-08-14
Add `--transaction-isolation` flag

timvaillancourt:

Related issue: https://github.com/github/gh-ost/issues/1262

### Description

This PR resolves https://github.com/github/gh-ost/issues/1262 by adding a `--transaction-isolation` flag that supports both `REPEATABLE-READ` _(default - what GitHub tests)_ and `READ-COMMITTED`

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -6,6 +6,7 @@
 package base
 
 import (
+	"errors"
 	"fmt"
 	"math"
 	"os"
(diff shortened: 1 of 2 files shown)


arthurschreiber:
Do we really need this flag? I'm absolutely sure that mysql replication uses `READ_COMMITTED` when applying RBR changes from the binlog. As gh-ost does not support statement based replication, there's no point in using `REPEATABLE_READ` for the changelog applier.

For the table copy part, I don't see a reason how `READ_COMMITTED` would have any negative side-effects either. Right now, `REPEATABLE_READ` might copy an "old" version of the row data, but the changelog applier will fix that up afterwards.

With `READ_COMMITTED`, we'll always read the "latest" version of the data, so in theory there'll be less changes that need to be applied by the changelog applier, but I don't see any negative sides to this. 🤔

timvaillancourt:
@arthurschreiber I pondered using `READ_COMMITTED` 100% in an earlier thread when setting transaction isolation was introduced, but there was some concerns around using a new isolation level. Allowing users that previously were using `READ_COMMITTED` before the enforcement was introduced seemed like the easiest way forward

But for the most part I agree, `REPEATABLE_READ` isolation is usually not required and can actually introduce stale results as you mention. There is one spot where a snapshot read might have a benefit, however: the calculation of ]the min/max chunk ranges](https://github.com/github/gh-ost/blob/master/go/logic/applier.go#L542-L561). A snapshot isolation guarantees both the min and max query are operating on the same data - which sounds like a good thing but I'm not 100% sure

PR #1442, opened 2024-08-18
Count rows queries

BarShauli555:

### Description


go/logic/inspect.go
```diff
@@ -554,7 +554,7 @@ func (this *Inspector) CountTableRows(ctx context.Context) error {
 		return err
 	}
 
-	query := fmt.Sprintf(`select /* gh-ost */ count(*) as count_rows from %s.%s`, sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.OriginalTableName))
+	query := fmt.Sprintf(`select /* gh-ost */ table_rows as count_rows from information_schema.tables where table_schema="%s" and table_name="%s"`, sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.OriginalTableName))
```
meiji163, line 557: table_rows from information_schema is an estimate, not necessarily the exact row count. Elsewhere we estimate the row count if `select count(*)` cannot complete. So we do not wish to change this.

Issue #1452, opened 2024-09-20
Fixing "FATAL No shared unique key can be found after ALTER! Bailing out"


htmsousa:
While using gh-ost, I came across the error "FATAL No shared unique key can be found after ALTER! Bailing out" already reported in issue https://github.com/github/gh-ost/issues/647.

Consider the following table:

```
CREATE TABLE `txn` (
   `txnId` int(10) unsigned NOT NULL AUTO_INCREMENT,
   `submittedDate` date NOT NULL DEFAULT '0000-00-00',
   `merchantId` int(10) unsigned NOT NULL DEFAULT '0',
   `amt` decimal(19,6) DEFAULT NULL,
   PRIMARY KEY (`txnId`,`submittedDate`)
 ) 
```

If one tries to do the alter table to replace ```PRIMARY KEY (`txnId`)``` with ```PRIMARY KEY (`txnId`,`submittedDate`)``` with ```--alter="DROP PRIMARY KEY, ADD PRIMARY KEY (txnId, submittedDate)"```, gh-ost will abort the operation with the error "**FATAL No shared unique key can be found after ALTER! Bailing out**"

On the page https://github.com/github/gh-ost/blob/master/doc/shared-key.md it is defined what is allowed and what is not and if this error happens, it is mentioned that a 2-step migration is required.

The example above is one of the "not allowed" cases. Looking at the code, the logic mandated that both original and ghost tables must share at least one "unique key". This makes sense to a certain point, otherwise, inserts in ghost table could fail with duplicate unique keys.

However, if we look at the alter table above, it is not possible to have duplicate unique key errors because the original unique key is still part of the new key. If the data in original table was unique according to the original unique key, it will still be unique if the new unique key in ghost table has the columns of the original unique key, even if the new column in the new unique key can have duplicate values.

In my view, the current code should be changed. Instead of trying to find pairs of same key in both tables, it should check if each unique key in original table exists as a subset of a unique key in the list of unique keys in ghost table.

I created a pull request with the code change to fix this problem. 

**Note 1:**
The proposed 2-step migration for the case above could be ok, but our table has half billion records, consuming 1.5 TB of space. When we run gh-ost on this table, it runs for almost 30 hours.. doing alter table in 2 steps would mean 60 hours migration.

**Note 2:** 
[...]

PR #1453, opened 2024-09-20
Improve logic to include originalUniqueKey in sharedUniqueKeys

htmsousa:
Until now, when building the list of SharedUniqueKeys, the code was selecting only the originalUniqueKeys that existed in ghostUniqueKeys list.

However, as part of the alter table, it is possible to add more unique keys or even composed primary keys with multiple columns.

The new logic keeps the old behaviour (it matches originalUniqueKey with exactly the same ghostUniqueKey) but also supports the cases where the originalUniqueKey is now a subset of one of the new ghostUniqueKeys.

If such a case happens, we can still use the originalUniqueKey as normal because a new ghostUniqueKey with the columns of originalUniqueKey plus some new columns is inherently unique just by the columns in the originalUniqueKey, no matter the value in the new columns of the new unique key.

Related issue: https://github.com/github/gh-ost/issues/1452

### Description


meiji163:
👋 Thanks for the PR! Can we add a test in `localtests/` to verify this case?

htmsousa:
> 👋 Thanks for the PR! Can we add a test in `localtests/` to verify this case?

Thanks for the feedback. I was reading the documentation, but it is not clear how to run the localtests locally. 
image

htmsousa:
@meiji163 I manage to make the localtests to run locally. Existing test was update and new one added.

htmsousa:
@meiji163 apologies but I made a mistake when I merged the master branch to my branch. it should be fixed now.

meiji163 (approved):

PR #1456, opened 2024-10-21
Switch all unit tests to use `stretchr/testify` for assertions.

arthurschreiber:
### Description

This PR switches all unit tests to use `stretchr/testify` for assertions.


go/sql/parser_test.go
```diff
@@ -52,23 +51,23 @@ func TestParseAlterStatementWithAutoIncrement(t *testing.T) {
...
-	test.S(t).ExpectFalse(parser.HasNonTrivialRenames())
-	test.S(t).ExpectFalse(parser.IsAutoIncrementDefined())
-	test.S(t).ExpectEquals(len(parser.columnRenameMap), 2)
-	test.S(t).ExpectEquals(parser.columnRenameMap["ts"], "ts")
-	test.S(t).ExpectEquals(parser.columnRenameMap["f"], "f")
+	require.NoError(t, err)
+	require.Equal(t, statement, parser.alterStatementOptions)
+	require.False(t, parser.HasNonTrivialRenames())
+	require.False(t, parser.IsAutoIncrementDefined())
+	require.Equal(t, 2, len(parser.columnRenameMap))
```
danieljoos, line 68: **nit**: there also is `require.Len()`

danieljoos (dismissed):

go/sql/parser_test.go
arthurschreiber: Nice! Will apply that change to all the cases I can find.

arthurschreiber: Done!

danieljoos (approved):

PR #1457, opened 2024-10-21
Use testcontainers to spawn MySQL server container in unit tests.

arthurschreiber:
### Description

Unit tests in `gh-ost` are pretty basic right now. They often only verify simple behaviour like query generation, but fall short of testing more complex behaviours like actually running the changelog applier, because those require an actual MySQL instance to be running, and then might also depend on the exact settings of that instance, which we might not even have control over.

By using `testcontainers-go`, we can spawn MySQL containers using _exactly_ the settings we want to test.

This adds a few simple test cases to the `Applier` tests that show how this could work.


meiji163 (approved):

go/logic/applier_test.go
```diff
@@ -185,3 +192,181 @@ func TestApplierInstantDDL(t *testing.T) {
...
+
+type ApplierTestSuite struct {
+	suite.Suite
+
+	mysqlContainer testcontainers.Container
+}
+
+func (suite *ApplierTestSuite) SetupSuite() {
+	ctx := context.Background()
+	req := testcontainers.ContainerRequest{
```
mdelapenya, line 204: Hello from Testcontainers Go community! 👋  Thanks for using us 

I'm here just to suggest using the dedicated MySQL module, if possible: https://golang.testcontainers.org/modules/mysql/

And please let us know if you need anything from our side 🙏

meiji163: 👋  thanks for the suggestion! We will look into using this.

PR #1461, opened 2024-10-23
Fix generated column in unique key error

meiji163:
### Description

This PR fixes a bug when the unique key contains a virtual column. The update query builder checks `uniqueKeyColumns` is a subset of `sharedColumns` which only contains non-virtual columns, so the migration fails.

We can reproduce the bug by adding `update` DML to the [generated-columns-unique](https://github.com/github/gh-ost/blob/master/localtests/generated-columns-unique/create.sql) test case.


go/sql/types.go
```diff
@@ -244,6 +245,19 @@ func (this *ColumnList) IsSubsetOf(other *ColumnList) bool {
 	return true
 }
 
+func (this *ColumnList) FilterBy(f func(Column) bool) *ColumnList {
```
hugodorea, line 248: Since we're creating a new instance of `ColumnList`, what if we turn this method into factory function? Something like:

```go 
 NewFilteredColumnList(ordinals ColumnsMap, f func(Column) bool) *ColumnList {
 ...
}
```

go/sql/types.go
```diff
@@ -244,6 +245,19 @@ func (this *ColumnList) IsSubsetOf(other *ColumnList) bool {
...
+	filteredList := &ColumnList{}
+	filteredList.Ordinals = this.Ordinals
+	filteredCols := []Column{}
```
susanzhang27, line 251: nit: filteredCols := make([]Column, len(this.columns))

susanzhang27: max size should always be the len of the columns?

go/sql/types.go
```diff
@@ -244,6 +245,19 @@ func (this *ColumnList) IsSubsetOf(other *ColumnList) bool {
...
+	for _, column := range this.columns {
+		if f(column) {
+			filteredCols = append(filteredCols, column)
+		}
+	}
+	filteredList.columns = filteredCols
+	return filteredList
```
susanzhang27, line 258: could always just return this at the end too

```
return &ColumnList{
        Ordinals: this.Ordinals,
        columns:  filteredCols,
    }
```

susanzhang27 (approved):

susanzhang27 (approved):

PR #1462, opened 2024-10-23
Use multiStatement to apply DML

meiji163:
### Description

This PR changes the Applier to batch together DML statements into [multi-statements](https://dev.mysql.com/doc/c-api/9.0/en/c-api-multiple-queries.html) in order to reduce the number of network trips in `ApplyDMLEventQueries`.

During testing on our `--test-on-replica` setup this provided around 10% increase in rows/second throughput

Rows Modified (rate); `dml-batch-size=10`
rows modified (rate); dml-batch-size=10

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -1234,46 +1234,68 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent)) 
 		if _, err := tx.Exec(sessionQuery); err != nil {
 			return rollback(err)
 		}
-		rowDeltas := make([]int64, 0, len(dmlEvents))
-		multiArgs := []driver.NamedValue{}
-		var multiQueryBuilder strings.Builder
+
+		buildResults := make([]*dmlBuildResult, 0, len(dmlEvents))
 		for _, dmlEvent := range dmlEvents {
 			for _, buildResult := range this.buildDMLEventQuery(dmlEvent) {
 				if buildResult.err != nil {
 					return rollback(buildResult.err)
 				}
-				for _, arg := range buildResult.args {
-					multiArgs = append(multiArgs, driver.NamedValue{Value: driver.Value(arg)})
-				}
-				rowDeltas = append(rowDeltas, buildResult.rowsDelta)
-				multiQueryBuilder.WriteString(buildResult.query)
-				multiQueryBuilder.WriteString(";\n")
+
+				buildResults = append(buildResults, buildResult)
 			}
 		}
 
-		//this.migrationContext.Log.Infof("Executing query: %s, args: %+v", multiQueryBuilder.String(), multiArgs)
 		execErr := conn.Raw(func(driverConn any) error {
 			ex, ok := driverConn.(driver.ExecerContext)
 			if !ok {
 				return fmt.Errorf("could not cast driverConn to ExecerContext")
 			}
+
+			nvc, ok := driverConn.(driver.NamedValueChecker)
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to NamedValueChecker")
+			}
+
+			var multiArgs []driver.NamedValue
+			multiQueryBuilder := strings.Builder{}
+			var rowDeltas []int64
+
+			for _, buildResult := range buildResults {
+				for _, arg := range buildResult.args {
+					nv := driver.NamedValue{Value: driver.Value(arg)}
+					nvc.CheckNamedValue(&nv)
+					multiArgs = append(multiArgs, nv)
+				}
+
+				multiQueryBuilder.WriteString(buildResult.query)
+				multiQueryBuilder.WriteString(";\n")
+
+				rowDeltas = append(rowDeltas, buildResult.rowsDelta)
+			}
+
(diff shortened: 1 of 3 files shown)


arthurschreiber (approved):

go/mysql/connection.go
```diff
@@ -132,6 +132,7 @@ func (this *ConnectionConfig) GetDBUri(databaseName string) string {
 	connectionParams := []string{
 		"autocommit=true",
 		"interpolateParams=true",
+		"multiStatements=true",
```
arthurschreiber, line 135: I think it would be better not to set this on all connections opened in `gh-ost`, but only for the `db` connection of the  `Applier` struct.

arthurschreiber: This can be done in `InitDBConnections`.

go/logic/applier.go
```diff
@@ -1225,26 +1234,70 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+			}
+
+			nvc, ok := driverConn.(driver.NamedValueChecker)
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to NamedValueChecker")
+			}
+
+			var multiArgs []driver.NamedValue
+			multiQueryBuilder := strings.Builder{}
+			var rowDeltas []int64
```
arthurschreiber, line 1262: This list could be pre-allocated to the length of `buildResults`

go/logic/applier.go
```diff
@@ -1225,26 +1234,70 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to ExecerContext")
+			}
+
+			nvc, ok := driverConn.(driver.NamedValueChecker)
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to NamedValueChecker")
+			}
+
+			var multiArgs []driver.NamedValue
```
arthurschreiber, line 1260: When building `buildResults`, we could count the number of `args` we have and pre-allicate this to the correct size.

go/logic/applier.go
```diff
@@ -1225,26 +1234,70 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+		execErr := conn.Raw(func(driverConn any) error {
+			ex, ok := driverConn.(driver.ExecerContext)
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to ExecerContext")
+			}
+
+			nvc, ok := driverConn.(driver.NamedValueChecker)
+			if !ok {
+				return fmt.Errorf("could not cast driverConn to NamedValueChecker")
+			}
```
arthurschreiber, line 1258: I'm wondering if we should just use unsafe casts and panic here. Otherwise, these errors might end up causing retries, but I don't see any sane way how the casts could fail.

go/logic/applier.go
```diff
@@ -1207,13 +1210,19 @@ func (this *Applier) buildDMLEventQuery(dmlEvent *binlog.BinlogDMLEvent) []*dmlB
...
+		conn, err := this.db.Conn(ctx)
 		if err != nil {
 			return err
 		}
+		defer conn.Close()
 
+		tx, err := conn.BeginTx(ctx, nil)
+		if err != nil {
+			return err
+		}
```
arthurschreiber, line 1225: We could have the `BEGIN` statement be the first in the list of the multistatement and avoid the network roundtrip for this.

`SET SESSION` can happen before the `BEGIN` now. The reason why it was done inside of the transaction was to ensure that it happens on the same connection, but we're now explicitly checking out the connection out of the database pool.

go/logic/applier.go
```diff
@@ -1225,26 +1234,70 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+			if err != nil {
+				err = fmt.Errorf("%w; query=%s; args=%+v", err, multiQueryBuilder.String(), multiArgs)
+				this.migrationContext.Log.Errorf("Error exec: %+v", err)
+				return err
 			}
+
+			mysqlRes, ok := res.(drivermysql.Result)
+			if !ok {
+				return fmt.Errorf("Could not cast %+v to mysql.Result", res)
+			}
```
arthurschreiber, line 1288: Again, we probably should use an unsafe cast and panic here to avoid the retry logic (and because this should never ever happen).

go/logic/applier.go
```diff
@@ -1225,26 +1234,70 @@ func (this *Applier) ApplyDMLEventQueries(dmlEvents [](*binlog.BinlogDMLEvent))
...
+
+			// each DML is either a single insert (delta +1), update (delta +0) or delete (delta -1).
+			// multiplying by the rows actually affected (either 0 or 1) will give an accurate row delta for this DML event
+			for i, rowsAffected := range mysqlRes.AllRowsAffected() {
+				totalDelta += rowDeltas[i] * rowsAffected
```
arthurschreiber, line 1293: if we add `BEGIN` to the multi statement, this calculation here will need to be updated accordingly.

arthurschreiber (approved):

Issue #1386, opened 2024-03-07
"attempt-instant-ddl" should support lock_wait_timeout


Hexcles:
Currently gh-ost only sets `lock_wait_timeout` when doing normal cutover: https://github.com/search?q=repo%3Agithub%2Fgh-ost%20lock_wait_timeout&type=code

When Instant DDL is used, there doesn't seem to be a way to set the lock timeout. We can either reuse the same `cut-over-lock-timeout-seconds` flag or introduce a new one specifically for Instant DDL.

hakusaro:
Peering into https://github.com/github/gh-ost/pull/1201, I found it perplexing that it says this in the docs:

> This is not a problem for most scenarios, but it could be a problem for users that start the DDL during a period with long running transactions.

This is really critical information, in my opinion, because if gh-ost is supposed to be inherently safe, it seems to jeopardize this safety by potentially creating table outages with no controllable timeout here. We primarily introduced gh-ost because of long-running transactions that were hard to pin down and a lack of safety with LHM in these scenarios. While we've mostly cleaned these up, I still think anything that could remotely incur a table outage should have defined characteristics for how long the table will be out.

meiji163:
Lock-timeout for instant DDL added in https://github.com/github/gh-ost/pull/1468

PR #1468, opened 2024-11-07
fix: use cut-over-lock-timeout for instant DDL

artemvovk:

Related issue: https://github.com/github/gh-ost/issues/1386

### Description

reusing the cut-over-lock-timeout from the cutover code. The lock wait timeout in the original code is actually set to double the setting, so we keep that consistent. Altho I do wonder why it's 2x.

there are no existing tests for the CutOverLockTimeoutSeconds currently, I can add some if that would help.


arthurschreiber:
I'm wondering if we shouldn't just set the lock timeout to the minimal possible value for instant ddl? 🤔

artemvovk:
that would be my first choice; but also having it configurable doesn't seem like the worst idea - I can update either way, delegating the decision to the reviwers.

meiji163 (approved):

meiji163:
Configurable lock timeout seems OK to me, I will merge this.

Issue #1259, opened 2023-03-02
Fail to verify SSL identity when recursing into a host with a different SSL cert


Hexcles:
Here when gh-ost is recursively finding masters:
https://github.com/github/gh-ost/blob/b7db8c6ca76712ae28bd97ca0f1c29fd152228cd/go/mysql/utils.go#L131

It copies the entire `connectionConfig`, including the private `tlsConfig` field:
https://github.com/github/gh-ost/blob/b7db8c6ca76712ae28bd97ca0f1c29fd152228cd/go/mysql/connection.go#L49

without updating its hostname:
https://github.com/github/gh-ost/blob/b7db8c6ca76712ae28bd97ca0f1c29fd152228cd/go/mysql/connection.go#L98

Hence, it tries to connect to a new host likely with a different hostname and certificate, but still asks x509 to verify against the old hostname, which always fails:
> x509: certificate is valid for [old hostname], not [newly found hostname]

petervandoros:
We have hit this issue as well.

I spent some time and created a PR that resolves the issues described in this issue and confirmed it working:

https://github.com/github/gh-ost/pull/1487

petervandoros:
The above PR has been merged. Just waiting on a release.

PR #1487, opened 2025-01-02
Fix TLS verify error when gh-ost discovers the replication master

petervandoros:

Related issue: https://github.com/github/gh-ost/issues/1259

### Description

This PR resolves TLS certificate verification failures when gh-ost discovers, or is forced to used, different master server hosts to what's specified by the `--host` arg.

Example error:

> x509: certificate is valid for [old hostname], not [newly found hostname]

There are three reasons that, when combined, cause this error:
1. The `DuplicateCredentials` [function](https://github.com/github/gh-ost/blob/master/go/mysql/connection.go#L45-L57) doesn't update the TLS config's `ServerName` property to match the connection's updated Key/Hostname.
2. Only the inspector's database TLS configuration is [registered with the MySQL driver during initialisation](https://github.com/github/gh-ost/blob/master/go/cmd/gh-ost/main.go#L307). The TLS configuration for other connections are not registered.
3. The config key used to register the TLS config with the mysql driver is [hard coded to `ghost`](https://github.com/github/gh-ost/blob/master/go/mysql/connection.go#L125).

####  Considerations

- I'm relatively new to Go, so please help me correct any Go idioms I missed or got wrong 😄 .
- Even though I updated and added tests I thought appropriate, I didn't add tests for verifying the TLS config because I'm unsure how or if that's possible. I'm open to some direction on this.
- Even though I have tested this PR in our test environment in AWS using Aurora 3 with MySQL 8.0 compatibility, I don't expect the changes to cause any issues in other environments or MySQL versions.


meiji163:
This LGTM, thanks for the fix!

meiji163 (approved):

petervandoros:
Thank you!

When could we expect the changes be released in a new version? (no urgency, just asking for internal planning)

Issue #1498, opened 2025-02-13
Add option to fail the gh-ost process on SQL warnings


grodowski:
My team is investigating using Gh-ost for online schema migrations and we'd like to propose a feature we've been relying on in our current approach with LHM.

A new CLI option (`raise_on_warnings`?) would query for `SHOW WARNINGS` after each insert batch and fail the migration if warnings are found. This would optionally prevent data loss when e.g.:
- adding a unique index with duplicate rows
- changing table's collation causing new duplicates
- adding a not null constraint with null values
- resizing columns

An equivalent in LHM is `raise_on_warnings`:  https://github.com/Shopify/lhm/blob/46204927a2933cfe239c77ac84ab58b5c7deb3a0/lib/lhm/chunker.rb#L92-L100 

I have a prototype that shows the direction of how this option could be implemented in `Applier.ApplyIterationInsertQuery`: https://github.com/Shopify/gh-ost/pull/1 

Some open questions:
- Is it desired to always process warnings (with the CLI option enabled) or only if a row count discrepancy has been detected. The latter is the current, less strict behaviour of Shopify/lhm

I'd love to hear your suggestions before we propose a PR contribution, thanks!

PR #1500, opened 2025-02-28
PanicOnWarnings option to detect SQL warnings and fail the copy process

grodowski:
Related issue: https://github.com/github/gh-ost/issues/1498

### Description

This PR adds a CLI flag `--panic-on-warnings` that makes `gh-ost` fail after copying each batch when:
- the expected, selected row count differs from actual copied row count (`affectedRows`)
- SQL warnings are present, indicating why some rows were dropped

I also decided to always log all SQL warnings, which could be useful to find issues with data truncation or not null constraints even when the row count matches.

Opening as draft with a few remaining items:
- ~review documentation~
- ~review test coverage~
- ~check explain analyze on the updated range select query~


grodowski:
Ran gh-ost locally on a `1.7GiB` table with `110k rows` with generated blob data to check if there's any noticable impact of the added `count(*)` for each batch.

Migrate cmd:

```
gh-ost --allow-on-master --allow-master-master --table="large_books" --alter="add index author_index(author(255))" --database="schema_migrations_test_app_development" --user="root" --password="" --initially-drop-old-table --initially-drop-ghost-table --default-retries=3 --verbose --debug --host=$MYSQL_HOST --port=$MYSQL_PORT --execute
```

Generic `time` benchmark on two binaries build on `master` and `grodowski/raise_on_warnings` (`gh-ost-new`)
         
```
gh-ost-master --allow-on-master --allow-master-master --table="large_books"    0.99s user 1.45s system 7% cpu 30.708 total
gh-ost-master --allow-on-master --allow-master-master --table="large_books"    0.98s user 1.43s system 7% cpu 31.321 total
```

```
gh-ost-new --allow-on-master --allow-master-master --table="large_books"       1.00s user 1.46s system 7% cpu 32.743 total
gh-ost-new --allow-on-master --allow-master-master --table="large_books"       0.98s user 1.41s system 7% cpu 31.117 total
```

Then, ran these explains on the copy query (`BuildUniqueKeyRangeEndPreparedQueryViaTemptable`):

```
-- before
explain analyze select /* gh-ost mydb.tbl test */ id
from (
    select
        id
    from
        large_books
    where id >= 50000 and id <= 52000
    order by
        id asc
    limit 1000) select_osc_chunk
order by
    id desc
limit 1

| -> Limit: 1 row(s)  (cost=1021..1021 rows=1) (actual time=0.943..0.943 rows=1 loops=1)
    -> Sort: select_osc_chunk.id DESC, limit input to 1 row(s) per chunk  (cost=1021..1021 rows=1) (actual time=0.942..0.942 rows=1 loops=1)
        -> Table scan on select_osc_chunk  (cost=906..921 rows=1000) (actual time=0.746..0.843 rows=1000 loops=1)
            -> Materialize  (cost=906..906 rows=1000) (actual time=0.744..0.744 rows=1000 loops=1)
                -> Limit: 1000 row(s)  (cost=806 rows=1000) (actual time=0.137..0.64 rows=1000 loops=1)
                    -> Filter: ((large_books.id >= 50000) and (large_books.id <= 52000))  (cost=806 rows=3992) (actual time=0.136..0.579 rows=1000 loops=1)
                        -> Covering index range scan on large_books using PRIMARY over (50000 <= id <= 52000)  (cost=806 rows=3992) (actual time=0.132..0.446 rows=1000 loops=1)
```

```
-- after
explain analyze select /* gh-ost mydb.tbl test */
    id,
    (select count(*) from (
        select
            id
        from
            large_books
        where id >= 50000 and id <= 52000
        order by
            id asc
        limit 1000
    ) select_osc_chunk)
from (
    select
        id
    from
        large_books
    where id >= 50000 and id <= 52000
    order by
        id asc
    limit 1000
) select_osc_chunk
order by
    id desc
limit 1;

| -> Limit: 1 row(s)  (cost=1021..1021 rows=1) (actual time=1.21..1.21 rows=1 loops=1)
    -> Sort: select_osc_chunk.id DESC, limit input to 1 row(s) per chunk  (cost=1021..1021 rows=1) (actual time=1.21..1.21 rows=1 loops=1)
        -> Table scan on select_osc_chunk  (cost=906..921 rows=1000) (actual time=0.955..1.08 rows=1000 loops=1)
            -> Materialize  (cost=906..906 rows=1000) (actual time=0.953..0.953 rows=1000 loops=1)
                -> Limit: 1000 row(s)  (cost=806 rows=1000) (actual time=0.161..0.813 rows=1000 loops=1)
                    -> Filter: ((large_books.id >= 50000) and (large_books.id <= 52000))  (cost=806 rows=3992) (actual time=0.16..0.732 rows=1000 loops=1)
                        -> Covering index range scan on large_books using PRIMARY over (50000 <= id <= 52000)  (cost=806 rows=3992) (actual time=0.157..0.574 rows=1000 loops=1)
-> Select #2 (subquery in projection; run only once)
    -> Aggregate: count(0)  (cost=1021..1021 rows=1) (actual time=1.04..1.04 rows=1 loops=1)
        -> Table scan on select_osc_chunk  (cost=906..921 rows=1000) (actual time=0.894..0.992 rows=1000 loops=1)
            -> Materialize  (cost=906..906 rows=1000) (actual time=0.894..0.894 rows=1000 loops=1)
                -> Limit: 1000 row(s)  (cost=806 rows=1000) (actual time=0.14..0.774 rows=1000 loops=1)
                    -> Filter: ((large_books.id >= 50000) and (large_books.id <= 52000))  (cost=806 rows=3992) (actual time=0.14..0.694 rows=1000 loops=1)
                        -> Covering index range scan on large_books using PRIMARY over (50000 <= id <= 52000)  (cost=806 rows=3992) (actual time=0.139..0.535 rows=1000 loops=1)
```

Looks like there will be no significant change in perf?

meiji163:
Looks good! I will test the performance this internally (hopefully this week)

go/logic/applier.go
```diff
@@ -753,6 +757,35 @@ func (this *Applier) ApplyIterationInsertQuery() (chunkSize int64, rowsAffected
...
+			var sqlWarnings []string
+			for rows.Next() {
+				var level, message string
+				var code int
+				if err := rows.Scan(&level, &code, &message); err != nil {
+					this.migrationContext.Log.Warningf("Failed to read SHOW WARNINGS row")
+					continue
+				}
+				pkDuplicateSuffix := fmt.Sprintf("for key '%s.PRIMARY'", this.migrationContext.GetGhostTableName())
+				if strings.HasPrefix(message, "Duplicate entry") && strings.HasSuffix(message, pkDuplicateSuffix) {
```
meiji163, line 781: We may have a different shared unique key than the primary key (`migrationContext.UniqueKey`)

grodowski: Good catch! Based on how the shared key is selected in `inspectOriginalAndGhostTables`, it seems the correct behavior here would be to ignore any warnings including `migrationContext.UniqueKey.Name` instead of just hardcoding the pk. This is a logic spill from LHM - I'll fix that and see if I can add a unit test.

grodowski, line 781: I think that'll do it. Started working on a new test for this in `applier_test.go`, but running into problems with testing and `testcontainers` stuck waiting for the `mysql` port, even though it's open. Need to debug this first to be able to iterate on the test.

```suggestion
				migrationUniqueKeySuffix := fmt.Sprintf("for key '%s.%s'", this.migrationContext.GetGhostTableName(), this.migrationContext.UniqueKey.Name)
				if strings.HasPrefix(message, "Duplicate entry") && strings.HasSuffix(message, migrationUniqueKeySuffix) {
					continue
				}
```

grodowski: Addressed in [36eaf27](https://github.com/github/gh-ost/pull/1500/commits/36eaf27e4c7f0a425ebd7192d18ff5e885744587) 👍

go/sql/builder.go
```diff
@@ -275,7 +275,7 @@ func BuildUniqueKeyRangeEndPreparedQueryViaOffset(databaseName, tableName string
 
 	uniqueKeyColumnNames := duplicateNames(uniqueKeyColumns.Names())
 	uniqueKeyColumnAscending := make([]string, len(uniqueKeyColumnNames))
-	uniqueKeyColumnDescending := make([]string, len(uniqueKeyColumnNames))
+	uniqueKeyColumnDescending := make([]string, len(uniqueKeyColumnNames)) // TODO unused variable
```
meiji163, line 278: outdated comment?

go/cmd/gh-ost/main.go
```diff
@@ -108,6 +108,7 @@ func main() {
 	chunkSize := flag.Int64("chunk-size", 1000, "amount of rows to handle in each iteration (allowed range: 10-100,000)")
 	dmlBatchSize := flag.Int64("dml-batch-size", 10, "batch size for DML events to apply in a single transaction (range 1-100)")
 	defaultRetries := flag.Int64("default-retries", 60, "Default number of retries for various operations before panicking")
+	flag.BoolVar(&migrationContext.PanicOnWarnings, "panic-on-warnings", false, "Panic when SQL warnings are encountered when copying a batch indicating data loss")
```
meiji163, line 111: Let's add this to `command-line-flags.md` as well

meiji163 (commented):
Almost ready to merge, just a couple comments 🚀

There is another edge case where the warning check doesn't work properly. If the shared unique key is being renamed, the `INSERT IGNORE` warning will say `Duplicate entry ... for key _${table}_gho.${renamed_unique_shared_key}`, because `migrationContext.UniqueKey.Name` is the name of the unique key on the original table.
For example, `localtests/swap-uk-uk` test case fails with `--panic-on-warnings`.
I believe we can fix it by changing `migrationContext.UniqueKey.Name` to be the name of the shared key on the ghost table.

Second, the main use of panic on warnings is to prevent data loss on `add unique key` migrations, but we should note that this PR doesn't detect dataloss from the binlog applier, which uses [`replace into`](https://github.com/github/gh-ost/blob/5c3e42ac91f63d34bab810d77cef68036aa5b543/go/sql/builder.go#L492-L498). We can address this in a followup PR.

go/cmd/gh-ost/main.go
meiji163: I think it's sensible to default to "true" in the future

go/sql/builder.go
grodowski: Ah, nope! This method only uses ascending order, just removed it in [74f60a9](https://github.com/github/gh-ost/pull/1500/commits/74f60a97c6d3afb588a80cc0a58c2e2fed37f6b4)

dev.yml
```diff
@@ -0,0 +1,21 @@
+name: gh-ost
```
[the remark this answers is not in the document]
grodowski: Removed for now 👍.  `dev.yml` powers `dev`, an internal project bootstrap tool used at Shopify, which I needed to set up `gh-ost` locally with Docker and Podman. If we contribute to `gh-ost` often enough, I might ask you someday to keep this file here 🙃

grodowski:
> If the shared unique key is being renamed, the INSERT IGNORE warning will say Duplicate entry ... for key _${table}_gho.${renamed_unique_shared_key}, because migrationContext.UniqueKey.Name is the name of the unique key on the original table.

Working on a fix for this ⏳ It gets somewhat more involved as we seem to rely on the original table key name e.g. [here](https://github.com/Shopify/gh-ost/blob/1487b5c84887e3753257eed3a897a0167513fb3e/go/sql/builder.go#L440)

> Second, the main use of panic on warnings is to prevent data loss on add unique key migrations, but we should note that this PR doesn't detect dataloss from the binlog applier, which uses [replace into](https://github.com/github/gh-ost/blob/5c3e42ac91f63d34bab810d77cef68036aa5b543/go/sql/builder.go#L492-L498). We can address this in a followup PR.

That's a good point. I believe https://github.com/Shopify/gh-ost/blob/1487b5c84887e3753257eed3a897a0167513fb3e/go/logic/applier.go#L1372-L1379 would need to re-use the `show warnings` code. Need to check if it would works with the multi-statement query.

meiji163:
> proposed fix in [1ca8ffc](https://github.com/github/gh-ost/pull/1500/commits/1ca8ffcb5fd26b3a0c4b63bbd2233e8cc15c16be)

Nice! It looks like the exact `Duplicate key` warning messages differs between versions (the peril of parsing error messages 😅 ). Perhaps `strings.HasPrefix(message, "Duplicate entry") && strings.Contains(message, this.migrationContext.UniqueKey.NameInGhostTable)` would work?

grodowski:
> Perhaps strings.HasPrefix(message, "Duplicate entry") && strings.Contains(message, this.migrationContext.UniqueKey.NameInGhostTable) would work?

Nice, glad that we caught it 😌 Fixed this in the latest commit 👍

meiji163:
> Edit: Tried to write a new unit test for the Applier and didn't see any SQL warnings written - it just overwrites on events by design. So is this about new rows being written by clients while a migration is running (is it a responsibility of the client to not do that?), or is there a race condition I don't see? An example scenario would be really helpful.

Yes `replace into` doesn't raise warning by design, so we'd have to change it to `insert` to detect data loss. e.g. 
- migration adding a unique key
- user inserts row that's has duplicate key, where the row was already copied to ghost table
- binlog applier sees the insert and replaces the row, without raising any warnings

We can consider it out of scope for this PR

meiji163:
Thanks for your work on this! 🚀

meiji163 (approved):

xiaoxuanfeng:
> > Edit: Tried to write a new unit test for the Applier and didn't see any SQL warnings written - it just overwrites on events by design. So is this about new rows being written by clients while a migration is running (is it a responsibility of the client to not do that?), or is there a race condition I don't see? An example scenario would be really helpful.
> 
> Yes `replace into` doesn't raise warning by design, so we'd have to change it to `insert` to detect data loss. e.g.
> 
> * migration adding a unique key
> * user inserts row that's has duplicate key, where the row was already copied to ghost table
> * binlog applier sees the insert and replaces the row, without raising any warnings
> 
> We can consider it out of scope for this PR

Hello
Ghost applies binlog incremental data:
The update event in binlog corresponds to the ghost updatte statement,
The delete event corresponds to the ghost delete statement,
The insert event corresponds to the ghost replace statement,
The reason for using 'replace' is because the existing data migration has already migrated this row of data to the shadow table. In fact, we can use 'insert IGNORE into' to solve all the problems easily, because we now have a solution to the conflict issue of 'insert IGNORE into'
@meiji163 @grodowski

xiaoxuanfeng:
> > > Edit: Tried to write a new unit test for the Applier and didn't see any SQL warnings written - it just overwrites on events by design. So is this about new rows being written by clients while a migration is running (is it a responsibility of the client to not do that?), or is there a race condition I don't see? An example scenario would be really helpful.
> > 
> > 
> > Yes `replace into` doesn't raise warning by design, so we'd have to change it to `insert` to detect data loss. e.g.
> > 
> > * migration adding a unique key
> > * user inserts row that's has duplicate key, where the row was already copied to ghost table
> > * binlog applier sees the insert and replaces the row, without raising any warnings
> > 
> > We can consider it out of scope for this PR
> 
> Hello Ghost applies binlog incremental data: The update event in binlog corresponds to the ghost updatte statement, The delete event corresponds to the ghost delete statement, The insert event corresponds to the ghost replace statement, The reason for using 'replace' is because the existing data migration has already migrated this row of data to the shadow table. In fact, we can use 'insert IGNORE into' to solve all the problems easily, because we now have a solution to the conflict issue of 'insert IGNORE into'

@meiji163

grodowski:
> We can consider it out of scope for this PR

https://github.com/github/gh-ost/pull/1633 by @ggilder addresses that. cc @meiji163

Issue #1484, opened 2024-12-20
Table Space Bloat Due to Gh-ost


xiehaopeng:
**Description:**
We have observed significant table space bloat as a result of using gh-ost to alter table. The number of rows before and after alter table has basically not changed, but the table space has expanded by 60%.The amount of table space expansion is related to the size of a single row in the table. When the size of a single row is about 4500 bytes, the expansion is more obvious.

We believe the primary cause of this issue appears to be the interchangeable use of "copy rows" and "apply events" during the gh-ost processes. When there are INSERT operations happening concurrently, this leads to a mix of sequential and out-of-order inserts. This mixed insertion pattern triggers InnoDB's insert point split strategy frequently. After the split, the leaf nodes contain unutilized space (page-level fragmentation), which results in wasted space and inflated table size.

**Steps to Reproduce:**
- First, mock a large table of 50GB, with the size of a single row controlled at around 4500 bytes.
- Then use native MySQL alter to organize table.
- Then write a script to simulate low-frequency insert traffic.
- Then execute gh-ost , and start the traffic simulation script before gh-ost.
- Finally, after gh-ost is completed, compare the size of the original table and the temporary table.

**A Possible Improvement Idea:**
Based on the ignoring binlog events feature of [feat binlog apply optimization #1378](https://github.com/github/gh-ost/pull/1378) we dynamically expand the maximum boundary value of the copy, ignore the large unique key insert event, which may avoid this problem.

disser4:
@xiehaopeng and @meiji163 our company was impacted by this to the tune of several TB getting larger. Is there any traction on this?

xiehaopeng:
> [@xiehaopeng](https://github.com/xiehaopeng) and [@meiji163](https://github.com/meiji163) our company was impacted by this to the tune of several TB getting larger. Is there any traction on this?

@disser4 Consider to cherry pick #1506  onto the release v1.1.7 and enabling the feature to avoid Table Space Bloat. For reference, our company has successfully used this feature for over a year without issues.

PR #1506, opened 2025-03-10
Feat ignore binlog events

xiehaopeng:

Related issue: https://github.com/github/gh-ost/issues/1484

### Description

Support for Ignoring Binlog Events Beyond Chunk Boundary Values
If the unique key value in a binlog event surpasses the maximum value defined for chunk iteration but remains below the overall maximum boundary for copying, the event is ignored. This approach is inspired by the design concept from shaohk's [feat binlog apply optimization](https://github.com/github/gh-ost/pull/1378). We have implemented the functionality to ignore such binlog events based on the code from release v1.1.7.

Dynamic Expansion of Maximum Copy Boundary Values
To prevent issues arising from the interchangeable use of "copy rows" and "apply events" during gh-ost processes, which can lead to [Table Space Bloat Due to Gh-ost #1484](https://github.com/github/gh-ost/issues/1484), we have introduced a mechanism to dynamically expand the maximum boundary value for copying. By ignoring large unique key insert events, this solution potentially mitigates the problem.


meiji163:
Thanks for the PR @xiehaopeng, it looks like a good optimization. I would like to do more testing with it

disser4:
@meiji163 can you let me know where you are with this review?

disser4:
cc:
@timvaillancourt / @rashiq

PR #1512, opened 2025-03-14
Update status format in docs

coding-chimp:
### Description

I noticed the status output in the docs is slightly outdated. Especially the throttling information having moved from `ETA` to `State` threw me a bit when I started building tooling around gh-ost. So, this PR is fixing it.


meiji163:
Thanks for the docs update!

meiji163 (approved):

Issue #1524, opened 2025-04-01
Password length should not be restricted to <=32 characters


joshvarner:
I believe this restriction originally came from an erroneous use of `report-password` in `go-mysql` which has since been resolved: https://github.com/go-mysql-org/go-mysql/issues/966

In order to fix this, we'll need to bump `go-mysql` to a newer version and then remove the length check.

meiji163:
closed by [#1525](https://github.com/github/gh-ost/pull/1525)

PR #1525, opened 2025-04-01
Remove restriction on password length and bump go-mysql to 1.11.0

joshvarner:
Related issue: https://github.com/github/gh-ost/issues/1524

### Description

In prior versions of `go-mysql`, the actual password was used for the `report-password` portion of `COM_REGISTER_SLAVE`. This has two negative side effects:

1. The password is shown plaintext in the output of `SHOW REPLICAS` on the source that was connected to (if `show-replica-auth-info=on` on the source)
2. The password length is limited to <=32 characters

We should update to a newer version of `go-mysql` that now removes this erroneous behavior and also remove our restriction on password length, since it will no longer apply.


meiji163:
If the password length limit is irrelevant now we should remove `localtests/fail-password-length`

joshvarner:
> If the password length limit is irrelevant now we should remove `localtests/fail-password-length`

Apologies for not catching this earlier-- the `localtests` mechanism wasn't super straightforward to me so I had not run them yet.

meiji163 (approved):

PR #1557, opened 2025-05-22
Remove counts from CalculateNextIterationRangeEndValues that caused problems with --panic-on-warnings

grodowski:
The split between Temptable and Offset query builders was originally introduced in https://github.com/github/gh-ost/pull/471. Then, both query builders have been changed to calculate actual chunk sizes in https://github.com/github/gh-ost/pull/1500.

The updated implementation of the Offset query introduced a bug, where a missing `order by` clause in `select_osc_chunk` can result in end values potentially surpassing the `chunk_size`. This wasn't detected during our initial testing where the query just returned rows in their original order, but may happen in real-world scenarios in case the db returns data in an undefined order.

An obvious fix would be to just add an `order by` to the Offset builder subquery, however since both builders use temptables now, it makes more sense to simplify and use only one of them. 

**Update**: using just one query builder was the initial approach in [660cbf9](https://github.com/github/gh-ost/pull/1557/commits/660cbf974570b70eb2a2c1312dd0425cd9813fff), however based on our findings that the count query can produce unpredictable results with concurrent writes (https://github.com/github/gh-ost/pull/1557#issuecomment-2902072776), I propose to drop the row count check in [adc7049](https://github.com/github/gh-ost/pull/1557/commits/adc7049addf95b938939e10275ad30fc27a8ad28). This will hopefully mitigate all count-related bugs, bring back the old query builders and make `PanicOnWarnigns` fail in cases where rows are not being dropped, but data is truncated (see added test with a truncated varchar column).

Related issue: https://github.com/github/gh-ost/issues/1498


grodowski:
Thanks for the context @meiji163, it does sound like, given our findings, we should abandon the idea of checking the row count with `--panic-on-warnings`. The original inspiration lives [here](https://github.com/Shopify/lhm/blob/2d4118ea3dcde9ac065628fe7d7f57f41dd4b6f2/lib/lhm/chunker.rb#L62-L73), but the logic might also contain bugs (e.g. `expected_rows = top - bottom + 1` causing the warning to not fire in some scenarios). 

Do you see any downsides of just dropping the row count check? I don't know there are types of warnings we might need to ignore besides the ones on `insert ignore`, though I do think we have the tooling to try it safely at my team. Then, the query builders can also go back to their more performant counterparts from https://github.com/github/gh-ost/pull/471.

grodowski:
@meiji163 added [adc7049](https://github.com/github/gh-ost/pull/1557/commits/adc7049addf95b938939e10275ad30fc27a8ad28) that removes `expectedRangeSize`, updated the PR description and also wrote a new test case to cover warnings without changes in row count.

meiji163 (approved):

Issue #1568, opened 2025-06-27
Conversion of character set from latin1 to utf8mb3 does not work


mattbooks:
When using gh-ost to migrate a table from latin1 to utf8mb3 (and likely other utf8) character encoding, the initial data copy works correctly, but new data inserted during the migration via binlog replication can cause the migration to fail if it contains characters that have different encodings between mysql latin1 and utf8mb3, for example: é à ç ñ Ä ß ø ÿ þ æ. `ERROR Error 1366 (HY000): Incorrect string value:` is thrown repeatedly.

These characters are all valid in latin1, but when gh-ost attempts to apply binary log events containing these characters to the ghost table without proper encoding conversion, errors can occur. It seems gh-ost is not converting these characters to the target character set when replacing from the binlog.

I have created a repo containing a reproduction: https://github.com/mattbooks/gh-ost-latin1-utf8-bug/tree/main

which includes a script that can reproduce the bug locally (my reproduction was with gh-ost 1.1.7 and mysql 8.0.42 on macos)

I have also included there example output.txt and error.txt from my local run which demonstrate that data is inserted correctly when copied, but if it is inserted after the migration is started, raises an error until the migration ultimately fails.

PR #1571, opened 2025-07-14
Fix conversion bug when string data comes from binlog

mattbooks:
Related issue: https://github.com/github/gh-ost/issues/1568

### Description

This PR fixes a string conversion bug by applying character set conversions when the input is `[]uint8`, in addition to the existing case of when the input is `string`.

When using gh-ost to migrate a table from latin1 to utf8mb3 character encoding, the initial data copy works correctly, but new data with special characters inserted during the migration via binlog replication fails with "Incorrect string value" errors.

The reason for this is that the data is a binary byte array when converted from the binlog, so the character set conversion is not applied.

I added a new test for latin1-encoded byte array input to this method and confirmed that the reproduction from the linked issue is fixed with this change.

### DIFF
--- a/go/sql/types.go
+++ b/go/sql/types.go
@@ -53,14 +53,22 @@ type Column struct {
 }
 
 func (this *Column) convertArg(arg interface{}, isUniqueKeyColumn bool) interface{} {
+	var arg2Bytes []byte
 	if s, ok := arg.(string); ok {
-		arg2Bytes := []byte(s)
-		// convert to bytes if character string without charsetConversion.
+		arg2Bytes = []byte(s)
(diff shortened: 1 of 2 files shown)


mattbooks:
@rashiq @meiji163 @timvaillancourt would love to get your feedback on this fix for https://github.com/github/gh-ost/issues/1568

meiji163:
Thanks for the PR @mattbooks! Can you `gofmt` the files so the linters pass?

mattbooks:
@meiji163 thanks, I believe I have fixed the formatting errors.

meiji163 (approved):

Issue #1569, opened 2025-07-10
gh-ost process stuck infinitely after writer database failure


grodowski:
We detected this issue accidentally when a small `gh-ost` migration got interrupted by a daily test database failover. Since the current writer became read-only, we saw lots of `FATAL Error 1290 (HY000): The MySQL server is running with the --read-only option so it cannot execute this statement` logs and had to step in and kill the process.

Failures like this while copying data in the `ApplyIterationInsertQuery` loop will be retried until `MaxRetries` and then the process will terminate through a `PanicAbort` error.

However, in the case described here, we were past the copy stage and observed `gh-ost` getting stuck in `throttled` state with the heartbeat lag growing indefinitely.

### Repro steps

Partial output printed by an integration test where I issued a `SET GLOBAL super_read_only=1` after the data is copied, but before cutover (`super` just because of how my test db was set up).

First, copying completes.

```
I, [2025-07-10T11:57:02.661777 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Copy: 2000/2000 100.0%; Applied: 0; Backlog: 0/1000; Time: 1s(total), 1s(copy); streamer: mysql-bin.000001:139473498; Lag: 0.01s, HeartbeatLag: 0.07s, State: migrating; ETA: due []
I, [2025-07-10T11:57:02.662144 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Grabbing voluntary lock: gh-ost.2831.lock
I, [2025-07-10T11:57:02.662327 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Setting LOCK timeout as 20 seconds
I, [2025-07-10T11:57:02.662529 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Looking for magic cut-over table
I, [2025-07-10T11:57:02.663413 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Creating magic cut-over table `agent_test`.`_gh_ost_users_20250710115701_del`
I, [2025-07-10T11:57:02.667448 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Magic cut-over table created
I, [2025-07-10T11:57:02.667498 #11938]  INFO -- : [gh-ost]: 2025-07-10 11:57:02 INFO Setting cut-over idle timeout as 30 seconds
[...]

PR #1574, opened 2025-08-06
Panic if InitiateHeartbeat exhausts retries to avoid looping infinitely

grodowski:
Based on experience, if the writer database fails inbeetween the copy & cutover stages (e.g. during cutover pause), the heartbeat writes will fail and stop, then leading to throttled state and an infinite loop of `throttler.shouldThrottle()`.

Since this state is irrecoverable, make the heartbeat writer panic if retries are exhausted, so that the migration can fail and be restarted later.

We've tested this in production now without issues, but if there's some desired behaviour that this PR changes, we might consider adding a new CLI option to control it - feedback welcome.

Closes https://github.com/github/gh-ost/issues/1569 (issue has some more context too)

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -574,6 +574,7 @@ func (this *Applier) InitiateHeartbeat() {
 			continue
 		}
 		if err := injectHeartbeat(); err != nil {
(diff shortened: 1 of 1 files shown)


meiji163:
Good catch @grodowski, I believe I have seen this happen as well.

Issue #1582, opened 2025-09-07
Cant migrate tables with text column unique keys


markmarkmarkthebest:
When running a DDL to migrate a table with a text column that is a unique key gh-ost fails to apply update DML  events.
It fails in the `updateModifiesUniqueKeyColumns` function in applier.go method. That is because text columns are mapped to unit8[] slices which cannot be compared using '!=': https://github.com/github/gh-ost/blob/7cc7cc8d28e17ad1b1eeee24bb0593782db41fed/go/logic/applier.go#L1301
That leads to a connection close and then no more DML events are handled for the table anymore.

markmarkmarkthebest:
go mysql type mapping: https://github.com/go-mysql-org/go-mysql/issues/770

PR #1583, opened 2025-09-07
fix(applier): added support for unique text column, they are mapped to uint8[]

markmarkmarkthebest:
Related issue: https://github.com/github/gh-ost/issues/1582

### Description

This PR changes the `updateModifiesUniqueKeyColumns` function in `applier.go` to use `reflect.DeepEqual` for value comparison to handle unique keys that are text fields because they are mapped to unit8[] slices.

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -8,6 +8,7 @@ package logic
 import (
 	gosql "database/sql"
 	"fmt"
+	"reflect"
 	"regexp"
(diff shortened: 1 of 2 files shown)


markmarkmarkthebest:
I can check if the go type of the column is uint8[] first and handle that case specifically if needed

meiji163 (approved):

PR #1536, opened 2025-04-21
Before the successful renaming, a session accessed the ghost table, w…

abaowhy:
…hich had already unlocked the original table.

Before the successful renaming, a session accessed the ghost table, which had already unlocked the original table. 
There is a very small probability that other sessions dml operations on the original table will occur, 
and this dml operation will appear in the original table after renaming, resulting in data loss.
https://cloud.tencent.com/developer/article/2303777

## A Pull Request should be associated with an Issue.

> We wish to have discussions in Issues. A single issue may be targeted by multiple PRs.
> If you're offering a new feature or fixing anything, we'd like to know beforehand in Issues,
> and potentially we'll be able to point development in a particular direction.

Related issue: https://github.com/github/gh-ost/issues/0123456789

> Further notes in https://github.com/github/gh-ost/blob/master/.github/CONTRIBUTING.md
> Thank you! We are open to PRs, but please understand if for technical reasons we are unable to accept each and any PR

### Description

This PR [briefly explain what it does]
Only when the rename session is waiting for the lock of the original table in performance_schema.metadata_locks, will the lock of the original table be released, and finally rename can be executed. If rename does not wait for the lock, it will exit gh-ost.
> In case this PR introduced Go code changes:

PR #1595, opened 2025-10-13
Resume from checkpoint 

meiji163:
## Description
This PR introduces a checkpoint mechanism that can be used to resume a migration. In combination with `--gtid`, this would allow the user to resume the migration using a different replica. If using file-based coordinates, it requires to resume using the same replica. This is a continuation of @shlomi-noach's POC in https://github.com/github/gh-ost/pull/343. Closes https://github.com/github/gh-ost/issues/205

**Usage**: run gh-ost normally with `--checkpoint` flag. If the migration is interrupted/killed, restart gh-ost with the same arguments with the additional `--resume` flag. By default the checkpoint is every 300 seconds, but can be configured with `--checkpoint-seconds`. Also see [doc/resume.md](https://github.com/github/gh-ost/blob/bcd19da16b08900288ccdc7202ccf5b675162da7/doc/resume.md).

## Details
The two main operations of gh-ost are applying DML events from the binlog and copying rows to the ghost table.
A checkpoint saves the state of both: 
1.  the binlog coordinates of the transaction last applied to the gh-ost table (`LastTrxCoords`)
2. the range last copied to the gh-ost table (`IterationRangeMin` and `IterationRangeMax`)

It is safe to resume the migration from this state because 
1. DML event application is idempotent at the row level. If binlog streamer resumes at coordinates smaller than or equal to the coordinates last processed by the applier, the final values should be the same even if some DML events are applied twice.
2. Copying a row is also idempotent since the second `INSERT` will fail with duplicate key error. Then the DML applier will bring the row up to date like usual

To store the checkpoint we use a new `_ghk` table, which looks like
```sql
CREATE TABLE _${original_tablename}_ghk (
    `gh_ost_chk_id` bigint auto_increment primary key,
    `gh_ost_chk_timestamp` bigint,
    `gh_ost_chk_coords` varchar(4096),
    `gh_ost_chk_iteration` bigint,
    `gh_ost_rows_copied` bigint,
    `gh_ost_dml_applied` bigint,
    `c1_min`,`c2_min`, ...,`cn_min`,
    `c1_max`,`c2_max`, ...,`cn_max`
);
```
where `(c1_min, c2_min..., cn_min)` and `(c1_max, ... cn_max)` are the created with the same types as the shared unique key `(c1, c2, ... cn)` used by gh-ost.

### Replica Tests
I tested resuming with `--test-on-replica` under synthetic sysbench OLTP write load of ~2k DML/sec. I created a sysbench table with 300M rows and ran a no-op migration with `--gtid` and `--checkpoint` set to timeout after 10min. 10 seconds after migration timed out, I started a new `gh-ost` process with `--resume`. When the migration finished the ghost and original tables were checksummed, revealing no data discrepancy. ✅

I repeated this test using an initial timeout of 20min and a waiting period of 1hr before resuming. The data integrity check also passed. In addition the test passed running on two testing replicas in production clusters.

###  Switching Replicas
I tested resuming `gh-ost` using a different replica than the original one it was attach to: 
1. Using the same 300M test table and sysbench write load, I started the migration `gh-ost --alter='add index k_2 (c)' --host='replica1' --gtid --checkpoint`
2. After 10min, I killed the migration 
3. After waiting 5min, I resumed the migration using a second replica: `gh-ost --alter='add index k_2 (c)' --host='replica2' --gtid --checkpoint --resume` 
4. After a few minutes, I killed the sysbench write load (so no DML happens after cutover).
5. Once migration completed, I checksummed the original and ghost table to verify data integrity. ✅ 

### Failover Test
Using the same setup, I tested resuming migration after a master failover triggered by [orchestrator](https://github.com/percona/orchestrator). The failover kills the migration, and I resumed the migration using the same replica. ✅

### DIFF
--- a/doc/resume.md
+++ b/doc/resume.md
@@ -1,0 +1,42 @@
+# Resuming Migrations
+
+`gh-ost` can attempt to resume an interrupted migration from a checkpoint if the following conditions are met:
+- The first `gh-ost` process was invoked with `--checkpoint`
+- The first `gh-ost` process had at least one successful checkpoint
+- The binlogs from the last checkpoint's binlog coordinates still exist on the replica gh-ost is inspecting (specified by `--host`)
+
+To resume, invoke `gh-ost` again with the same arguments with the `--resume` flag.
+
+> [!WARNING]
+> It is recommended use `--checkpoint` with `--gtid` enabled so that checkpoint binlog coordinates store GTID sets rather than file positions. In that case, `gh-ost` can resume using a different replica than it originally attached to.
+
+## Example
+The migration starts with a `gh-ost` invocation such as:
+```shell
+gh-ost \
+--chunk-size=100 \
+--host=replica1.company.com \
+--database="mydb" \
+--table="mytable" \
+--alter="add column mycol varchar(20)"
+--gtid \
+--checkpoint \
+--checkpoint-seconds=60 \
+--execute
+```
+
+In this example `gh-ost` writes a checkpoint to a table `_mytable_ghk` every 60 seconds. After `gh-ost` is interrupted/killed, the migration can be resumed with:
+```shell
+# resume migration
+gh-ost \
+--chunk-size=100
+--host=replica1.company.com \
+--database="mydb" \
+--table="mytable" \
+--alter="add column mycol varchar(20)"
+--gtid \
+--resume \
+--execute
+```
+
+`gh-ost` then reconnects at the binlog coordinates of the last checkpoint and resumes copying rows at the chunk specified by the checkpoint. The data integrity of the ghost table is preserved because `gh-ost` applies row DMLs and copies row in an idempotent way.


smartinec:
The query at the beginning of ReadLastCheckpoint() isn't properly quoted using backticks, so we are getting this error since our DB name has a dash in it:

```
2025-11-21 18:26:41 ERROR No checkpoint found, unable to resume: Error 1064 (42000): You have an error in your SQL syntax; check the manual that corresponds to your MySQL server version for the right syntax to use near '-database._my_table_ghk order by gh_ost_chk_id desc limit 1' at line 1
```

The database name is "my-database".

meiji163:
Thanks @smartinec! fixed in https://github.com/github/gh-ost/pull/1610 . Please let me know if you have any other issues trying out `--checkpoint` 🙇

smartinec:
> Thanks @smartinec! fixed in #1610 . Please let me know if you have any other issues trying out `--checkpoint` 🙇

Did you see my comment on https://github.com/github/gh-ost/pull/1536

PR #1607, opened 2025-11-23
Revertible Migration

meiji163:
## Description

This PR introduces revertible migrations, following the approach suggested by @jonahberquist and outlined in https://github.com/github/gh-ost/issues/302#issuecomment-298843019. In case a migration causes production impact after the cut-over, it can be reverted quickly while preserving the writes that happened after the cut-over. 

**Usage**: When gh-ost is invoked with the `--checkpoint` flag and the migration completes, the migration can be reverted by invoking gh-ost again with the `--revert` flag and the `--old-table` flag specifying the name of the "old" table from the first migration e.g. `_mytable_del`. Also see [docs/revert.md](https://github.com/github/gh-ost/blob/fb3cb43779f9009d1176d78ef755095aa75816c8/doc/revert.md).

Note that the checkpoint table (name ends with `_ghk`) will not be automatically dropped unless `--ok-to-drop-table` is provided.

**Hooks**:  gh-ost hook scripts now include the `GH_OST_REVERT` environment variable with value "true" or "false", indicating if gh-ost is running in revert mode.

Closes https://github.com/github/gh-ost/issues/302.

## Details
When gh-ost is invoked with `--checkpoint` it will write a final checkpoint to the checkpoint (`_ghk`) table after the cut-over is successful. This checkpoint contains the binlog coordinates of the last inserted `AllEventsUpToLockProcessed` row in the changelog (`_ghc`) table.  During the [cut-over](https://github.com/github/gh-ost/blob/1b7933afa466248e682c8487ef1347bf724e02b4/go/logic/migrator.go#L714-L728) stage, `AllEventsUpToLockProcessed` is written to the `_ghc` table after the original table is locked. Therefore no writes after the checkpoint coordinates are applied to the original table.

After the migration completes, the original table is renamed to e.g. `_mytable_del`. Invoking `gh-ost` with `--revert --old-table="_mytable_del"` will read the checkpoint and start application of DML events to `_mytable_del` starting from the checkpoint coordinates. This is possible as long as the binlogs containing these coordinates still exist. The cut-over then proceeds the same as a regular migration, renaming the "old" table to `_mytable_rev_del`. 

In summary, the original table is now reinstated as `mytable` and all writes since the first cut-over have been applied to it. 

## Testing
On a testing MySQL primary-replica cluster, I created a sysbench test table with 5M rows:
```sql
CREATE TABLE `sbtest1` (
  `id` int NOT NULL AUTO_INCREMENT,
  `k` int NOT NULL DEFAULT '0',
  `c` char(120) NOT NULL DEFAULT '',
  `pad` char(60) NOT NULL DEFAULT '',
  PRIMARY KEY (`id`),
  KEY `k_1` (`k`)
)
```
I ran `gh-ost` with `--checkpoint --gtid --alter="drop index k_1"` while under sysbench [oltp_write_only](https://github.com/akopytov/sysbench/blob/master/src/lua/oltp_write_only.lua) workload, with ~3000 qps.

After the cut-over, I waited 60s before starting `gh-ost --revert`. Before the revert migration finished, I killed the sysbench workload. When the revert was complete, I checksummed the original and `_rev_del` tables to verify data integrity ✅  

Next I repeated the test, but started the revert using a different replica than the original migration. ✅

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1722,10 +1722,12 @@ func (this *Migrator) finalCleanup() error {
 		if err := this.retryOperation(this.applier.DropCheckpointTable); err != nil {
 			return err
 		}
-	} else {
-		if !this.migrationContext.Noop {
-			this.migrationContext.Log.Infof("Am not dropping old table because I want this operation to be as live as possible. If you insist I should do it, please add `--ok-to-drop-table` next time. But I prefer you do not. To drop the old table, issue:")
-			this.migrationContext.Log.Infof("-- drop table %s.%s", sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.GetOldTableName()))
+	} else if !this.migrationContext.Noop {
+		this.migrationContext.Log.Infof("Am not dropping old table because I want this operation to be as live as possible. If you insist I should do it, please add `--ok-to-drop-table` next time. But I prefer you do not. To drop the old table, issue:")
+		this.migrationContext.Log.Infof("-- drop table %s.%s", sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.GetOldTableName()))
+		if this.migrationContext.Checkpoint {
+			this.migrationContext.Log.Infof("Am not dropping checkpoint table without `--ok-to-drop-table`. To drop the checkpoint table, issue:")
+			this.migrationContext.Log.Infof("-- drop table %s.%s", sql.EscapeName(this.migrationContext.DatabaseName), sql.EscapeName(this.migrationContext.GetCheckpointTableName()))
 		}
 	}
 	if this.migrationContext.Noop {


danieljoos (approved):
this is great!

tomkrouper:
👋 is there a plan to cut a release with this change? Last release was 2024.

jonahberquist:
Love to see this shipped 🙌🏻

doc/resume.md
```diff
@@ -4,6 +4,7 @@
 - The first `gh-ost` process was invoked with `--checkpoint`
 - The first `gh-ost` process had at least one successful checkpoint
 - The binlogs from the last checkpoint's binlog coordinates still exist on the replica gh-ost is inspecting (specified by `--host`)
+- The checkpoint table (name ends with `_ghk`) still exists
 
 To resume, invoke `gh-ost` again with the same arguments with the `--resume` flag.
```
artemvovk, line 9: not totally related to this PR but to the feature in general: the example below shows `--resume` flag usage, but it is unclear if `--checkpoint` and `--checkpoint-seconds` are necessary when `resume`-ing the migration? They're present for the first run but not for the second. Will gh-ost just "figure out" to continue checkpointing or is it not able checkpoint on `--resume`

PR #1611, opened 2025-12-03
Change coordinates column in checkpoint table to `TEXT`

meiji163:
### Description

This PR changes the `gh_ost_chk_coords` column in the checkpoint table from `VARCHAR(4096)` to ascii `TEXT` column, so it can store GTID sets up to 65,535 characters long.


hemode:
Yes

في الأربعاء، 3 ديسمبر 2025، 4:54 ص Andy Edison ***@***.***>
كتب:

Issue #1615, opened 2026-01-16
aurora rds and performance_schema.setup_instruments error


egorzagorskii:
Hello. Tried to use gh-ost with aurora rds and got an error:

./gh-ost \
  --initially-drop-ghost-table \
  --conf=/etc/mysql-aws/xxx.cnf \
  --host=xxx.eu-central-1.rds.amazonaws.com \
  --database=new_order_01 \
  --table=ride \
  --alter="ADD COLUMN foo VARCHAR(255) NOT NULL DEFAULT ''" \
  --allow-on-master \
  --execute

[2026/01/16 15:39:22] [info] binlogsyncer.go:191 create BinlogSyncer with config {ServerID:99999 Flavor:mysql Host:xxx.eu-central-1.rds.amazonaws.com Port:3306 User:root Password: Localhost: Charset: SemiSyncEnabled:false RawModeEnabled:false TLSConfig:<nil> ParseTime:false TimestampStringLocation:UTC UseDecimal:true RecvBufferSize:0 HeartbeatPeriod:0s ReadTimeout:0s MaxReconnectAttempts:0 DisableRetrySync:false VerifyChecksum:false DumpCommandFlag:0 Option:<nil> Logger:0xc00045a3c0 Dialer:0x7a8120 RowsEventDecodeFunc:<nil> TableMapOptionalMetaDecodeFunc:<nil> DiscardGTIDSet:false EventCacheCount:10240 SynchronousEventHandler:<nil>}
[2026/01/16 15:39:22] [info] binlogsyncer.go:443 begin to sync binlog from position (mysql-bin-changelog.000019, 44488)
[2026/01/16 15:39:22] [info] binlogsyncer.go:409 Connected to mysql 8.0.39 server
[2026/01/16 15:39:22] [info] binlogsyncer.go:868 rotate to (mysql-bin-changelog.000019, 44488)
2026-01-16 15:39:24 ERROR query performance_schema.setup_instruments with name wait/lock/metadata/sql/mdl error: sql: no rows in result set
2026-01-16 15:39:24 ERROR Unable to enable metadata lock instrument, see further error details. Bailing out
2026-01-16 15:39:24 FATAL 2026-01-16 15:39:24 ERROR query performance_schema.setup_instruments with name wait/lock/metadata/sql/mdl error: sql: no rows in result set

i haven't found any related options to skip this check.

meiji163:
[...]

PR #1616, opened 2026-01-16
Add flag --skip-metadata-lock-check

meiji163:
## Description
Since https://github.com/github/gh-ost/pull/1536, `performance_schema.metadata_locks` is required to check the rename session holds the metadata lock on the migrated table during cut-over. On some setups such as Aurora RDS `performance_schema` is not enabled by default and it may be infeasible to enable.
This PR adds `--skip-metadata-lock-check` flag to skip the check in this case.

Although the case described in https://github.com/github/gh-ost/pull/1536 is uncommon, data integrity is top priority for `gh-ost`, so we recommend to enable `performance_schema` if possible. 

Closes https://github.com/github/gh-ost/issues/1615

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -1350,7 +1350,7 @@ func (this *Applier) AtomicCutOverMagicLock(sessionIdChan chan int64, tableLocke
 
 	this.migrationContext.Log.Infof("Session renameLockSessionId is %+v", *renameLockSessionId)
 	// Checking the lock is held by rename session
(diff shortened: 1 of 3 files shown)


susanzhang27 (approved):

saimitta-looker:
@meiji163  Can we include this v1.1.18? v1.1.18-rc introduce metadata checks which require access to performance_schema and option to skip this check should be included in the same release candidate as well to avoid breaking users who can't turn on performance_schema metadata_locks

Issue #1622, opened 2026-01-30
Migration fails with "ghc table does not exist"


ajm188:
This was running off of `master` against an 8.4.6 database.

```
+ ./gh-ost -allow-on-master -attempt-instant-ddl -dml-batch-size 50 -exact-rowcount -force-named-cut-over -force-named-panic -postpone-cut-over-flag-file /tmp/queries/bb2d953d-431d-4bf0-
b6b2-d7ed1a2ae9d0/postpone-cutover -throttle-additional-flag-file /tmp/queries/bb2d953d-431d-4bf0-b6b2-d7ed1a2ae9d0/throttle -timestamp-old-table --skip-metadata-lock-check -database test -alter 'add index idx_user_start_time_end_time (user_id, start_time, end_time)' -table queries -host $hostname -user root -ask-pass
[2026/01/30 17:57:10] [info] binlogsyncer.go:191 create BinlogSyncer with config {ServerID:99999 Flavor:mysql Host:[redacted] Port:3306 User:root Password: Localhost: Charset: SemiSyncEnabled:false RawModeEnabled:false TLSConfig:<nil> ParseTime:false TimestampStringLocation:UTC UseDecimal:true RecvB
ufferSize:0 HeartbeatPeriod:0s ReadTimeout:0s MaxReconnectAttempts:0 DisableRetrySync:false VerifyChecksum:false DumpCommandFlag:0 Option:<nil> Logger:0x40004c23c0 Dialer:0x329d40 RowsEv
entDecodeFunc:<nil> TableMapOptionalMetaDecodeFunc:<nil> DiscardGTIDSet:false EventCacheCount:10240 SynchronousEventHandler:<nil>}
[2026/01/30 17:57:10] [info] binlogsyncer.go:443 begin to sync binlog from position (mysql-bin-changelog.027237, 3002)
[2026/01/30 17:57:10] [info] binlogsyncer.go:409 Connected to mysql 8.4.6 server
[2026/01/30 17:57:10] [info] binlogsyncer.go:868 rotate to (mysql-bin-changelog.027237, 3002)
# Migrating `test`.`queries`; Ghost table is `test`.`_queries_gho`
# Migrating `test`.`queries`; Ghost table is `test`.`_queries_gho`
# Migrating ip-172-17-5-122:3306; inspecting ip-172-17-5-122:3306; executing on bastion-0
# Migration started at Fri Jan 30 17:57:10 +0000 2026
[...]

PR #1623, opened 2026-01-30
fix: add missing error check for WriteChangelogState in initiateApplier

ajm188:
The WriteChangelogState call at line 1365 was missing error handling, which could cause the migration to fail silently if the changelog table was not properly created. This manifested as a "Table doesn't exist" error later in the migration process.

All other calls to WriteChangelogState in the codebase properly check for errors, so I'm guessing this was just overlooked

Relates to https://github.com/github/gh-ost/issues/1622. This doesn't actually solve "why didn't the changelog table get created in my migration" but it does handle the error when trying to write to that table.

### Description

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1362,7 +1362,9 @@ func (this *Migrator) initiateApplier() error {
 				return err
 			}
 		}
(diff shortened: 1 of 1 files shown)


meiji163:
Thanks for catching that!

Issue #1629, opened 2026-02-10
Support Go 1.25


jorendorff:
I had trouble running `script/cibuild` in part because `script/ensure-go-installed` contains a regexp that must match the user's Go version, and the regexp does not match `go1.25.6` (the version I have). That could be fixed by adding a `5` to the regexp:

```sh
SUPPORTED_GO_VERSIONS='go1.2[012345]'
```

Separately, the script's behavior when the regexp doesn't match isn't super helpful. It tries to download and install a local copy of a version of Go, but on Mac this doesn't work. I'm not sure what's going wrong. The content of `.vendor/golocal` after the script runs is like this:

```console
~/src/gh-ost/.vendor/golocal (cibuild) $ ls -lR
total 148264
-rw-r--r--  1 jorendorff  staff      1813 Feb 10 10:00 Distribution
-rw-r--r--  1 jorendorff  staff  75906569 Feb 10 10:00 go1.23.0.darwin-amd64.pkg
drwx------  6 jorendorff  staff       192 Dec 31  1969 org.golang.go.pkg
drwx------  3 jorendorff  staff        96 Dec 31  1969 Resources

./org.golang.go.pkg:
total 154328
-rw-r--r--  1 jorendorff  staff   3936029 Aug  7  2024 Bom
-rw-r--r--  1 jorendorff  staff       569 Feb 10 10:00 PackageInfo
-rw-r--r--  1 jorendorff  staff  75071153 Aug  7  2024 Payload
-rw-r--r--  1 jorendorff  staff       299 Aug  7  2024 Scripts

./Resources:
total 24
-rw-r--r--  1 jorendorff  staff  9552 Feb 10 10:00 background.png
```

PR #1634, opened 2026-02-28
Add support for go 1.25

ggilder:
Related issue: https://github.com/github/gh-ost/issues/1629

### Description

Adds support for go 1.25


ggilder:
FWIW I have run `script/cibuild` and `localtests/test.sh` under go 1.25 (on macOS) and all the tests pass (except for the sysbench performance tests, ran into some unrelated setup issues with sysbench on macOS). However, it looks like CI only runs with the version set in `go.mod` (1.23). Let me know if you think that should be updated to 1.25; we could also change the workflow to use a matrix of go versions.

PR #1648, opened 2026-03-17
Early instant DDL attempt

hasan-dot:
### Description

This PR moves the `--attempt-instant-ddl` check to run **before** ghost table and binlog streaming setup.

---

### Problem

When `--attempt-instant-ddl` is enabled, gh-ost currently performs all the heavy setup work **before** even trying instant DDL:

1. Start binlog streaming
2. Create changelog table
3. Create ghost table + ALTER it
4. **Then** attempt `ALGORITHM=INSTANT` on the original table
5. If it succeeds → clean up the ghost/changelog tables that were just created

For large tables this wastes significant time and resources on setup that gets thrown away when instant DDL succeeds. On extremely large tables (billions of rows), the unnecessary ghost table creation and binlog streaming initialization adds meaningful overhead.

### Solution

Move the instant DDL attempt to **right after** `initiateInspector()` (which discovers the master connection config) but **before** `initiateStreaming()` and `initiateApplier()`. If instant DDL succeeds, the migration completes immediately without ever creating ghost tables, changelog tables, or starting binlog streaming.

### Benchmark results

Benchmarked on M1 Max with MySQL 8.0.42 in Docker (3 iterations each):

| Scenario    | Instant DDL (early) | Full Row Copy | Speedup |
| ----------- | ------------------- | ------------- | ------- |
| Empty table | **25ms**            | 1,123ms       | **45x** |
| 10k rows    | **27ms**            | 2,659ms       | **99x** |
| 100k rows   | **38ms**            | 3,622ms       | **95x** |

Instant DDL is ~25-38ms regardless of table size — it's a metadata-only operation. The speedup grows with data size, and on production tables with millions of rows this would be the difference between milliseconds and hours.

Benchmark gist: https://gist.github.com/hasan-dot/23697ec67401348f7c4d64f60212374b

### CLI flag

| Flag                    | Description                                                                                            |
| ----------------------- | ------------------------------------------------------------------------------------------------------ |
| `--attempt-instant-ddl` | (existing, behavior changed) Try instant DDL **early** before ghost table setup, fall back to row copy |

### Migration flow comparison

**Before (current):**
```mermaid
graph LR
    A[Inspector] --> B[Binlog Streaming]
    B --> C[Ghost Table + Changelog]
    C --> D{Attempt instant DDL}
    D -->|success| E[Done]
    D -->|fail| F[Row Copy]
```

**After (this PR):**
```mermaid
graph LR
    A[Inspector] --> B{Attempt instant DDL}
    B -->|success| C[Done]
    B -->|fail| D[Binlog Streaming]
    D --> E[Ghost Table + Changelog]
    E --> F[Row Copy]
```

### Instant DDL operations (MySQL 8.0+)

These operations complete in milliseconds regardless of table size ([source](https://dev.mysql.com/doc/refman/8.0/en/innodb-online-ddl-operations.html)):
- Adding a column (at end: 8.0.12+; any position: 8.0.29+)
- Dropping a column
- Renaming a column
- Adding/dropping virtual generated columns
- Setting or dropping a column default value
- Modifying the definition of an ENUM or SET column
- Changing the index type
- Renaming a table

### Testing

- [x] `go build ./...` passes
- [x] `go vet ./...` passes
- [x] New integration tests pass (testcontainers + MySQL 8.0)
- [x] Benchmarks confirm expected speedup


ggilder:
@hasan-dot Are you still interested in landing this PR? If not I might take a stab at this.

PR #1656, opened 2026-04-03
chore(script): use GOPATH variable instead of hardcoded path

Ocheretovich:
Replace hardcoded GOPATH path with $GOPATH variable to improve portability of the build script.

### DIFF
--- a/script/build
+++ b/script/build
@@ -14,7 +14,7 @@ version=$(git rev-parse HEAD)
 describe=$(git describe --tags --always --dirty)
 
 export GOPATH="$PWD/.gopath"
-cd .gopath/src/github.com/github/gh-ost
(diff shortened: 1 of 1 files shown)


ericyan:
Hi @Ocheretovich, thanks for the PR.

Looking at the code change, `GOPATH` is set to `$PWD/.gopath` immediately above, this `cd` resolves to the same location as the existing relative path. As a result, I don’t think this change improves configurability or portability; it just expresses the same path differently.

Is there a specific scenario in your environment where the working directory may change before this `cd` runs?

Ocheretovich:
Hi @ericyan You are right, what is missing here is simply a stylistic improvement, the GOPATH omission is specified in the line above and effectively points to the same path as before.

Issue #1660, opened 2026-04-22
Using panic-on-warnings when a table has varbinary columns can fail due to MySQL warning 1300


ggilder:
When gh-ost replays a binlog DML event, the go-mysql library returns varbinary column values as a Go `string` (not `[]byte`). In convertArg, the existing code only converted string → []byte when the column had a non-empty Charset (e.g. utf8mb4 for varchar). varbinary columns have no character set, so Charset is always "", and the string fell through unconverted.

The Go MySQL driver sends `string` args as MYSQL_TYPE_VAR_STRING with the connection's utf8mb4 charset metadata attached, causing MySQL to validate the bytes. If a varbinary value (e.g. a binary UUID) contains byte sequences that are invalid utf8mb4, MySQL emits Warning 1300. With gh-ost's panic-on-warnings enabled, this aborts the migration.

`varbinary` columns are commonly used for UUIDs which have arbitrary bytes and can easily trigger this problem.

PR #1661, opened 2026-04-22
Fix Warning 1300 for varbinary columns with bytes invalid as utf8mb4

ggilder:
Related issue: https://github.com/github/gh-ost/issues/1660

### Description

From the related issue:

> When gh-ost replays a binlog DML event, the go-mysql library returns varbinary column values as a Go `string` (not `[]byte`). In convertArg, the existing code only converted string → []byte when the column had a non-empty Charset (e.g. utf8mb4 for varchar). varbinary columns have no character set, so Charset is always "", and the string fell through unconverted.
>
> The Go MySQL driver sends `string` args as MYSQL_TYPE_VAR_STRING with the connection's utf8mb4 charset metadata attached, causing MySQL to validate the bytes. If a varbinary value (e.g. a binary UUID) contains byte sequences that are invalid utf8mb4, MySQL emits Warning 1300. With gh-ost's panic-on-warnings enabled, this aborts the migration.
>
> `varbinary` columns are commonly used for UUIDs which have arbitrary bytes and can easily trigger this problem.

Fix: add an else-if branch that detects binary storage types by MySQLType (binary, varbinary, *blob) and returns []byte, so the driver sends MYSQL_TYPE_BLOB (binary data) with no charset validation.

MySQLType is used rather than Charset == "" alone because test Column objects built via NewColumnList leave MySQLType unset, which would have changed the return type for all no-charset columns in existing tests. In production, inspect.go always populates MySQLType from information_schema.data_type.

### DIFF
--- a/go/sql/types_test.go
+++ b/go/sql/types_test.go
@@ -95,6 +95,47 @@ func TestConvertArgBinaryColumnPadding(t *testing.T) {
 	require.Equal(t, []byte{0x00, 0x00}, resultBytes[18:])
 }
 
+func TestConvertArgVarbinaryStringWithInvalidUTF8Bytes(t *testing.T) {
+	// go-mysql returns varbinary binlog row values as Go `string` (not `[]uint8`).
+	// When convertArg receives a string for a column with no Charset (varbinary),
+	// it must return []byte — not the original string. The Go MySQL driver sends
+	// string args as MYSQL_TYPE_VAR_STRING with utf8mb4 charset metadata, which
+	// causes MySQL to validate the bytes and emit Warning 1300 for invalid sequences.
+	// gh-ost's panic-on-warnings then turns that warning into a fatal migration error.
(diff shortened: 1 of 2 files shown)


meiji163:
Just ran into this myself 😄 nice timing!

go/sql/types_test.go
```diff
@@ -95,6 +95,47 @@ func TestConvertArgBinaryColumnPadding(t *testing.T) {
...
+	col := Column{
+		Name:      "uuid",
+		Charset:   "",
+		MySQLType: "varbinary(16)", // set by inspect.go from information_schema COLUMN_TYPE
+	}
+
+	result := col.convertArg(rawBytes)
+
+	require.IsType(t, []byte{}, result)
+	require.Equal(t, []byte(rawBytes), result.([]byte))
```
meiji163: appease linter
```suggestion
	require.Equal(t, rawBytes, result.([]byte))
```

meiji163 (approved):

PR #1637, opened 2026-03-04
Prevent permanent worker deadlock when cutover times out waiting for binlog sentinel

VarunChandola:
Fixes #971.

We recently hit a case where gh-ost stalled permanently mid-cutover: heartbeat lag grew
without bound and no further cutover attempts were made despite the process still running.

## Background

As part of atomic cutover, gh-ost locks the original table, waits for all in-flight DML
to be applied to the ghost table, then executes the rename. The "wait for in-flight DML"
step works via a sentinel value written to the changelog table. This sentinel travels via
the binlog stream back to gh-ost, which applies binlog events sequentially through a
single-threaded `applyEventsQueue`. The sentinel is enqueued onto that same queue, so by
the time the applier executes it, all preceding DML is guaranteed to have been applied:

```go
// onChangelogStateEvent — spawns a goroutine to enqueue the sentinel func
var applyEventFunc tableWriteFunc = func() error {
    this.allEventsUpToLockProcessed <- &lockProcessedStruct{...}
    return nil
}
go func() {
    this.applyEventsQueue <- newApplyEventStructByFunc(&applyEventFunc)
}()
```

`waitForEventsUpToLock` then blocks waiting to receive on `allEventsUpToLockProcessed`,
which only fires once the applier has worked through the queue and executed the sentinel.

## The bug

Under high DML load the queue backlog is large enough that the sentinel isn't executed
before `CutOverLockTimeoutSeconds` expires. This is subtle: the logs show the sentinel
being *intercepted* quickly, which is easy to misread as the sentinel having been
delivered. `Handled changelog state AllEventsUpToLockProcessed` only means
`onChangelogStateEvent` spawned the goroutine to push onto `applyEventsQueue` — not that
the applier has executed it or that `allEventsUpToLockProcessed` has been written to.

From our logs:
```
09:56:49.616  Writing changelog state: AllEventsUpToLockProcessed:177240580961633980
09:56:49.616  Tables locked
09:56:49.620  Intercepted changelog state AllEventsUpToLockProcessed   ← binlog event seen
09:56:49.620  Handled changelog state AllEventsUpToLockProcessed       ← goroutine spawned
09:56:49.623  Waiting for events up to lock
09:56:52.617  Timeout while waiting for events up to lock              ← 3s later, timed out
09:56:52.617  Dropping magic cut-over table
09:56:53.053  Tables unlocked
09:56:53.058  Looking for magic cut-over table                         ← no further attempts
```

[...]

PR #1666, opened 2026-04-30
Fix OOM when allEventsUpToLockProcessed buffer equals MaxRetries()

dnovitski:
## Problem

PR #1637 buffered `allEventsUpToLockProcessed` to `MaxRetries()` to prevent a goroutine deadlock when `waitForEventsUpToLock` times out during cutover. However, when `--default-retries` is set to a very large value (e.g. `9999999999999`), Go tries to allocate a channel with trillions of buffer slots, causing an immediate OOM crash before the migration even starts.

### Crash

```
fatal error: runtime: out of memory

runtime.makechan(...)
github.com/github/gh-ost/go/logic.NewMigrator(...)
    go/logic/migrator.go:116
```

## Fix

Replace the `MaxRetries()`-sized buffer with a **buffer of 1** and **overwrite-oldest (latest-wins) send semantics**.

When the buffer is full (receiver timed out on a previous attempt), the stale message is drained before sending the current sentinel:

```go
select {
case this.allEventsUpToLockProcessed <- lps:
default:
    // Buffer full — drain stale value, then send current one
    select { case <-this.allEventsUpToLockProcessed: default: }
    select { case this.allEventsUpToLockProcessed <- lps: default: }
}
```

This guarantees:
- ✅ **Current sentinel is always delivered** — stale messages are drained first
- ✅ **`executeWriteFuncs` worker never blocks** — the send is always non-blocking
- ✅ **Constant memory** — buffer of 1 regardless of `MaxRetries()`
- ✅ **No extra timeouts** — unlike a drop-newest approach, the current attempt's message is preserved

### Why drop-newest would be wrong

A naive non-blocking send (`select { default: }`) would drop the **current** sentinel when the buffer is full. Since `waitForEventsUpToLock` only succeeds on an exact challenge match, dropping the current message causes a false timeout — potentially while the table is locked during two-step cutover.

Overwrite-oldest avoids this: the stale message (which would only be skipped by the receiver anyway) is drained to make room for the current one.

## Tests

- **Overwrite-oldest test**: sends two sentinels without consuming the first, verifies the channel contains only the latest message
- **Extreme MaxRetries regression test**: sets `MaxRetries()` to `9999999999999` and verifies `NewMigrator` succeeds with channel capacity 1

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -111,9 +111,10 @@ func NewMigrator(context *base.MigrationContext, appVersion string) *Migrator {
 		ghostTableMigrated:       make(chan bool),
 		firstThrottlingCollected: make(chan bool, 3),
 		rowCopyComplete:          make(chan error),
-		// Buffered to MaxRetries() to prevent a deadlock when waitForEventsUpToLock times out.
-		// (see https://github.com/github/gh-ost/pull/1637)
-		allEventsUpToLockProcessed: make(chan *lockProcessedStruct, context.MaxRetries()),
+		// Buffered with capacity 1; the send uses overwrite-oldest semantics
+		// to prevent both deadlock (see https://github.com/github/gh-ost/pull/1637)
+		// and OOM when MaxRetries() is extremely large.
+		allEventsUpToLockProcessed: make(chan *lockProcessedStruct, 1),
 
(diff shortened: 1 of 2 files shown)


meiji163 (approved):

meiji163:
LGTM 👍

Issue #1670, opened 2026-05-06
throttle() blocks forever when context is cancelled during abort


ggilder:
## Summary

When `abort()` is called (e.g. due to heartbeat failures after a MySQL failover), it cancels the migration context and causes `initiateThrottlerChecks` to exit via `ctx.Done()`. However, `initiateThrottlerChecks` exits without calling `SetThrottled(false, ...)`, leaving `isThrottled` permanently `true`.

The `throttle()` loop only checks `IsThrottled()` and has no awareness of context cancellation, so it spins indefinitely — preventing the migration goroutine from ever returning and leaving the process deadlocked.

## Reproduction scenario

1. Migration is running and throttled (e.g. due to replica lag)
2. A MySQL failover occurs — the applier's connection starts hitting `Error 1290: read-only`
3. `injectHeartbeat` fails `MaxRetries()` times → `PanicAbort` is sent → `abort()` is called
4. Context is cancelled → `initiateThrottlerChecks` exits via `ctx.Done()`
5. `isThrottled` remains `true` — the goroutine that would set it `false` has exited
6. `throttle()` loops forever; the migration process never exits

## Observable symptoms

- Status line keeps printing with `Lag:` and `HeartbeatLag:` counting up (other goroutines still alive)
- `State: throttled, lag=Xs` is frozen at the lag value from when throttle first triggered
- Process will not terminate without `SIGKILL`

## Root cause

`throttle()` uses `time.Sleep(250ms)` with no `ctx.Done()` check:

```go
func (thlr *Throttler) throttle(onThrottled func()) {
    for {
        if shouldThrottle, _, _ := thlr.migrationContext.IsThrottled(); !shouldThrottle {
            return
        }
        if onThrottled != nil {
            onThrottled()
        }
        time.Sleep(250 * time.Millisecond) // no ctx.Done() check
    }
}
```

When the goroutine responsible for calling `SetThrottled(false, ...)` exits due to context cancellation, nothing else can unblock this loop.

PR #1671, opened 2026-05-06
Prevent throttle() from blocking forever when context is cancelled

ggilder:
Related issue: https://github.com/github/gh-ost/issues/1670

### Description

This PR fixes a deadlock where a migration process becomes permanently stuck after `abort()` is called (e.g. following heartbeat failures due to a MySQL failover).

When the context is cancelled, `initiateThrottlerChecks` exits via `ctx.Done()` without calling `SetThrottled(false)`, leaving `isThrottled` permanently `true`. The `throttle()` loop had no awareness of context cancellation — it only checked `IsThrottled()` — so it would spin indefinitely, preventing the migration goroutine from ever returning.

The fix replaces `time.Sleep(250ms)` with a `select` that also listens on `ctx.Done()`, so `throttle()` unblocks immediately when the migration is aborted.

### DIFF
--- a/go/logic/throttler.go
+++ b/go/logic/throttler.go
@@ -504,6 +504,8 @@ func (thlr *Throttler) initiateThrottlerChecks() {
 // throttle sees if throttling needs take place, and if so, continuously sleeps (blocks)
 // until throttling reasons are gone
 func (thlr *Throttler) throttle(onThrottled func()) {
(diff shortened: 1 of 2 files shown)


ggilder:
CI is passing on my fork here: https://github.com/ggilder/gh-ost/pull/7

meiji163 (approved):

PR #1679, opened 2026-05-18
Add query builders for separate SELECT+INSERT copy queries

danieljoos:

Related issue (internal): https://github.com/github/database-infrastructure/issues/8163

### Description

This PR is part of the move-tables feature development PR stack.
It introduces query builders for separate `SELECT` + `INSERT` queries that are used for the move-tables mode instead of the range-`INSERT` query that usually runs on the primary host.
For move-tables the copy step needs to first `SELECT` on the source, then `INSERT` on the target. Of course this comes with much higher cost: copying the data to and from the host that runs `gh-ost`.


danieljoos:
Closing, using a feature branch instead of stacked PRs.

Issue #1669, opened 2026-05-06
Slow pagination for composite primary keys


coding-chimp:
We have several tables with composite primary keys `(shop_id, id)`. We noticed that queries used to determine the next chunk can sometimes run very slowly. Here's an example:

```sql
explain analyze select `shop_id`, `id`
    -> from `orders`
    -> where ((`shop_id` > 12345) or (((`shop_id` = 12345)) AND (`id` > 23456))) and ((`shop_id` < 65432) or (((`shop_id` = 65432)) AND (`id` < 98765)) or ((`shop_id` = 65432) and (`id` = 98765)))
    -> order by `shop_id` asc, `id` asc
    -> limit 1 offset 999\G
*************************** 1. row ***************************
EXPLAIN: -> Limit/Offset: 1/999 row(s)  (cost=142 rows=1) (actual time=7785..7785 rows=1 loops=1)
    -> Filter: (((orders.shop_id > 12345) or ((orders.shop_id = 12345) and (orders.id > 23456))) and ((orders.shop_id < 65432) or ((orders.shop_id = 65432) and (orders.id < 98765)) or ((orders.id = 98765) and (orders.shop_id = 65432))))  (cost=142 rows=1352) (actual time=7785..7785 rows=1000 loops=1)
        -> Covering index scan on orders using PRIMARY  (cost=142 rows=1352) (actual time=0.0621..6706 rows=21.4e+6 loops=1)

1 row in set (7.79 sec)
```

We tried simplifying the query with a row constructor expression, but that did not change the performance:

```sql
explain analyze select `shop_id`, `id`
    -> from `orders`
    -> where (`shop_id`, `id`) > (12345, 23456) and (`shop_id`, `id`) <= (65432, 98765)
    -> order by `shop_id` asc, `id` asc
    -> limit 1 offset 999\G
*************************** 1. row ***************************
EXPLAIN: -> Limit/Offset: 1/999 row(s)  (cost=45.4 rows=1) (actual time=7520..7520 rows=1 loops=1)
    -> Filter: (((orders.shop_id,orders.id) > (12345,23456)) and ((orders.shop_id,orders.id) <= (65432,98765)))  (cost=45.4 rows=1000) (actual time=7520..7520 rows=1000 loops=1)
        -> Covering index scan on orders using PRIMARY  (cost=45.4 rows=1000) (actual time=0.0616..6661 rows=21.4e+6 loops=1)

1 row in set (7.52 sec)
```

[...]

PR #1686, opened 2026-05-21
Improve performance for tables with composite primary keys

coding-chimp:

Related issue: https://github.com/github/gh-ost/issues/1669

### Description

This PR introduces an optimized query pattern for tables with composite primary keys. As described in the linked issue, gh-ost's queries against these tables can currently be very slow. We've seen a single query take up to 16 minutes, resulting in slow migrations that might not even finish when gh-ost falls behind in binlog streaming. The new query pattern has a constant query time (~30ms in our testing) regardless of the table's data shape. We're only introducing it for two-column composite keys because the complexity would grow exponentially with more columns.

We're introducing the new query pattern in `BuildRangeInsertQuery`, `BuildUniqueKeyRangeEndPreparedQueryViaOffset`, `BuildUniqueKeyRangeEndPreparedQueryViaTemptable`.

I've tested the changes against real data on a test DB, and the migration took ~12h to complete. In comparison, I'm currently running the same migration again without these changes, and it has only reached 15.6% after 22h. The current ETA is 87h.


go/sql/builder.go
```diff
@@ -323,6 +334,88 @@ func BuildRangeInsertQuery(databaseName, originalTableName, ghostTableName strin
...
+	col1Name := EscapeName(cols[0].Name)
+	col2Name := EscapeName(cols[1].Name)
+	col1StartVal := rangeStartValues[0]
+	col2StartVal := rangeStartValues[1]
+	col1EndVal := rangeEndValues[0]
+	col2EndVal := rangeEndValues[1]
+	col2StartOp := string(minRangeComparisonSign)
+	fromClause := fmt.Sprintf("%s.%s force index (%s)", databaseName, originalTableName, uniqueKey)
+
+	if sameFirstColumnValue(rangeStartArgs, rangeEndArgs) {
```
coding-chimp, line 378: `sameFirstColumnValue` reports whether the first column's range start and end bounds are equal. When true, the range fits within a single `col1` partition, and a simpler query must replace the 3-part UNION.

It's not just a performance shortcut, it's a correctness requirement. Look at what the UNION produces when `col1_start == col1_end`:

  - Part 1: `col1 = x AND col2 > start_col2` - no upper bound on `col2`, so it includes all rows, including rows above `end_col2`
  - Part 2: `col1 > x AND col1 < x` - returns nothing
  - Part 3: `col1 = x AND col2 <= end_col2` - no lower bound, so it includes rows below `start_col2`

The `UNION ALL` result is essentially every row where `col1 = x`. The outer `ORDER BY ... LIMIT 1 OFFSET chunkSize-1` then picks the Nth row, which is the wrong boundary.

The simplified query fixes this:

`WHERE col1 = ? AND col2 > ? AND col2 <= ?`

Both bounds are explicit, and because `col1` is pinned with equality, MySQL can use the composite index directly as a seek on `(col1, col2)`.

go/sql/builder.go
```diff
@@ -442,6 +545,200 @@ func BuildUniqueKeyRangeEndPreparedQueryViaTemptable(databaseName, tableName str
...
+		databaseName, tableName, hint,
+		m.col1Name, m.col2Name,
+		part1, part2, part3,
+		m.orderByAsc,
+		chunkSize-1,
+	)
+	return result, explodedArgs, nil
+}
+
+func buildUniqueKeyRangeEndTwoColumnViaTemptable(
```
coding-chimp, line 684: The UNION optimization is arguably unnecessary for the `ViaTemptable` code path, since it'll only be called once at the end of the migration, when the query is unlikely to be slow. However, it provides consistency with `ViaOffset` and doesn't add much code.

meiji163 (dismissed):

meiji163 (approved):

olegkv:
@meiji163 @coding-chimp 
Replacing index range scan with UNION ALL is not free, MySQL will materialize UNION ALL into temp table, sort it and then apply limit and offset. The previous query, when works well is a single range scan with no sort (using index order) and no temptable. So, buildUniqueKeyRangeEndTwoColumnViaOffset will also create a temp table and sort it, same as buildUniqueKeyRangeEndTwoColumnViaTemptable.
In general, MySQL is able to merge several index ranges into one operation:
https://dev.mysql.com/doc/refman/8.4/en/range-optimization.html
So, it seems that bad query plans are not a norm, but the result of data skewing in a particular case. 
If so, this change looks more like additional feature, not necessarily needed for general use case.
Maybe this change should be under a feature flag, so it can be enabled and disabled?

Update: I ran an experiment on large table with ~30 columns and two-column PK. The query with UNION ALL was several times slower than the regular query that uses two ranges and OR, something like 3ms (OR) vs 20ms (UNION ALL)

PR #1699, opened 2026-06-05
Don’t parse binlog events of unrelated tables

coding-chimp:
### Description

This change avoids decoding row payloads for DML events on tables that gh-ost is not listening to.

Previously, gh-ost decoded every row event from the binlog before filtering by database/table name in the event streamer. This meant DML activity on unrelated tables still paid the full row decoding cost, even though those events were discarded later.

With this change, gh-ost now uses go-mysql’s `RowsEventDecodeFunc` hook to inspect the row event header/table-map metadata first. If no registered listener is interested in the event’s database/table, gh-ost skips decoding the row data entirely.

### DIFF
--- a/go/binlog/gomysql_reader.go
+++ b/go/binlog/gomysql_reader.go
@@ -21,6 +21,24 @@ import (
 	uuid "github.com/google/uuid"
 )
 
+type RowsEventFilterFunc func(databaseName, tableName string) bool
+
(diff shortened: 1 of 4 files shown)


meiji163 (approved):

meiji163:
Nice optimization @coding-chimp! Can you pull in the master branch?

Issue #1681, opened 2026-05-18
Move table mode for gh-ost


danieljoos:
This document specifies the work needed inside gh-ost to support a new **move table mode**, in which one or more tables are copied from a source MySQL cluster to a target MySQL cluster, with a cooperative cutover protocol coordinated via hooks.

## Scope

**In scope:**

- One or more tables in a single gh-ost run, sharing one source binlog stream.
- GTID-only operation.
- gh-ost issues the source `RENAME TABLE` itself; ACL changes and feature-flag flips are delegated to hook scripts.

**Explicitly deferred or out of scope:**

- Schema transformation (`--alter`) combined with move.
- File+position binlog drain (non-GTID).
- Inline checksumming or target-binlog monitoring (use `pt-table-checksum` externally).
- Revert / rollback of a completed move.
- Source-side grant cleanup.

## Design

Move table mode is enabled by a single new flag `--move-tables=t1,t2,...`. The mode reuses the existing inspector, binlog streamer, chunk iterator, throttler, hook executor, and checkpoint mechanism, while introducing a second database connection (the target), per-table runtime state, and a cooperative cutover protocol driven by gh-ost.

### Architecture overview

```
┌──────────────────────┐                    ┌──────────────────────┐
│     Source MySQL     │                    │     Target MySQL     │
│                      │                    │                      │
│  ┌─────────┐         │                    │  ┌─────────┐         │
│  │ table_A │◄── Inspector reads ──┐       │  │ table_A │         │
│  │ table_B │                      │       │  │ table_B │         │
│  └─────────┘         │            │       │  └─────────┘         │
│                      │            │       │       ▲              │
│  Binlog ────────────────► EventsStreamer  │       │              │
│                      │            │       │   Applier writes ────┘
│                      │        gh-ost      │                      │
│                      │       process      │  ┌─────────┐         │
│                      │            │       │  │  _ghk    │        │
│                      │            └───────│──│checkpoint│        │
│                      │                    │  └─────────┘         │
└──────────────────────┘                    └──────────────────────┘
```

**Key principle:** limit changes to the hot code path (eg. binlog streamer). All move-table-specific logic lives in higher layers — the migrator (cutover orchestration and multi-table iteration) and the applier (where it connects and what tables it writes to).

### Connection model

| Connection | Standard mode | Move table mode |
|---|---|---|
| Inspector | reads from source | reads from source (unchanged) |
| Binlog streamer | reads from source | reads from source (unchanged) |
[...]

PR #1700, opened 2026-06-05
refactor(move-tables): clean up prototype phase 1.2 skip-tables (#8206)

womoruyi:
Productionize the prototype-hacked phase 1.2 implementation (skip ghost/changelog/heartbeat in move-tables mode) into clean, tested, reviewable code matching the design doc acceptance criteria. No behavior change for existing gh-ost users — `move-tables` mode itself is opt-in via the `--move-tables` flag.

- Ref: [database-infrastructure#8206](https://github.com/github/database-infrastructure/issues/8206) ([Task] [1.2] Skip ghost/changelog tables, heartbeat in gh-ost move-tables mode)
- Parent epic: [database-infrastructure#8204](https://github.com/github/database-infrastructure/issues/8204) ([Epic] [Phase 1] extend gh-ost to support single table moves)
- Boundary issue (NOT in this PR): [database-infrastructure#8207](https://github.com/github/database-infrastructure/issues/8207) ([Task] [1.3] CRUD new table on target cluster) — target-table creation code remains as prototype-hacked, will be productionized separately
- Design: [move_table_mode.md §1.2](https://github.com/github/gh-ost-tablemove-poc/blob/9dc6df75c4c88ff473906a497836c7518f5614ec/design/move_table_mode.md#12-skip-ghost-table-changelog-table-and-heartbeat)

### What approach did you choose and why?

The prototype on `move-tables/1.2-skip-ghost-tables` wired phase 1.2 into the existing `Migrate()` machinery via inline `IsMoveTablesMode()` guards (see @chriskirkland's `TODO(chriskirkland): work left to do` block at `migrator.go:772-779`). It worked end-to-end but was not reviewable as #8206's deliverable. This PR productionizes it.

**Scope (#8206 only — the skips):**

1. **Replaced `WriteChangelog` bypass TODO with intentional comment.** `applier.go:756`. The single low-level chokepoint is the right design — there are three callers in move-tables mode (`ReadMigrationRangeValues`, `printStatus`, `finalCleanup`) and gating each individually would scatter guards and drift when new callers are added. Comment now documents why the bypass lives where it does.

2. **Removed demo hack code.** Four artifacts marked TEMPORARY / TODO(replace) by @chriskirkland for local demo:
   - `simulateMoveTablesCutover()` function (`migrator.go:933-960`)
   - `TmpCutoverFilename` polling loop in `MoveTables()` (`migrator.go:889-916`)
   - `TmpCutoverFilename` field on the `MoveTables` struct (`context.go:282`)
[...]

PR #1703, opened 2026-06-10
Merge move-tables PRs onto feature branch: applier logic

danieljoos:

Related issue (internal): https://github.com/github/database-infrastructure/issues/8175
Related issue (internal): https://github.com/github/database-infrastructure/issues/8163

### Description

This PR is part of the move-tables feature development.
It combines 3 PRs from the previous stacked PRs.

It introduces query builders for separate `SELECT` + `INSERT` queries that are used for the move-tables mode instead of the range-`INSERT` query that usually runs on the primary host.

For move-tables the copy step needs to first `SELECT` on the source, then `INSERT` on the target. Of course this comes with much higher cost: copying the data to and from the host that runs `gh-ost`.

It introduces a new function `ApplyIterationMoveTableCopyQueries` to the `Applier` that will be used for copying rows in the move-tables mode.

It adapts the `Applier` to write DML events to the target database when in move-tables mode.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1724,7 +1724,12 @@ func (mgtr *Migrator) iterateChunks() error {
 					// _ghost_ table, which no longer exists. So, bothering error messages and all, but no damage.
 					return nil
 				}
-				_, rowsAffected, _, err := mgtr.applier.ApplyIterationInsertQuery()
+				var rowsAffected int64
+				if mgtr.migrationContext.IsMoveTablesMode() {
+					_, rowsAffected, _, err = mgtr.applier.ApplyIterationMoveTableCopyQueries()
+				} else {
+					_, rowsAffected, _, err = mgtr.applier.ApplyIterationInsertQuery()
+				}
 				if err != nil {
 					return err // wrapping call will retry
 				}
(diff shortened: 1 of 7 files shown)


go/logic/migrator.go
```diff
@@ -1724,7 +1724,12 @@ func (mgtr *Migrator) iterateChunks() error {
 					// _ghost_ table, which no longer exists. So, bothering error messages and all, but no damage.
 					return nil
 				}
-				_, rowsAffected, _, err := mgtr.applier.ApplyIterationInsertQuery()
+				var rowsAffected int64
```
womoruyi, line 1727: @danieljoos quick boundary question based on the Copilot comment on these same lines . [#8207](https://github.com/github/database-infrastructure/issues/8207) (target table CRUD),  is under my purview from (#1700 and #1704), and also touches `MoveTables()` initialization, . 

I was wondering if you were planning to wire `prepareQueries()` into `MoveTables()` here as a follow-up commit, or does that stay open for whoever picks up the next sub-issue that touches initialization (would be [#8207](https://github.com/github/database-infrastructure/issues/8207) on my side)? Either is fine, just want to know whether the scope expands on my end!

zacharysierakowski (approved):

PR #1708, opened 2026-06-12
Move tables 1.6 crash safe resume

zacharysierakowski:
closes https://github.com/github/database-infrastructure/issues/8210

Related issue (internal): https://github.com/github/database-infrastructure/issues/8210

### Description

This PR updates move-tables mode to be crash-resume safe. It manages the `checkpoint` table life, and a backgrounded checkpoint process as well as introduces new checkpoint data specifically for the move-tables cutover. Per the [docs on crash safe resume across cutover](https://github.com/github/gh-ost-tablemove-poc/blob/design-docs/design/move_table_mode.md#16-crash-safe-resume-across-cutover), we now store that the move-tables cutover has started (a bool) and the move-tables cutover drain GTID. We use these two attributes for `--move-tables --resume` to pickup if the process failed mid-cutover. It's safe for failure between T1-T3 (the rename and the drain) and also safe for failure post drain, pre hook call (between T3-T5). I simple `--resume` will pick back up in the cutover protocol and ensure the rename isn't re-run, the drain is complete, and the complete hook is fired. 

### Demo

https://github.com/user-attachments/assets/943ef4fc-5c0d-4f94-b243-f4801a3e6e36

### DIFF
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -374,7 +374,12 @@ func main() {
 			log.Fatal("--postpone-cut-over-flag-file must be specified when using --move-tables")
 		}
 		if !migrationContext.Checkpoint {
-			log.Fatal("--checkpoint must be specified when using --move-tables")
+			log.Infof("--move-tables requires checkpointing; enabling --checkpoint")
+			migrationContext.Checkpoint = true
+		}
+		if !migrationContext.UseGTIDs {
+			log.Infof("--move-tables requires GTID coordinates for cutover drain checks; enabling --gtid")
+			migrationContext.UseGTIDs = true
 		}
 		migrationContext.MoveTables.TableNames = strings.Split(*moveTables, ",")
 		for i := range migrationContext.MoveTables.TableNames {
--- a/go/cmd/gh-ost/main.go
+++ b/go/cmd/gh-ost/main.go
@@ -373,6 +373,9 @@ func main() {
 		if migrationContext.PostponeCutOverFlagFile == "" {
 			log.Fatal("--postpone-cut-over-flag-file must be specified when using --move-tables")
 		}
+		if !migrationContext.Checkpoint {
+			log.Fatal("--checkpoint must be specified when using --move-tables")
+		}
 		migrationContext.MoveTables.TableNames = strings.Split(*moveTables, ",")
 		for i := range migrationContext.MoveTables.TableNames {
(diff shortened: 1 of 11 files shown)


go/cmd/gh-ost/main.go
```diff
@@ -373,6 +373,14 @@ func main() {
...
 			log.Fatal("--postpone-cut-over-flag-file must be specified when using --move-tables")
 		}
+		if !migrationContext.Checkpoint {
+			log.Infof("--move-tables requires checkpointing; enabling --checkpoint")
+			migrationContext.Checkpoint = true
+		}
+		if !migrationContext.UseGTIDs {
+			log.Infof("--move-tables requires GTID coordinates for cutover drain checks; enabling --gtid")
+			migrationContext.UseGTIDs = true
+		}
```
zacharysierakowski, line 383: I chose to _not_ fail the request without these, and just enable them. Could also change it to a `Fatal` and force the operator to rerun with the required flags. But that feels kinda silly if they are always required.

danieljoos (approved):

go/logic/applier.go
```diff
@@ -1024,6 +1087,51 @@ func (apl *Applier) ReadLastCheckpoint() (*Checkpoint, error) {
...
+		drainGTID, err := mysql.NewGTIDBinlogCoordinates(drainGTIDStr)
+		if err != nil {
+			return nil, err
+		}
+		chk.MoveTablesCutOverDrainGTID = drainGTID
+	}
+	return chk, nil
+}
+
+func (apl *Applier) ReadMoveTablesCutOverCheckpoint() (*Checkpoint, error) {
```
chriskirkland, line 1100: [question] why can't we re-use `ReadLastCheckpoint` for this?  the only significant difference I see is that the original fn is L1053-4 which we should able to ignore.

I assume there's something subtle I'm missing 🤔

chriskirkland (approved):

go/logic/applier.go
zacharysierakowski: I think we technically could have. There might have been a reason at some point during my commit history but I agree it looks like it would be safe to use `ReadLastCheckpoint`. The big difference is that `ReadMoveTablesCutOverCheckpoint` has a where clause for `gh_ost_move_tables_cutover_started = 1` and to make sure the drain GTID is set.  But the one callsite i see of `ReadMoveTablesCutOverCheckpoint` does that check in the code as well 

https://github.com/github/gh-ost/blob/f5273fa23bab1c3c0346632a6a499c42fb55e6a2/go/logic/migrator.go#L1020-L1024

PR #1709, opened 2026-06-12
Move Tables: throttle based on target credentials

danieljoos:

Related issue (internal): https://github.com/github/database-infrastructure/issues/8212

### Description

This PR adapts the throttler to use the target credentials in move-tables mode.
Status variables are fetched from the target server and control replicas are expected to be in the target cluster.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -1809,17 +1809,13 @@ func (mgtr *Migrator) addDMLEventsListener() error {
 
 // initiateThrottler kicks in the throttling collection and the throttling checks.
 func (mgtr *Migrator) initiateThrottler() {
-	if mgtr.migrationContext.IsMoveTablesMode() {
-		// TODO(chriskirkland): throttle against the target cluster
-		mgtr.migrationContext.Log.Info("Skipping throttling in move tables mode")
-		return
-	}
-
 	mgtr.throttler = NewThrottler(mgtr.migrationContext, mgtr.applier, mgtr.inspector, mgtr.appVersion)
 
 	go mgtr.throttler.initiateThrottlerCollection(mgtr.firstThrottlingCollected)
 	mgtr.migrationContext.Log.Infof("Waiting for first throttle metrics to be collected")
-	<-mgtr.firstThrottlingCollected // replication lag
+	if !mgtr.migrationContext.IsMoveTablesMode() {
+		<-mgtr.firstThrottlingCollected // replication lag
+	}
 	<-mgtr.firstThrottlingCollected // HTTP status
 	<-mgtr.firstThrottlingCollected // other, general metrics
 	mgtr.migrationContext.Log.Infof("First throttle metrics collected")
--- a/go/logic/throttler.go
+++ b/go/logic/throttler.go
@@ -176,6 +176,17 @@ func (thlr *Throttler) collectReplicationLag(firstThrottlingCollected chan<- boo
 	}
 }
 
+// controlReplicaConnectionConfig returns the connection config used to read
+// replication lag from a control replica. In move-tables mode the lag is read
+// from the target cluster's replicas, otherwise from the source (inspector)
+// cluster's replicas.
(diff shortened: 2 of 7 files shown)


go/logic/migrator.go
```diff
@@ -1636,17 +1636,13 @@ func (mgtr *Migrator) addDMLEventsListener() error {
...
 	mgtr.throttler = NewThrottler(mgtr.migrationContext, mgtr.applier, mgtr.inspector, mgtr.appVersion)
 
 	go mgtr.throttler.initiateThrottlerCollection(mgtr.firstThrottlingCollected)
 	mgtr.migrationContext.Log.Infof("Waiting for first throttle metrics to be collected")
-	<-mgtr.firstThrottlingCollected // replication lag
+	if !mgtr.migrationContext.IsMoveTablesMode() {
+		<-mgtr.firstThrottlingCollected // replication lag
+	}
 	<-mgtr.firstThrottlingCollected // HTTP status
 	<-mgtr.firstThrottlingCollected // other, general metrics
```
zacharysierakowski, line 1820: Can you explain why only this first `firstThrottlingCollected` is gated behind move tables mode? Can't really tell just looking at this diff

zacharysierakowski: Oh I see down below we aren't emitting `collectReplicationLag` when in move tables. What's the reason for that?

zacharysierakowski: oh okay, reading into the code more is this because `collectReplicationLag` relies on the heartbeat table in some way? And we don't do the heartbeat in move tables

go/logic/throttler.go
```diff
@@ -443,7 +455,9 @@ func (thlr *Throttler) collectGeneralThrottleMetrics() error {
 // that may affect throttling. There are several components, all running independently,
 // that collect such metrics.
 func (thlr *Throttler) initiateThrottlerCollection(firstThrottlingCollected chan<- bool) {
-	go thlr.collectReplicationLag(firstThrottlingCollected)
+	if !thlr.migrationContext.IsMoveTablesMode() {
+		go thlr.collectReplicationLag(firstThrottlingCollected)
+	}
 	go thlr.collectControlReplicasLag()
```
zacharysierakowski, line 469: Do we need to gate this one behind move tables mode? Quick glance at the code of `collectControlReplicasLag`, it looks like it relies on the changelog table which we also don't have in move-tables mode right?

danieljoos: ~yes, the `collectReplicationLag` is disabled for move-tables mode.~

~In that function, gh-ost typically collects the replication lag from the replica it is attached to (reading the binlog data).
So for migrations it then knows the "impact" of the copy-writes it did on the corresponding primary - and if these are causing too much stress on the replica it is attached to. It then begins to throttle automatically.~

~For move-tables we can't use that mechanism, as we're only connected to the target primary and not reading the binlog on the target side.~

sorry misread your question. I'll look into that.

go/logic/migrator.go
danieljoos: Yes, some of the `collect...Lag` goroutines in `throttler.go` get that `firstThrottlingCollected` channel handed over and write to it as soon as they're set up.
As we're skipping the `collectReplicationLag` goroutine for move-tables, we also must not await the channel, here. Otherwise it would block forever at the `initializeThrottler` func.

go/logic/throttler.go
danieljoos: Adapted the code to get the `SecondsBehind` from the replica status instead when in move-tables mode.

zacharysierakowski (approved):
Changes look good here AFAICT based on the diff. Since we're getting close to code complete on https://github.com/github/database-infrastructure/issues/8204, i'm thinking it probably makes sense to start looking into getting better integration test coverage (re [stage 3 in the move-tables design](https://github.com/github/gh-ost-tablemove-poc/blob/design-docs/design/move_table_mode.md#stage-3--tests-and-documentation)). Hoping we might be able to get _some_ sort of better coverage for the throttling changes.

PR #1718, opened 2026-06-19
Introduce a SourcePrimaryConnectionConfig for primary-required move-table ops

zacharysierakowski:
re: https://github.com/github/database-infrastructure/issues/8275

### Description

This PR does a few things that were lingering on our backlog.

1. Primarily, it adds a separate connection specifically for the _source primary_. We **need** this for things like the write cutover RENAME, as well as the post-move-tables cleanup for DROP the source renamed table (if specified by the cli arg)
2. It also condenses the write cutover T1 and T2 ops so that it's safer by running the RENAME and GTID gather in a _single_ DB query using a multi-statement. This is safer because it lessens the chance of crash between the RENAME and the GTID query. And still ensures we're running the two on the same connection.
3. Adds a guard to ensure the operator is providing a pointer to the _primary_  for the TARGET. All target operations require to run against the primary (unlike the source) so this is just a way to provide an early exit
4. Adds another guard to ensure the operator is providing a pointer to a _replica_ (if possible) for the SOURCE. If there are no replicas, or the operator is absolutely sure they want to use the primary, they can pass  `--allow-on-source-primary` to bypass this guard

#### Auto resolve source primary from source replica

image

#### [Guard] Early exit if target is a replica

image

#### [Guard] Early exit if source is a _primary_ and have not specified the new override arg (--allow-on-source-primary)

image


script/move-tables/reset
zacharysierakowski, line 1: random script i've had stashed for a while that's been handy between local testing so i don't have to do a full teardown/setup.

happy to remove from this PR if we don't want it merged.

go/logic/streamer.go
```diff
@@ -240,32 +239,6 @@ func (es *EventsStreamer) Close() (err error) {
...
 
-// DropSourceOldTable drops the source "__del" table in move-tables mode. The
-// __del table is the post-cutover rollback handle on the source cluster; it is
-// only dropped after a successful run when --ok-to-drop-table is set.
-// The applier's dropTable targets the move-tables target cluster,
-// so the source-side drop is owned by the streamer: its `db`
-// handle uses InspectorConnectionConfig (the source) and stays open in both the
-// normal and the cutover-resume paths (Close() only closes the binlog reader,
-// not `db`).
-func (es *EventsStreamer) DropSourceOldTable() error {
```
zacharysierakowski, line 251: This was only added recently as part of https://github.com/github/gh-ost/pull/1717. I added it knowing that it _probably_ wasn't the right spot for it, but without having the distinction between the primary vs. replica connections (this PR) it was hard to figure out the best spot for this drop operation to dangle from

go/logic/migrator.go
```diff
@@ -1660,6 +1712,126 @@ func (mgtr *Migrator) initiateServer() (err error) {
...
+// terminology. Safe to call once per run.
+func (mgtr *Migrator) setupMoveTablesSourcePrimary(dbVersion string) error {
+	cfg, err := mgtr.resolveSourcePrimaryConnectionConfig(dbVersion)
+	if err != nil {
+		return fmt.Errorf("failed to resolve move-tables source primary: %w", err)
+	}
+	mgtr.migrationContext.MoveTables.SourcePrimaryConnectionConfig = cfg
+
+	uri := cfg.GetDBUri(mgtr.migrationContext.DatabaseName) + "&multiStatements=true"
+	db, _, err := mysql.GetDB(mgtr.migrationContext.Uuid, uri)
```
danieljoos, line 1789: There is a `mysql.GetDBVersion()` function that retrieves and caches the version for the given URI.
Might be better to get the actual DB version of the primary instead of relying on the replica having the same version as the primary.
For us, it might not be a big deal though, as all of our hosts are at least MySQL 8.0.

danieljoos (approved):

PR #1722, opened 2026-06-22
Multi-flavor docker replica-test harness

conf:
### Background

> _Make the change easy, then make the easy change.
> Kent Beck_

Hey 👋 
First of all, thanks for the awesome tool, it was a real saver for our quite big db with not-so-great schema, I really appreciate it.

This is a pre-requisite for adding MariaDB support to gh-ost, which comes in the next PR if this one goes well.
I decided to ship this set of changes first because it was quite hard to untangle the needs of different MySQL flavours.

### Mandatory AI disclaimer
Obviously, I heavily used Claude during the work on this PR, but it's not just some blindly vibe-coded change. 
I did a manual code review (several times) and tried to keep the amount of changes to the minimum. 
Also it is worth to mention that I heavily tested it locally across all versions specified in Github Actions.

### tl;dr
Run the tests with
```bash
TEST_MYSQL_IMAGE=mysql:8.4.3 ./script/docker-gh-ost-replica-tests up
TEST_MYSQL_IMAGE=mysql:8.4.3 ./script/docker-gh-ost-replica-tests run
TEST_MYSQL_IMAGE=mysql:8.4.3 ./script/docker-gh-ost-replica-tests down # don't forget to specify TEST_MYSQL_IMAGE
```
`TEST_MYSQL_IMAGE` can be any MySQL flavour specified in `.github/workflows/replica-tests.yml`.

### Description
I tried to make replica-test harness to be easily extendable across different MySQL versions and flavours, such as Percona and (not yet available here) MariaDB, so in case there's a need to support some other version with slightly different values, only `docker/script` folder will need some tuning and not the whole testing harness.

<hr>
Rework the replica-test harness to run the localtests suite against MySQL 5.7, 8.0, 8.4 and Percona Server 8.0 from one parameterized setup.

- Move docker-compose into script/ and split per-version server config into docker/<flavor>-<major.minor>/ dirs (common.cnf, create_user.sql, start_replication.sql), selected from TEST_MYSQL_IMAGE. Optional flavor.env per dir for knobs (Percona reads /etc/my.cnf.d).
- Configure replication from the harness once both servers are up rather than at container init, so that replica connects on the first try.
- Keep mysql_native_password for 5.7/8.0/percona, caching_sha2_password for 8.4.
- Speed-ups: native mysql client with docker-exec fallback, detect server version once, shorter replication polls, configurable copy-throttle window (THROTTLE_SECONDS, default 2s to still exercise binlog DML replay).
- toxiproxy split into its own compose file.


meiji163 (approved):

meiji163:
HI 👋 thanks for your contribution! It would be great to officially support MariaDB

conf:
Wow, that was super fast! Thanks!

PR #1724, opened 2026-06-22
feat: 🦭 MariaDB support 🦭

### LINKED ISSUE #725: [question] MariaDB compatibility
Hi,
We're going to update our MariaDB servers up to v10.3 and probably play with 10.4 because of cool instant alter feature.
In the meantime, gh-ost is one of the most critical tools we rely (thank you!).
We've played with gh-ost and MariaDB 10.3 and everything looks fine. But to be sure, could you please say what current status of compatibility is with MariaDB?
Is it safe to use gh-ost with MariaDB?
(Didn't find any notes about it)

Thanks.

conf:
Closes #725.

## 🦭 MariaDB support 🦭

The PR is split in 2 commits:
- MariaDB 11.x support
- MariaDB 10.x support

Both commits pass the test suite and can be merged separately (if needed).

### MariaDB 11.x support
gh-ost mostly works against MariaDB unchanged; this commit makes the few
MariaDB-specific notable spots correct and runs the replica test suite against
MariaDB 11.4 and 11.8 (LTS) alongside the MySQL/Percona flavors.

Core:
- ReplicaTermFor keeps the legacy slave/master replication terminology on
  MariaDB. MariaDB reports versions >= 10 (greater than the 8.4 cutoff) but
  never adopted the new replica/source terms, and its SHOW REPLICA STATUS
  still emits legacy column names, so the MySQL 8.4 term map doesn't apply.
- Accept BINLOG MONITOR as well as REPLICATION CLIENT in the grant check
  (MariaDB 10.5+ renamed the privilege).

Test harness (localtests):
- test.sh: pick slave/master terminology for MariaDB, wait for replica
  catch-up via MASTER_GTID_WAIT/gtid_binlog_pos, and skip gtid_mode-gated
  tests where @@gtid_mode is unsupported instead of aborting the run.
- Resolve the per-flavor config dir by specificity (docker/<flavor>-<major.minor>,
  then <flavor>-<major>, then bare <flavor>) so versions that share one config
  live in a single dir. All MariaDB versions use one docker/mariadb dir.
- Make the zero-date failure tests deterministic across flavors with an
  explicit sql_mode (NO_ZERO_IN_DATE,NO_ZERO_DATE,...); they relied on MySQL's
  default sql_mode, which MariaDB doesn't share.
- Drop the incidental NOT NULL from the generated-column tests (MariaDB
  forbids NOT NULL on generated columns) so they run on MariaDB; keep
  generated-columns-unique skipped (it puts a generated column in the PRIMARY
  KEY, which MariaDB forbids, Error 1903).
- Add mariadb:11.4.12 and mariadb:11.8.8 to the CI matrix.

## MariaDB 10.x support
Add the MariaDB 10.x LTS lines (10.5, 10.6, 10.11) to the replica test suite.

MariaDB versions below 11.1 need one core fix: gh-ost passes transaction_isolation as
a DSN param, which the driver applies as `SET transaction_isolation=<v>` on each
connection. That system variable does not exist on MariaDB < 11.1 (which has
only tx_isolation), so the connection fails with Error 1193 "Unknown system
variable 'transaction_isolation'". tx_isolation was in turn removed in MySQL
8.0, so no single variable name is portable across the supported servers.

- mysql.OpenDB strips the transaction_isolation param and instead runs the
  SQL-standard `SET SESSION TRANSACTION ISOLATION LEVEL <level>` on each new
  connection via a connector wrapper, accepted by every supported MySQL and
  MariaDB version. GetDBUri still emits the param unchanged, so DSN
  construction and its tests are unaffected.
- Add mariadb:10.5.29, mariadb:10.6.27 and mariadb:10.11.18 to the CI matrix.
<hr>


.github/workflows/replica-tests.yml
```diff
@@ -10,7 +10,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        image: ['mysql:5.7.41','mysql:8.0.41','mysql:8.4.3','percona/percona-server:8.0.41-32']
+        image: ['mysql:5.7.41','mysql:8.0.41','mysql:8.4.3','percona/percona-server:8.0.41-32','mariadb:10.5.29','mariadb:10.6.27','mariadb:10.11.18','mariadb:11.4.12','mariadb:11.8.8']
```
meiji163, line 13: Can we restrict the CI tests to one MariaDB version, or one for 10.x and one for 11.x?

meiji163:
This is great! 🚀 🚀

meiji163 (changes requested):

.github/workflows/replica-tests.yml
conf: They're all [LTS releases](https://endoflife.date/mariadb), except 10.5, which is EOL already. If you think it's too much, I can remove some of them, let me know what you think.

meiji163: should be fine actually 👍  I'm wondering what versions should be officially supported? I don't have experience with MariaDB in production

meiji163 (approved):

meiji163 (approved):

.github/workflows/replica-tests.yml
conf: I picked the versions based on what [AWS currently supports](https://docs.aws.amazon.com/AmazonRDS/latest/UserGuide/MariaDB.Concepts.VersionMgmt.html#MariaDB.Concepts.VersionMgmt.Supported), i.e. several versions from every LTS line, except 10.5. I decided to include 10.5 too since it's MariaDB's equivalent of MySQL 5.7, which is also covered by gh-ost's CI.

conf:
🎉 🎉 🎉 Awesome!  🎉 🎉 🎉 
Thanks a lot!!!

PR #1726, opened 2026-06-23
Move-tables. Multi-table

zacharysierakowski:
re: https://github.com/github/database-infrastructure/issues/8205

## Multi-table `--move-tables`

Extends `feature-move-tables` to move **N tables in one run** (`--move-tables=a,b,c`): one binlog stream, per-table interleaved row copy, a single per-run crash-safe checkpoint, and **one atomic multi-table RENAME at cutover**. A single table is just N=1 — no "representative" table, no index-0 special-casing.

**Key decisions:** per-table state containers (`MoveTable`) instead of a representative table; run-wide artifacts named from a stable hash of the table set (`_gho_<token>_ghk`); one checkpoint table with a hex-serialized per-table key range; GTID-based drain + atomic `RENAME TABLE t1 TO …, t2 TO …`. Single-table-only paths **fail fast** (panic/error) in move-tables mode.

**Read in this order:**
1. `go/base/context.go` — `MoveTable` + `InitMoveTableContainers`, `OrderedMoveTables`, `MoveTablesRunToken`, checkpoint/`_del` naming *(start here)*
2. `applier.go` — per-table query builders + `buildDMLEventQuery` routing; target-table create/validate
3. `migrator.go: iterateChunksMoveTables` — interleaved per-table copy
4. `checkpoint.go` + applier `Write/ReadMoveTableCheckpoints` + migrator resume — checkpoint redesign
5. `migrator.go: moveTablesCutOver` / `drainMoveTablesCutOver` / `resume…FromCheckpoint` — atomic cutover
6. The cross-cutting **guards** (context/inspect/applier/migrator) — the "no representative table" invariant
7. hooks / server / throttler — multi-table awareness
8. tests (`localtests/move-tables/*`, `*_test.go`) + `script/move-tables/*`

**Validated:** unit (checkpoint serialize/run-token), integration (1/2/3-table idle + DML, resume-after-crash ×3, atomic-cutover), and manual e2e on 2 clusters / 3–4 tables with live writes.

## Demo

https://github.com/user-attachments/assets/bcb477fd-0f96-4f15-9792-c892ac817187


go/base/context.go
```diff
@@ -418,6 +557,9 @@ func getSafeTableName(baseName string, suffix string) string {
 // GetGhostTableName generates the name of ghost table, based on original table name
 // or a given table name
 func (mctx *MigrationContext) GetGhostTableName() string {
+	if mctx.IsMoveTablesMode() {
+		panic("GetGhostTableName() must not be called in move-tables mode; there is no ghost table (the target keeps each migrated table's name)")
```
zacharysierakowski, line 561: I chose to add panics to function calls that I felt were dangerous enough if they accidentally were called by a codepath from `--move-tables`. Happy to discuss alternatives, or ditching the panics altogether, but they gave me confidence while making the changes across this PR. 

One thought - we _could_ swap the raw panics for `failpoint` panics and only enable them for integration tests if we wanted to get fancier while removing intentional panics from the prod binary. But I don't know if that gains us _that_ much.

go/base/context.go
```diff
@@ -1189,6 +1354,84 @@ func (mctx *MigrationContext) IsMoveTablesMode() bool {
...
+//   - deterministic: the same table set always yields the same token, so a
+//     resumed run finds the same run-wide artifacts (e.g. the checkpoint table).
+//   - order-independent: --move-tables=a,b and --move-tables=b,a match.
+//   - fixed-length: independent of how many tables are moved (so it never blows
+//     past identifier length limits the way a concatenation of names would).
+//
+// It is used to name run-wide singular artifacts (checkpoint table, applier
+// advisory lock, serve socket) so they never depend on any single migrated
+// table name. Returns "" outside move-tables mode.
+func (mctx *MigrationContext) MoveTablesRunToken() string {
```
zacharysierakowski, line 1401: This is the func that determines the unique identifier per run based on the table names. More details in the comment above but just wanted to flag it directly with a comment. 

A non-start alternative could have been concating all the table names, but that would make for a _really_ long identifier in most cases. 

Another alternative could have been to require [ForceTmpTableName](https://github.com/github/gh-ost/blob/64a6e3ddfd5de895e40ebd7bb0715c6166fc03a2/go/cmd/gh-ost/main.go#L186) in `--move-tables` mode, so that the operator chooses what the identifier for the table is. We'd potentially need to extend that a bit for other callsites like the server sock file, etc.

go/logic/applier.go
```diff
@@ -809,12 +959,6 @@ func (apl *Applier) CreateCheckpointTable() error {
 		"`gh_ost_dml_applied` bigint",
 		"`gh_ost_is_cutover` tinyint(1) DEFAULT '0'",
 	}
-	if apl.migrationContext.IsMoveTablesMode() {
-		colDefs = append(colDefs,
-			"`gh_ost_move_tables_cutover_started` tinyint(1) DEFAULT '0'",
-			"`gh_ost_move_tables_drain_gtid` text charset ascii",
-		)
-	}
 	for _, col := range apl.migrationContext.UniqueKey.Columns.Columns() {
```
zacharysierakowski, line 962: we had originally made these conditional changes to `CreateCheckpointTable` in our feature branch for single move. These removals just get it back to baseline (`main`) since we have our own `CreateCheckpointTable` variant for move-tables now `createMoveTablesCheckpointTable`

go/logic/checkpoint.go
zacharysierakowski, line 1: all the funky hex serialization stuff lives in this diff.

alternative could have been to create a checkpoint table _per_ table, but there is shared state across the checkpoints that's valuable to keep in a single table. This way, on crash-resume when mid cutover, we can easily just query the last checkpoint row.

go/logic/hooks.go
zacharysierakowski, line 1: no major changes here, just some guards against function calls we don't want called in move-tables mode (panic)

go/logic/throttler.go
```diff
@@ -193,12 +193,15 @@ func (thlr *Throttler) collectControlReplicasLag() {
...
-	)
+	var replicationLagQuery string
+	if !thlr.migrationContext.IsMoveTablesMode() {
+		replicationLagQuery = fmt.Sprintf(`
+			select value from %s.%s where hint = 'heartbeat' and id <= 255
+			`,
+			sql.EscapeName(thlr.migrationContext.DatabaseName),
+			sql.EscapeName(thlr.migrationContext.GetChangelogTableName()),
+		)
+	}
```
zacharysierakowski, line 204: These lines just build a query that's only run in NOT move tables mode (below). However, building the query here calls `GetChangelogTableName` which is now guarded with the panic. So this just wraps the query building in the move tables check

localtests/move-tables/atomic-multi-table-cutover/tables.txt
zacharysierakowski, line 1: @chriskirkland and I talked about making some changes to the _existing_ integration tests for the crash/resume so that they cover multi-table as well.

go/base/context.go
```diff
@@ -76,6 +79,139 @@ func NewThrottleCheckResult(throttle bool, reason string, reasonHint ThrottleRea
 	}
 }
 
+// MoveTable holds the per-table runtime state for a single table within a
```
chriskirkland, line 82: 👍🏼

go/base/context.go
```diff
@@ -1189,6 +1354,84 @@ func (mctx *MigrationContext) IsMoveTablesMode() bool {
...
+// or nil if it has not been initialized.
+func (mctx *MigrationContext) GetMoveTable(tableName string) *MoveTable {
+	if mctx.MoveTables.Tables == nil {
+		return nil
+	}
+	return mctx.MoveTables.Tables[tableName]
+}
+
+// OrderedMoveTables returns the per-table containers in --move-tables order.
+// Iteration must always use this deterministic order, never the Tables map's
```
chriskirkland, line 1378: ❤️

go/logic/applier.go
```diff
@@ -224,7 +277,20 @@ func buildMigrationLockName(db, table string) string {
 // preventing two gh-ost processes from migrating the same table concurrently
 // on the same MySQL server.
 func (apl *Applier) AcquireMigrationLock(ctx context.Context) error {
-	lockName := buildMigrationLockName(apl.migrationContext.GetTargetDatabaseName(), apl.originalTableName())
+	// One advisory lock per run. In move-tables mode it is keyed on the
+	// set-derived run token (not any single table) so two processes moving the
+	// same set of tables collide, while a single-table run keeps its table-keyed
```
chriskirkland, line 282:
```
	// One advisory lock per run. In move-tables mode it is keyed on the
	// set-derived run token (not any single table) so two processes moving the
	// same set of tables collide, while a single-table run keeps its table-keyed
```

On it's face, I think this implies that intersecting move-table sets could be migrated at the same time. I think the CREATEs for the overlapping tables on the target would break for all but the first run (assuming they use the same database/schema).  Not sure if this is a realizable concern at present... **is there anything else preventing that kind of conflict?**

Entirely possible this is covered by one of the design docs.

go/base/context.go
```diff
@@ -76,6 +79,139 @@ func NewThrottleCheckResult(throttle bool, reason string, reasonHint ThrottleRea
...
+//
+// The range/iteration fields are guarded by the per-table rangeMutex. The
+// applier-wide "current applied source coordinates" mutex stays single — there
+// is one applied stream feeding all tables.
+type MoveTable struct {
+	// Identity.
+	SourceDatabaseName string
+	SourceTableName    string
+	TargetDatabaseName string
+	TargetTableName    string
```
chriskirkland, line 96: _Open handed opinion..._

Currently we have these as invariants in how move-tables works:
* table name is the same on the source and target
* database is the same for all tables in the same move-table set

Having source/target versions of each inside of `MoveTable` is convenient but might be misleading and/or make the code harder to reason about.  It looks like they're mostly used in the applier rather than the MT itself, so we might be able to store less state here.

A clarifying comment might be enough here. 🤷🏼

go/logic/applier.go
```diff
@@ -1028,7 +1207,140 @@ func (apl *Applier) WriteCheckpoint(chk *Checkpoint) (int64, error) {
...
+// reflects the freshest run-wide state.
+func (apl *Applier) WriteMoveTableCheckpoints(rows []*Checkpoint) error {
+	if len(rows) == 0 {
+		return nil
+	}
+	escaped := make([]string, len(moveTablesCheckpointColumns))
+	for i, c := range moveTablesCheckpointColumns {
+		escaped[i] = sql.EscapeName(c)
+	}
+	placeholders := "(" + strings.TrimSuffix(strings.Repeat("?, ", len(moveTablesCheckpointColumns)), ", ") + ")"
```
chriskirkland, line 1239:
```suggestion
	placeholders := "(" + strings.Join(strings.Repeat("?", len(moveTablesCheckpointColumns), ", ") + ")"
```

go/logic/migrator.go
```diff
@@ -803,32 +807,82 @@ func (mgtr *Migrator) Revert() error {
...
+		mt.RowsEstimate = rowsEstimate
+		mt.CreateTableStatement = createStatement
+		totalRowsEstimate += rowsEstimate
+	}
 
-	// In move-tables mode source and target schemas match, so shared columns are identical.
-	mgtr.migrationContext.SharedColumns = mgtr.migrationContext.OriginalTableColumns
-	mgtr.migrationContext.MappedSharedColumns = mgtr.migrationContext.OriginalTableColumns
+	// Aggregate the row estimate across all tables for overall progress reporting.
+	atomic.StoreInt64(&mgtr.migrationContext.RowsEstimate, totalRowsEstimate)
```
chriskirkland, line 859: I think there's an existing bug here.  If a table has primary key of type `UNSIGNED BIGINT`, it would need `uint64` to avoid overflow here.  That's not super likely to happen with single table, but more likely with multi table.

I don't think we'll be in the business of moving > 2^64 rows in a single cohort, but we might want to file an issue or something 🤷🏼 

cc @danieljoos

chriskirkland (approved):
Looks great! No major concerns (for anything introduced here).  Just some questions/comments that are non-blocking

go/base/context.go
```diff
@@ -429,11 +571,13 @@ func (mctx *MigrationContext) GetGhostTableName() string {
 	}
 }
 
-// GetTargetTableName generates the name of the target table, based on original table name and
-// the migration context (i.e. move-tables mode).
+// GetTargetTableName generates the name of the target table. In move-tables mode
+// each table keeps its own name on the target, so there is no single target
+// table name; per-table code uses MoveTable.TargetTableName instead, and calling
+// this is a programmer error that panics to fail fast.
 func (mctx *MigrationContext) GetTargetTableName() string {
```
danieljoos: This function has only been introduced by the move-tables mode.
Maybe we could just remove it again?

go/logic/applier.go
```diff
@@ -492,9 +603,13 @@ func (apl *Applier) tableExists(tableName string) (tableFound bool) {
 	return (m != nil)
 }
 
+// originalTableName returns the single migrated table. It is a representative
+// accessor that has no meaning in move-tables mode (every table is handled
+// through its own MoveTable container), so calling it there is a programmer
+// error and panics to fail fast rather than silently operate on the wrong table.
 func (apl *Applier) originalTableName() string {
```
danieljoos, line 610: Same here, the `originalTableName()` func was introduced by our move-table changes. Maybe we can just get rid of it again.

go/logic/inspect.go
```diff
@@ -131,7 +137,7 @@ func (isp *Inspector) InspectOriginalTable() (err error) {
 
 func (isp *Inspector) originalTableName() string {
```
danieljoos, line 138: and another one here: the `originalTableName()` method has been introduced with move-tables.
Whole func could removed, imo.

go/logic/migrator.go
danieljoos: I think we'll never reach more than 2^63 rows even across multiple tables.
Not even sure if that's possible at all in InnoDB.

danieljoos (approved):

go/logic/applier.go
zacharysierakowski: Great callout! I should have mentioned this as an implication to the choice here. 

But yea at the very start it checks _each_ table to ensure they aren't created on the target. I felt like that was enough to protect us. 

Simple alt would be to create multiple locks if we want to be extra safe?

go/base/context.go
zacharysierakowski: Hmmm, happy to chat about this further but my gut says - since it's already separated, leaving it separated is not only easier for us right now, but could also be easier for someone in the future if they wanted to support a table move that actually targets a different DB/table-name? 

Let mw know what you think but can definitely start with a comment.

go/logic/applier.go
zacharysierakowski: the suggestion here doesn't compile. I think if we wanted to change it, we'd need `slices.Repeat`:

```
	placeholders := "(" + strings.Join(slices.Repeat([]string{"?"}, len(moveTablesCheckpointColumns)), ", ") + ")"
```

i don't really have a strong opinion here though lol lean towards just leaving it unless you prefer the Join/slices.Repeat

go/logic/migrator.go
zacharysierakowski: hmmm. I can change `RowsEstimate` to a uint64 if we are concerned enough. The usage of that field seems minimal so probably low risk / easy effort.

go/base/context.go
zacharysierakowski: Oh sweet, great call! Done in 87385247fda7121685a20f905c0dc6b847c3fa1b

go/logic/applier.go
zacharysierakowski: I don't see as much harm in leaving this one here. There's a lot of callsites to it from our original PR that introduces it for single-table move. And if i remove it, it means all those callsites switch to `apl.migrationContext.OriginalTableName` directly and we lose our guard against calling accidentally calling `OriginalTableName` in move-tables mode

go/logic/inspect.go
zacharysierakowski: same comment as [here](https://github.com/github/gh-ost/pull/1726#discussion_r3481770375)

go/logic/applier.go
chriskirkland: Yeah, I think we don't have to overcomplicate. Mostly wanted to make sure I understand the protections here.

go/base/context.go
chriskirkland: > since it's already separated, leaving it separated is not only easier for us right now, but could also be easier for someone in the future if they wanted to support a table move that actually targets a different DB/table-name?

I buy that 👍🏼

chriskirkland (approved):

Issue #1728, opened 2026-06-25
Released assets naming scheme is confusing


conf:
## Description
I'm trying to automate installing `gh-ost` in my Dockerfile. And current assets naming scheme is a bit confusing and makes automation unnecessarily harder.
Consider this screenshot:

<img width="1112" height="444" alt="Image" src="https://github.com/user-attachments/assets/47a423f4-523e-47f6-8dca-00a58ea1c53b" />

- `gh-ost` turns out to be Darwin-arm64, but it's not really obvious from the name
- RPM and Deb packages have release version embedded in their file name, while other files do not do that.
- RPM's embedded version is different from Deb's version, i.e. `1.1.10-1` vs `1.1.10`
- Biggest painpoint - TAR archives have generated timestamp in their names which makes automating harder - instead of just downloading from latest release, you need to list the available assets inside a release first and only then you can pick the one you need.
- TAR archives have `-binary` suffix
- Also, RPM and Deb packages lack ARM64 versions

## Proposed naming scheme
- Do not embed timestamps in file names
- Do not embed release versions in file names. Github download url already includes release version, i.e. https://github.com/github/gh-ost/releases/download/v1.1.10/gh-ost-1.1.10-1.x86_64.rpm - second version embedding of 1.1.10 seems excessive, we can do just https://github.com/github/gh-ost/releases/download/v1.1.10/gh-ost.x86_64.rpm
- Provide ARM64 versions of Deb and RPM packages
- Strip `-binary` suffix from TAR archives
- Not sure what to do with generic `gh-ost` - rename to `gh-ost-darwin-arm64` (would make os/arch obvious), make it universal binary (pack amd64 and arm64 in one binary and name it `gh-ost-darwin`), or just delete since all other versions already cover all choices.

I'd love to supply a PR if we agree this needs to be fixed.
Cheers!

meiji163:
I agree with changing the release naming if you want to open a PR!

conf:
@meiji163 Done, please, have a look at #1729.

conf:
This can be closed since #1729 is merged.

PR #1729, opened 2026-06-25
build: standardize release asset naming

conf:

Related issue: https://github.com/github/gh-ost/issues/1728

### Description

This PR reworks the release packaging so the published assets have stable,
predictable names that are easy to automate against (e.g. installing a fixed
version from a `Dockerfile` without first listing a release's assets), and
slims down the packaging tooling along the way.

#### Asset naming (`build.sh`)

- **Drop the generated timestamp** from tarball names — the biggest pain point,
  since it forced consumers to list a release's assets before they could pick a
  download.
- **Drop the `-binary` infix** from tarball names.
- **Name macOS tarballs with `darwin`** (matches `GOOS`) instead of `osx`.
- **Drop the version from `.rpm`/`.deb` filenames** — the release download URL
  already carries it (`.../download/v1.1.10/gh-ost.x86_64.rpm`). Package
  *metadata* still records the version and epoch.
- **Build arm64 `.rpm` and `.deb`** in addition to amd64, using each
  ecosystem's arch convention (`x86_64`/`aarch64` for rpm, `amd64`/`arm64`
  for deb).
- **Stop publishing the bare `gh-ost` binary** — it was just the last build
  (darwin/arm64) left over in the build dir, and its name made the OS/arch
  non-obvious. Tarballs and packages already cover every OS/arch.
- **Emit a `SHA256SUMS` manifest** so a download can be verified with
  `sha256sum -c SHA256SUMS` (or `--ignore-missing` for a single file).

Resulting assets:

```
gh-ost-linux-amd64.tar.gz    gh-ost.x86_64.rpm    gh-ost.amd64.deb
gh-ost-linux-arm64.tar.gz    gh-ost.aarch64.rpm   gh-ost.arm64.deb
gh-ost-darwin-amd64.tar.gz
gh-ost-darwin-arm64.tar.gz   SHA256SUMS
```

#### Packaging tooling (`Dockerfile.packaging`, `script/dock`)

- Collapse the per-package `RUN apt-get install` layers into one, drop packages
  already provided by the `golang:bookworm` base (git, tar, curl, gcc, g++,
  bash) and the unused `rsync`, and drop the legacy GOPATH `src/` layout (the
  build is module-mode). Only `ruby`/`ruby-dev`/`build-essential` (for fpm) and
  `rpm` (for rpmbuild) are installed now.
- Use a multi-stage build with a `scratch` artifacts stage so
  `docker build --output type=local` exports the assets straight to the host.
  This drops the intermediate container, bind mount, `-it` TTY and
  `find | xargs cp` step from `script/dock pkg`.
- Fix the stale/misleading usage text in `script/dock`.

#### Verification

Built all assets via `script/dock pkg` and verified: every tarball/package
contains a binary of the expected arch, the rpm/deb arch metadata is correct,
the deb version is `1:1.1.10` (epoch preserved), `sha256sum -c SHA256SUMS`
round-trips green, and the exported directory contains only the release assets.

> :warning: **Note for the release-upload step:** `SHA256SUMS` has no `gh-ost`
> prefix, so it will not match a `gh-ost*` upload glob. The upload step needs to
> include it explicitly (or upload the whole `/tmp/gh-ost-release` directory).

### DIFF
--- a/Dockerfile.packaging
+++ b/Dockerfile.packaging
@@ -1,20 +1,19 @@
-FROM golang:1.25.9-bookworm
+FROM golang:1.25.9-bookworm AS build
 
-RUN apt-get update
-RUN apt-get install -y ruby ruby-dev rubygems build-essential
-RUN gem install fpm
-ENV GOPATH=/tmp/go
+# build.sh shells out to fpm to produce the .deb/.rpm packages. fpm is a Ruby
+# gem (needs ruby + ruby-dev + a C toolchain), and the rpm target needs
+# rpmbuild from the `rpm` package. Everything else build.sh uses (git, tar,
+# gcc, bash, ...) already ships in the golang:bookworm base image.
+RUN apt-get update \
+    && apt-get install -y --no-install-recommends ruby ruby-dev build-essential rpm \
+    && rm -rf /var/lib/apt/lists/* \
+    && gem install --no-document fpm
 
-RUN apt-get install -y curl
-RUN apt-get install -y rsync
-RUN apt-get install -y gcc
-RUN apt-get install -y g++
-RUN apt-get install -y bash
-RUN apt-get install -y git
-RUN apt-get install -y tar
-RUN apt-get install -y rpm
-
-RUN mkdir -p $GOPATH/src/github.com/github/gh-ost
-WORKDIR $GOPATH/src/github.com/github/gh-ost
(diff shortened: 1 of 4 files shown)


meiji163 (approved):

conf:
Thanks a lot! Any ETA for the release?

Issue #1735, opened 2026-07-02
Deadlock during cleanup when --attempt-instant-ddl succeeds (GhostTableMigrated signal never drained)


peterbollen:
### Overview

When `--attempt-instant-ddl` is used and the instant `ALTER` succeeds, gh-ost can deadlock during final cleanup and hang forever instead of exiting.

### Root cause

`initiateApplier` writes a `GhostTableMigrated` changelog row. The streamer's changelog listener callback (`Migrator.onChangelogStateEvent`) publishes that signal synchronously via `base.SendWithContext(ctx, mgtr.ghostTableMigrated, true)` **while holding `EventsStreamer.listenersMutex`** (the send happens inside `notifyListeners`, which holds the mutex for the duration of the callback).

On the normal migration path there is a dedicated receiver:

```go
if !mgtr.migrationContext.Resume {
    <-mgtr.ghostTableMigrated
}
```

But on the instant-DDL success path (`go/logic/migrator.go`), `Migrate()` returns early right after `finalCleanup()` and **never receives** from `ghostTableMigrated`:

```go
if err := mgtr.applier.AttemptInstantDDL(); err == nil {
    if err := mgtr.finalCleanup(); err != nil {
        return nil
    }
    ...
    return nil
}
```

So the changelog send blocks forever, keeping `listenersMutex` held. `finalCleanup()` then closes the binlog reader, whose rows-event decode callback (`EventsStreamer.shouldDecodeRowsEvent`) needs the **same** mutex, and `BinlogSyncer.Close()` waits (via `WaitGroup`) for that goroutine to exit → permanent deadlock. gh-ost hangs and never completes the migration.

### Reproduction

Run an instant-DDL-eligible migration against MySQL 8.0 with `--attempt-instant-ddl`, e.g. adding a column with a default:

```
gh-ost --attempt-instant-ddl --execute \
  --alter="ADD COLUMN c INT NOT NULL DEFAULT 1" \
  --host=127.0.0.1 --port=3306 --database=db --table=t ...
```

gh-ost applies the instant DDL, logs the "migrated instantly" success, but then hangs in cleanup instead of exiting.

### Proposed fix

Drain the `GhostTableMigrated` signal on the instant-DDL success path before `finalCleanup()`, mirroring the receive already present on the normal path, guarded by `!Resume` (resume migrations never emit the signal). I have a PR ready with the fix plus regression tests.

PR #1736, opened 2026-07-02
Fix instant-DDL deadlock on GhostTableMigrated changelog signal

peterbollen:
Related issue: https://github.com/github/gh-ost/issues/1735

### Description

Fixes a deadlock that hangs gh-ost during cleanup when `--attempt-instant-ddl` is used and the instant `ALTER` succeeds.

`initiateApplier` writes a `GhostTableMigrated` changelog row. The streamer's changelog listener callback (`onChangelogStateEvent`) publishes that signal synchronously via `base.SendWithContext(ctx, ghostTableMigrated, true)` while holding `EventsStreamer.listenersMutex` (the send runs inside `notifyListeners`, which holds the mutex for the whole callback).

On the normal path there is a matching `<-mgtr.ghostTableMigrated` receiver. The instant-DDL success path, however, returns early right after `finalCleanup()` and never receives, so the send blocks forever holding `listenersMutex`. `finalCleanup()` then closes the binlog reader, whose rows-event decode callback (`shouldDecodeRowsEvent`) needs the same mutex, and `BinlogSyncer.Close()` waits for that goroutine to exit → permanent deadlock.

**Fix:** drain the `GhostTableMigrated` signal on the instant-DDL success path before `finalCleanup()`, mirroring the receive on the normal path. The drain is extracted into `Migrator.drainGhostTableMigrated()` and guarded by `!Resume`, since resume migrations never emit the signal (`initiateApplier` only writes it when `!Revert && !Resume`).

### Tests

- `TestMigratorDrainGhostTableMigrated` — the drain consumes the signal and unblocks the publisher; it is skipped for resume migrations.
- `TestEventsStreamerInstantDDLDeadlockIsResolvedByDraining` — reproduces the exact deadlock (listener callback blocked on the send while holding `listenersMutex`; `shouldDecodeRowsEvent` blocked on the same mutex) and proves that draining the signal resolves it.

Both pass with `-race`.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -471,6 +471,28 @@ func (mgtr *Migrator) checkAbort() error {
 	return nil
 }
 
+// drainGhostTableMigrated consumes the "GhostTableMigrated" changelog signal that
+// initiateApplier emits once the ghost table is in place. It must be called on the
+// instant-DDL success path before finalCleanup runs.
+//
+// The streamer's changelog listener callback (onChangelogStateEvent) publishes this
+// signal synchronously while holding the streamer's listenersMutex. On the normal
+// migration path there is a dedicated receiver (see below in Migrate), but the
+// instant-DDL path returns early and would otherwise never receive. The blocked send
+// keeps listenersMutex held forever, so when finalCleanup closes the binlog reader
+// its rows-event decode callback (shouldDecodeRowsEvent) deadlocks trying to acquire
+// the same mutex. Draining the signal here releases the listener and avoids the
+// deadlock.
+//
+// Resume migrations never emit the signal (initiateApplier only writes it when
+// !Revert && !Resume), so there is nothing to drain in that case.
+func (mgtr *Migrator) drainGhostTableMigrated() {
+	if mgtr.migrationContext.Resume {
+		return
+	}
+	<-mgtr.ghostTableMigrated
+}
+
 // Migrate executes the complete migration logic. This is *the* major gh-ost function.
 func (mgtr *Migrator) Migrate() (err error) {
 	mgtr.migrationContext.Log.Infof("Migrating %s.%s", sql.EscapeName(mgtr.migrationContext.DatabaseName), sql.EscapeName(mgtr.migrationContext.OriginalTableName))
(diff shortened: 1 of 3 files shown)


ggilder:
Independent confirmation of the diagnosis in #1735, plus one issue with the fix as written.

**Confirming the bug.** We hit this in CI while bumping gh-ost from a 2026-05-14 build to v1.1.11. Our instant-DDL integration tests hung indefinitely — the migration stayed in `running` forever, never reaching a terminal state, exactly as described here. Worth noting for anyone triaging severity: the deadlock isn't reachable before #1699. `shouldDecodeRowsEvent` is what introduces a second contender for `listenersMutex`, so prior to that commit the blocked send merely leaked a goroutine. #1699 is what turns it into a hard hang, which makes this a regression for anyone on `--attempt-instant-ddl` upgrading across it.

**The concern with the fix.** `drainGhostTableMigrated()` uses a bare blocking receive:

```go
func (mgtr *Migrator) drainGhostTableMigrated() {
	if mgtr.migrationContext.Resume {
		return
	}
	<-mgtr.ghostTableMigrated
}
```

The sender publishes via `base.SendWithContext`, which **abandons the send** when the context is cancelled. That means a bare receive on this channel becomes permanently unsatisfiable at exactly the moment the context is cancelled — so this trades the instant-DDL deadlock for an abort-path deadlock on the same channel.

Concretely: `AttemptInstantDDL()` retries up to `--default-retries` (#1667), and the heartbeat escalates to `PanicAbort` after its own successive-failure threshold. If the primary becomes unreachable during those retries, `abort()` cancels the context, the pending changelog send gives up, and `drainGhostTableMigrated()` then blocks forever — after the instant DDL may already have been applied. `Migrate()` never returns, its deferred `teardown()` never runs, and because the status/throttler tickers exit on `finishedMigrating` rather than on the context, the process stays alive logging a frozen status line.

**Suggestion** — mirror #1677 here:

```go
func (mgtr *Migrator) drainGhostTableMigrated() error {
	if mgtr.migrationContext.Resume {
		return nil
	}
	select {
	case <-mgtr.ghostTableMigrated:
		return nil
	case <-mgtr.migrationContext.GetContext().Done():
		return mgtr.checkAbort()
	}
}
```

For what it's worth, I've opened #1758 applying the same treatment to the *other* remaining bare receive on this channel — the `<-mgtr.ghostTableMigrated` on the normal path in `Migrate()`. The two changes are complementary and touch adjacent lines, so they'll likely need light coordination on whichever lands second. Happy to fold this drain into #1758 instead if that's easier for review, or to help rebase this one (it's currently showing as conflicting).

ggilder:
I've pushed a more complete fix to https://github.com/github/gh-ost/pull/1758 ; if that merges, this PR can be closed

Issue #1738, opened 2026-07-28
Release binary reports 20 HIGH CVEs (Go toolchain, x/crypto, docker/docker)


davidham:
## Summary

A Trivy scan of the `gh-ost` binary shipped in the v1.1.10 release reports **20 HIGH severity CVEs** (0 critical). All of them come from the Go toolchain the binary was built with, and from two modules recorded in its build info.

This is a practical blocker for regulated deployments. We embed the released binary in a downstream image for a FedRAMP-authorized platform, which cannot ship an image containing any Critical or High CVE.

## Findings

Scanning the v1.1.10 `linux/amd64` binary (`trivy rootfs --scanners vuln --severity HIGH,CRITICAL`):

| Source | Version in binary | Count | CVEs |
|---|---|---|---|
| Go stdlib (build toolchain) | go1.25.9 | 8 | CVE-2026-27145, -33811, -33814, -39820, -39822, -39836, -42499, -42504 |
| `golang.org/x/crypto` | v0.45.0 | 9 | CVE-2026-39828, -39829, -39830, -39831, -39832, -39835, -42508, -46595, -46597 |
| `github.com/docker/docker` | v28.0.1+incompatible | 3 | CVE-2026-34040, CVE-2026-41567, CVE-2026-42306 |

## Why `docker/docker` is in a schema-migration binary

`go/logic/test_utils.go` has no `_test.go` suffix, so it is part of the ordinary build of `package logic`. It imports `testcontainers-go`, and `package logic` is imported by `go/cmd/gh-ost`, so `testcontainers` and its transitive dependencies -- including `github.com/docker/docker` -- get compiled into the released binary and recorded in its build info.

`go version -m` on the v1.1.10-equivalent binary records **60 modules**, most of which are container-testing infrastructure that never runs in production.

Notably, this cannot be fixed by bumping the dependency. All three docker CVEs are first patched in 29.3.1, but at v29 the module was renamed to `github.com/moby/moby/v2`, and the v29 tags are named `docker-v29.x.y`, which Go cannot resolve as module versions. The `github.com/docker/docker` path is frozen at `v28.5.2+incompatible`, so there is no reachable fixed version on that path.

## Suggested fix

[...]

PR #1739, opened 2026-07-28
refactor(logic): scope testcontainers helpers to the test build

davidham:
Related issue: https://github.com/github/gh-ost/issues/1738

[Update 7/29/2026] Rebased onto current `master` and reduced to the single commit you asked for, @ericyan. The toolchain and dependency bumps are dropped.

### Description

`go/logic/test_utils.go` has no `_test.go` suffix, so it is part of the ordinary build of `package logic`. It imports `testcontainers-go`, and `package logic` is imported by `go/cmd/gh-ost`, so testcontainers and its transitive dependencies are compiled into the released binary and recorded in its build info.

This PR renames the file to `test_utils_test.go`, so it stays available to the tests in the same package while being excluded from the ordinary build. Every identifier it declares is referenced only from `applier_test.go`, `streamer_test.go` and `migrator_test.go`, so it is a **pure rename -- zero line changes**. The rebase preserves your `port.Num()` change from #1741; the file content is byte-identical to `master`.

### Measured effect

`linux/amd64`, `CGO_ENABLED=0`, built with go1.25.12:

| | master | this PR |
|---|---|---|
| modules in build info | 63 | 22 |
| binary size | 18.64 MB | 16.99 MB |
| Trivy HIGH/CRITICAL | 0 | 0 |

To be explicit about the framing, since it has changed since I opened this: **this PR fixes no current CVE.** After #1741 and #1742, `master` already scans clean, and so does this branch. The value here is what you identified -- `testcontainers-go` does not belong in the shipped binary. It drops 41 modules and ~1.65 MB, and it means a future advisory anywhere in the container-testing tree will not surface in scans of a binary that never calls into that code. The 3 `docker/docker` CVEs in v1.1.10 were a symptom of this; the upgrade cleared them, the rename removes the mechanism.

### Verification

- `script/cibuild` -- clean: gofmt, build, and full unit suite including the testcontainers-backed `go/logic` tests (61.4% coverage) against testcontainers v0.43.0
- `golangci-lint run --config=.golangci.yml` -- 0 issues


ericyan:
Hi @davidham, thanks for your contribution.

I think the third commit is the one that brings unique value here. Would you cherry-pick that commit for this PR? I'd prefer to address the toolchain and dependencies updates separately.

davidham:
Hi @ericyan, thanks for your reply and the dependency updates. I rebased onto master and kept only the rename commit. Have a great day!

ericyan (approved):

ericyan:
Thank you for your contribution ❤️

PR #1746, opened 2026-07-31
move-tables: support generated and JSON columns

ericyan:
### Description

This PR fixes generated column handling in move-tables mode.

#### Generated columns support

Move-tables mode set `SharedColumns` and `MappedSharedColumns` to the full source column list:

```go
mt.SharedColumns = columns
mt.MappedSharedColumns = columns
```

Since `columns` includes generated columns, both the row-copy `INSERT` and the binlog-driven `INSERT`/`UPDATE` named them in their value lists, and MySQL rejected the migration:

```text
Error 3105 (HY000): The value specified for generated column 'x' in table 'y' is not allowed.
```
Standard migrations avoid this because `getSharedColumns()` removes generated columns; move-tables mode had no equivalent.

This is fixed by excluding generated columns from the writable set and then route row-copy builders through the filtered list.

#### Missing column metadata

Investigating the reported schema surfaced a second, independent defect. `applyColumnTypes` was only ever called on unique-key columns in move-tables mode, so `Type`, `IsUnsigned`, `Charset`, `MySQLType`, and `BinaryOctetLength` were empty everywhere else. That silently disabled encoding logic in `convertArg` and `buildColumnsPreparedValues`:

| Missing metadata | Consequence |
| --- | --- |
| `IsUnsigned` | `BIGINT UNSIGNED` values above 2^63 written to the target as negative numbers |
| `Type == JSONColumnType` | JSON bound as plain `?` instead of `convert(? using utf8mb4)` |
| `BinaryOctetLength` | Fixed-width `BINARY` not zero-padded (#909) |
| `MySQLType` binary/blob | Sent as `MYSQL_TYPE_VAR_STRING` rather than `MYSQL_TYPE_BLOB`, producing Warning 1300 under `--panic-on-warnings` |

These column metadata are now populated.

---

### DIFF
--- a/go/base/context.go
+++ b/go/base/context.go
@@ -100,8 +100,8 @@ type MoveTable struct {
 	CreateTableStatement string
 
 	// Schema, captured from the source (or from the target, on resume). In
-	// move-tables mode source and target schemas match, so the shared columns are
-	// identical to the original columns.
+	// move-tables mode source and target schemas match, so shared columns are the
+	// original columns minus generated columns, which MySQL recomputes on the target.
 	OriginalTableColumns        *sql.ColumnList
 	OriginalTableVirtualColumns *sql.ColumnList
 	OriginalTableUniqueKeys     [](*sql.UniqueKey)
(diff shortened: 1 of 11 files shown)


zacharysierakowski (approved):
🚀

PR #1747, opened 2026-08-05
Add opt-in --analyze-ghost-table-before-cutover

VandhanaSelvaprakash-at:
Supersedes #1419 (original approach and credit to @wangzihuacool) and closes the gap discussed in #1418. Thanks @ericyan / @timvaillancourt for the go-ahead.

## What

Adds an opt-in `--analyze-ghost-table-before-cutover` flag (default off). When set, gh-ost runs an explicit `ANALYZE TABLE` on the ghost table immediately before cut-over, so the freshly swapped table doesn't briefly serve traffic with near-zero InnoDB row estimates — which the optimizer can cost as a free full scan, flipping plans on hot paths until statistics recompute.

## Why opt-in

Per @shaohk and @timvaillancourt on #1419: on partitioned tables `ANALYZE` cost grows with partition count and the statement replicates. So it's gated behind a flag, default off, intended for small non-partitioned tables — opt-in users can report on performance.

## How it differs from #1419

1. **Runs after the postpone gate releases** (before the source lock and before `--test-on-replica` stops replication) — a postponed cut-over doesn't re-stale its statistics before the swap.
2. **A failed ANALYZE aborts the migration (fatal), fail-closed.** `ANALYZE TABLE` surfaces table-level failures (missing table, storage-engine errors) as `Msg_type=Error` result rows while the statement succeeds at the protocol level, so the rows are inspected and cut-over is refused unless status is OK with no Error rows.

## Tests

- `TestClassifyAnalyzeTableResult` — DB-free table test of the result-row classifier (status-OK, case folding, error rows, full-scan ordering, non-OK status, empty result).
- `ApplierTestSuite.TestAnalyzeGhostTable` — real-MySQL test covering the happy path, the fail-open regression, and the statement-error branch.
 - `TestAnalyzeGhostTableBeforeCutOver` — migrator-level test of the cut-over orchestration contract: the flag gates the call, a successful ANALYZE runs exactly once, and a failed ANALYZE propagates so `cutOver()` aborts (via `Log.Fatale`) before replica-stop or cut-over locking begins. The gating lives in a narrow `Migrator.analyzeGhostTableBeforeCutOver(analyze func() error)` seam so the contract is testable without a live applier.
  - `localtests/analyze-ghost-table-before-cutover` — end-to-end migration with the flag enabled and ongoing DML, exercising the real cut-over ANALYZE path against MySQL.
  
Also folds in a separable pre-existing fix: `TeardownSuite` → `TearDownSuite` across the applier/migrator/streamer suites (testify never invoked the misspelled method, leaking a testcontainer per suite). Happy to split it into its own PR if you'd prefer.

DCO signed off.

### DIFF
--- a/go/logic/applier.go
+++ b/go/logic/applier.go
@@ -566,18 +566,18 @@ func classifyAnalyzeTableResult(tableName string, rows []analyzeTableResultRow) 
 // cannot prove plan safety — plan checks belong to the orchestrating layer, which
 // knows the table's context. The caller must treat a returned error as fatal,
 // not retriable.
-func (this *Applier) AnalyzeGhostTable() error {
+func (apl *Applier) AnalyzeGhostTable() error {
 	query := fmt.Sprintf(`analyze /* gh-ost */ table %s.%s`,
-		sql.EscapeName(this.migrationContext.DatabaseName),
-		sql.EscapeName(this.migrationContext.GetGhostTableName()),
+		sql.EscapeName(apl.migrationContext.DatabaseName),
+		sql.EscapeName(apl.migrationContext.GetGhostTableName()),
 	)
-	this.migrationContext.Log.Infof("Running ANALYZE TABLE on ghost table %s.%s before cut-over",
-		sql.EscapeName(this.migrationContext.DatabaseName),
-		sql.EscapeName(this.migrationContext.GetGhostTableName()),
+	apl.migrationContext.Log.Infof("Running ANALYZE TABLE on ghost table %s.%s before cut-over",
+		sql.EscapeName(apl.migrationContext.DatabaseName),
+		sql.EscapeName(apl.migrationContext.GetGhostTableName()),
 	)
 	analyzeStartTime := time.Now()
 	var rows []analyzeTableResultRow
-	err := sqlutils.QueryRowsMap(this.db, query, func(rowMap sqlutils.RowMap) error {
+	err := sqlutils.QueryRowsMap(apl.db, query, func(rowMap sqlutils.RowMap) error {
 		rows = append(rows, analyzeTableResultRow{
 			msgType: rowMap.GetString("Msg_type"),
 			msgText: rowMap.GetString("Msg_text"),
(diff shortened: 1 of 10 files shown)


go/logic/applier.go
```diff
@@ -526,6 +526,78 @@ func (apl *Applier) CreateGhostTable() error {
...
+// AnalyzeGhostTable runs an explicit ANALYZE TABLE on the ghost table, forcing a
+// synchronous InnoDB persistent-statistics recompute before cut-over. Without it the
+// freshly swapped table can serve traffic with a near-zero row estimate, which the
+// optimizer costs as a free full scan — the failure mode motivating upstream #1419.
+// No row-count assertion follows the ANALYZE: on a freshly built, compact ghost a
+// successful ANALYZE yields correct statistics by construction, and a row count
+// cannot prove plan safety — plan checks belong to the orchestrating layer, which
+// knows the table's context. The caller must treat a returned error as fatal,
+// not retriable.
+func (this *Applier) AnalyzeGhostTable() error {
```
ericyan: It is probably best to follow the Go convention:

```suggestion
func (apl *Applier) AnalyzeGhostTable() error {
```

ericyan (commented):
Hi @VandhanaSelvaprakash-at, thanks for the PR! This looks great overall. Just need a tiny change to make the linter happy.


Could we add a small, deterministic `Migrator-level` test for this as well?

The applier tests cover `ANALYZE` execution and result parsing well, but they do not verify the cut-over orchestration contract: that the flag gates the call, it runs only after postpone releases, and an analysis failure stops before replica-stop or cut-over locking/retry work begins. A narrow test seam around the analysis invocation is fine; no need to abstract the full `Applier` just for this.

Adding `localtest` for this case would also be nice, but that is not required.

go/logic/applier.go
VandhanaSelvaprakash-at: addressed in follow-up commit

VandhanaSelvaprakash-at:
> Hi @VandhanaSelvaprakash-at, thanks for the PR! This looks great overall. Just need a tiny change to make the linter happy.
> 
> Could we add a small, deterministic `Migrator-level` test for this as well?
> 
> The applier tests cover `ANALYZE` execution and result parsing well, but they do not verify the cut-over orchestration contract: that the flag gates the call, it runs only after postpone releases, and an analysis failure stops before replica-stop or cut-over locking/retry work begins. A narrow test seam around the analysis invocation is fine; no need to abstract the full `Applier` just for this.
> 
> Adding `localtest` for this case would also be nice, but that is not required.

 Thanks @ericyan! All three addressed in the latest commit:

  1. Linter — renamed the AnalyzeGhostTable receiver this → apl to match the rest of applier.go (staticcheck ST1016, which was the golangci-lint failure). golangci-lint run ./... at v2.11 (matching CI) is now clean.
  2. Migrator-level test — added TestAnalyzeGhostTableBeforeCutOver, backed by a narrow seam Migrator.analyzeGhostTableBeforeCutOver(analyze func() error) that owns the gating (no full-Applier abstraction, per your note; behavior unchanged). It covers the orchestration contract: the flag gates the call, a successful ANALYZE runs exactly once, and a failed ANALYZE propagates an error so cutOver() aborts via Log.Fatale before replica-stop or the cut-over lock/retry switch. The placement relative to the postpone gate is fixed structurally by the call site.
  3. localtest — added analyze-ghost-table-before-cutover: a small non-partitioned table with ongoing DML and the flag enabled, so the real cut-over path runs ANALYZE end-to-end before the swap. Passes locally against 8.0.41.

ericyan (approved):

Issue #884, opened 2020-09-13
gh-ost hang on INFO Waiting for ghost table to be migrated. Current lag is 0s in MySQL 8.0.21


barucho:
i am runnign gh-ost and the gh-ost is not passing the stage 

```
INFO Waiting for ghost table to be migrated. Current lag is 0s
```
```+-----------+
| @@version |
+-----------+
| 8.0.21    |

```

```
/usr/bin/gh-ost  --max-load=Threads_running=500   --critical-load=Threads_running=1000  --initially-drop-ghost-table --initially-drop-old-table  --chunk-size=600  --max-lag-millis=3000 --user=dbschema --password=**** --assume-master-host=pdb-mysql-pii-sdl-playground0a.42.net --throttle-control-replicas=pdb-mysql-pii-sdl-playground0a.96.net --host=pdb-mysql-pii-sdl-playground1b.42.net --database=test --table=T1 --verbose  --alter="add e varchar(2),add index e_idx (e)"  --assume-rbr  --allow-master-master  --cut-over=default  --concurrent-rowcount  --default-retries=120  --panic-flag-file=/tmp/ghost.panic.flag  --exact-rowcount  --postpone-cut-over-flag-file=/tmp/ghost.postpone.flag --debug --execute
```

````
2020-09-13 19:18:55 INFO starting gh-ost 1.0.49
2020-09-13 19:18:55 INFO Migrating `test`.`T1`
2020-09-13 19:18:55 INFO connection validated on pdb-mysql-pii-sdl-playground1b.42.net:3306
2020-09-13 19:18:55 INFO User has REPLICATION CLIENT, REPLICATION SLAVE privileges, and has ALL privileges on `test`.*
2020-09-13 19:18:55 INFO binary logs validated on pdb-mysql-pii-sdl-playground1b.42.net:3306
2020-09-13 19:18:55 INFO Inspector initiated on pdb-mysql-pii-sdl-playground1b.42.net:3306, version 8.0.21
2020-09-13 19:18:55 INFO Table found. Engine=InnoDB
2020-09-13 19:18:55 DEBUG Estimated number of rows via STATUS: 80
2020-09-13 19:18:55 DEBUG Validated no foreign keys exist on table
2020-09-13 19:18:55 DEBUG Validated no triggers exist on table
2020-09-13 19:18:55 INFO Estimated number of rows via EXPLAIN: 80
[...]

Issue #380, opened 2017-03-01
Waiting for ghost table to be migrated... forever?


lrouxel:
Hi,

Not sure if I'm hitting a bug but I don't really understand what's happening.

My setup : 
Two Percona Server 5.6.34 as active-passive master-master (statement-based replication).
(IPs are 10.42.10.80 and 10.42.10.81)
A third one, slave of the active master, with row-based replication and log-slave-updates as required.
(IP is 10.42.20.83)

I'm running gh-ost on the third one.
I've tried running dry run migrations on small or empty tables and... nothing happens, it gets stuck forever with no debug info.

The ghost tables are created on both masters, and the show processlist shows connections in Sleep mode on all 3 servers.

Table is created like this : 

```
mysql> create table emptytable (id int unsigned not null primary key);
Query OK, 0 rows affected (0.01 sec)
```

And here we go : 

```
./gh-ost \
--conf=/root/.my.cnf \
--max-load=Threads_running=3000 \
--critical-load=Threads_running=5000 \
--max-lag-millis=60000 \
--database="test" \
--table="emptytable" \
--verbose \
--alter="engine=innodb" \
--allow-master-master \
--exact-rowcount \
--panic-flag-file=/tmp/ghost.panic.flag \
--postpone-cut-over-flag-file=/tmp/ghost.postpone.flag \
--skip-foreign-key-checks \
--debug \
--host=10.42.20.83
2017-03-01 23:14:15 INFO starting gh-ost 1.0.34
2017-03-01 23:14:15 INFO Migrating `test`.`emptytable`
2017-03-01 23:14:15 INFO connection validated on 10.42.20.83:3306
2017-03-01 23:14:15 INFO User has ALL privileges
2017-03-01 23:14:15 INFO binary logs validated on 10.42.20.83:3306
2017-03-01 23:14:15 INFO Restarting replication on 10.42.20.83:3306 to make sure binlog settings apply to replication thread
2017-03-01 23:14:15 DEBUG Replication restarted
2017-03-01 23:14:15 INFO Table found. Engine=InnoDB
2017-03-01 23:14:15 DEBUG Estimated number of rows via STATUS: 0
2017-03-01 23:14:15 WARNING --skip-foreign-key-checks provided: will not check for foreign keys
2017-03-01 23:14:15 DEBUG Validated no triggers exist on table
2017-03-01 23:14:15 INFO Estimated number of rows via EXPLAIN: 1
[...]

PR #1758, opened 2026-08-20
Handle context cancellation while waiting for ghost table migration

ggilder:
### Description

`Migrate()` blocks on an unbuffered receive from `ghostTableMigrated` while waiting for the ghost table to be created:

```go
mgtr.migrationContext.Log.Infof("Waiting for ghost table to be migrated. Current lag is %+v", initialLag)
<-mgtr.ghostTableMigrated
```

The only sender is `onChangelogStateEvent()`, which publishes via `base.SendWithContext()`. If the migration aborts during this window, `abort()` cancels the migration context, so `SendWithContext()` takes its `ctx.Done()` branch and returns *without ever sending*. Nothing else writes to the channel, so `Migrate()` blocks forever:

- `Migrate()` never returns, so its deferred `teardown()` never runs
- `finishedMigrating` is therefore never set
- the status and throttler tickers exit on `finishedMigrating` rather than on the context, so they keep looping

The result is a process that has already decided to abort but stays alive indefinitely, logging a frozen status line, until it is killed externally. This is the same type of deadlock, and the same fix, as #1677 applied to `consumeRowCopyComplete()` — `ghostTableMigrated` is its remaining sibling.

### Change

Extract the wait into `waitForGhostTableMigrated()` and select on the migration context alongside the channel, returning `checkAbort()` so the original abort error is surfaced rather than a bare context error. The extraction mirrors `consumeRowCopyComplete()` and makes the behaviour testable without a database.

`TestAbort_DuringGhostTableWait` follows the existing `TestAbort_*` convention and blocks until its timeout (i.e. fails) without this change; `TestWaitForGhostTableMigrated` covers the normal path.

### Possibly related

Not claiming these are fixed by this change — their root causes aren't established, and this only addresses hangs caused by an abort cancelling the context during the wait — but they are reports of hanging at this exact point, so noting them for reference: #884, #380. #1735 concerns the same `ghostTableMigrated` channel in a different scenario.

On `script/cibuild`: formatting, build, and all unit tests pass. The three testcontainers-based suites (`TestApplier`, `TestMigrator`, `TestEventsStreamer`) cannot run in my environment — no rootless Docker provider — and fail identically on unmodified `master`, so they are unaffected by this change and left to CI.

### DIFF
--- a/go/logic/migrator.go
+++ b/go/logic/migrator.go
@@ -267,6 +267,22 @@ func (mgtr *Migrator) consumeRowCopyComplete() {
 	}()
 }
 
+// waitForGhostTableMigrated blocks until the ghost table has been migrated, or
+// until the migration context is cancelled by an abort. The only sender on
+// ghostTableMigrated publishes via base.SendWithContext, which stops sending
+// once the context is cancelled, so waiting on the channel alone would block
+// forever after an abort.
+func (mgtr *Migrator) waitForGhostTableMigrated() error {
+	select {
+	case <-mgtr.ghostTableMigrated:
+		mgtr.migrationContext.Log.Debugf("ghost table migrated")
+		return nil
+	case <-mgtr.migrationContext.GetContext().Done():
+		// Abort cancelled the context
+		return mgtr.checkAbort()
+	}
+}
+
 func (mgtr *Migrator) canStopStreaming() bool {
 	return atomic.LoadInt64(&mgtr.migrationContext.CutOverCompleteFlag) != 0
 }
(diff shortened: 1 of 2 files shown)


ggilder:
Passing CI on my fork https://github.com/ggilder/gh-ost/pull/8

ggilder:
@meiji163 any thoughts on this?

ggilder:
@timvaillancourt @meiji163 @ericyan I've updated this PR to handle the two remaining deadlock issues in gh-ost that I've been able to identify. Build is passing on my fork: https://github.com/ggilder/gh-ost/pull/8

Please take a look when you get a chance, thanks!

Issue #1630, opened 2026-02-11
Optimize the lock time in the cutOver phase


dongwenpeng:
We found that gh-ost always requires a lock time of 1 second or even longer during the cutOver phase, regardless of whether the database is busy.

Log:
```
Done waiting for events up to lock; duration=975.482481ms
```

This indicates that the write lock duration on the change table is approximately 1 second.

The investigation revealed that the source of the long-term lock holding was time.Sleep().

Source code:

```
func (this *Migrator) executeWriteFuncs() error {
	...
	for {
		select {
		case eventStruct := <-this.applyEventsQueue:
			{
				if err := this.onApplyEventStruct(eventStruct); err != nil {
					return err
				}
			}
		default:
			{
				select {
				case copyRowsFunc := <-this.copyRowsQueue:
					{
						...
					}
				default:
					{
						// Hmmmmm... nothing in the queue; no events, but also no row copy.
						// This is possible upon load. Let's just sleep it over.
						this.migrationContext.Log.Debugf("Getting nothing in the write queue. Sleeping...")
						time.Sleep(time.Second)
					}
				}
			}
		}
	}
}
```

The presence of `time.Sleep(time.Second)` causes the `waitForEventsUpToLock` function in the `cutOver` phase to be unable to quickly obtain the binlog processing completion notification, resulting in an excessively long locking time.

Because the binlog processing completion notification is in the `this.applyEventsQueue`, it is ultimately triggered by the execution of the `this.onApplyEventStruct()` function.

I tried removing `time.Sleep` and running `gh-ost` again to get the lock time.

```
Done waiting for events up to lock; duration=9.305327ms
```

Thanks

Issue #1191, opened 2022-10-21
Cut-off transaction is very slow, leading to downtimes in production[Question]


Shukla-Ankur:
MySql v8.0.23

1. I recently did multiple migrations in production. I noticed for table switch at the end, gh-ost takes a very long time. here are actual logs from production. <b>
2. The row count in these tables ranged from 15k to 5M. Also, I did not see any correlation between table size and lock duration, which is as expected.  <b>
3. Below are actual logs from production run. 
`2022-10-19 07:53:14 INFO Lock & rename duration: 1.017494674s. During this time, queries on `entity` were blocked
2022-10-19 07:21:28 INFO Lock & rename duration: 1.042562025s. During this time, queries on `entity` were blocked
2022-10-19 07:50:31 INFO Lock & rename duration: 2.022935482s. During this time, queries on `entity` were blocked
2022-10-19 07:45:12 INFO Lock & rename duration: 2.020735566s. During this time, queries on `entity` were blocked
2022-10-19 07:42:29 INFO Lock & rename duration: 1.01602586s. During this time, queries on `entity` were blocked`

4. Such high cut off time means a downtime. I expected some single digit millisecond operation for this transaction. One question here- does this timestamp include any lock wait time also for tx to even acquire a lock? However, these tables almost did not have any traffic.

dm-2:
:wave: @Shukla-Ankur the [cut-over step](https://github.com/github/gh-ost/blob/master/doc/cut-over.md) is a [complicated process](https://github.com/github/gh-ost/issues/82) which requires coordination between a few different operations to ensure that the table cut-over happens atomically - and as a result, it takes some time to complete.

The times you are seeing are normal, this is a sample of cutover times for some  of our recent schema migrations (on both very small and very large tables):
```
INFO Lock & rename duration: 1.034362751s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 1.958744449s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 2.274112085s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 5.060551315s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 1.622521302s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 1.026753275s. During this time, queries on <redacted> were blocked
INFO Lock & rename duration: 1.042612384s. During this time, queries on <redacted> were blocked
[...]

PR #1772, opened 2026-09-19
perf(cutover): drop two 1s sleeps from the locked cut-over window

conf:

Related issue: https://github.com/github/gh-ost/issues/1630

### Description

This PR removes two fixed 1-second sleeps that run **while the original table is write-locked**, i.e. while every query on it is blocked.

#### 1. The idle sleep in `executeWriteFuncs` (the one reported in #1630)

When both the events queue and the row-copy queue are empty, the loop slept a flat second. Once row copy is done that is the steady state, so the `AllEventsUpToLockProcessed` sentinel — which arrives on `applyEventsQueue` *after* the tables are locked — had to wait out the remainder of that sleep:

```
Done waiting for events up to lock; duration=990.976125ms
```

It now blocks on both queues with a 1s timeout instead, which is the shape `executeDMLWriteFuncs` already uses a few lines below. Same log line after:

```
Done waiting for events up to lock; duration=1.314334ms
```

#### 2. The `retryOperation` backoff in `waitForRename`, one step later

With #1 fixed the cut-over still took ~1s, and the remaining second is in the next phase. After issuing the atomic RENAME in a goroutine, gh-ost polls `SHOW PROCESSLIST` for it through `retryOperation`, whose backoff is a flat 1 second. The first poll usually loses the race against the RENAME registering its metadata-lock wait, so it paid that full second — again, with the table locked:

```
22:36:39 Done waiting for events up to lock; duration=1.314334ms
22:36:39 Issuing and expecting this to block: rename ...
22:36:40 Found atomic RENAME to be blocking, as expected.
22:36:40 Lock & rename duration: 1.019256291s
```

Across the full `localtests` suite this hit 6 of the 73 cases that reach cut-over (~8%), which is why it does not show on every run.

It now polls every 10ms, for a budget of twice the RENAME's own `lock_wait_timeout` — past that the RENAME has errored out and set `tableRenameKnownToHaveFailed`, so `waitForRename` returns immediately and the attempt count never runs out.

Both call sites go through a new `retryOperationWithInterval`, which is the existing `retryOperation` body with the attempt count and wait made explicit; `retryOperation` now delegates to it with its previous `MaxRetries()`/1s values. No behaviour change for its other callers.

### Results

`Lock & rename duration` over the 73 `localtests` cases that reach cut-over. Before, it was ~1s on essentially every cut-over:

| | p50 | p90 | p99 | max |
|---|---|---|---|---|
| mariadb:11.8 | 9ms | 21ms | 90ms | 90ms |
| mysql:8.4.3 | 23ms | 31ms | 96ms | 96ms |

The remaining tail is **not** this code path. It is `waitForEventsUpToLock`: on both flavours the same four cases, which happen to see no DML during the migration, spend 54–77ms waiting for the sentinel to come back through an otherwise idle binlog stream. No case with concurrent DML exceeds 50ms on either flavour (0/60 and 0/59). That delay reproduces with a plain `mariadb-binlog --read-from-remote-server` client, so it is in binlog delivery rather than in gh-ost, and it predates this change.

### Test change

`TestCutOverLossDataCaseLockGhostBeforeRename` took its lock on the ghost table one second *after* removing the postpone flag, relying on cut-over still being in flight. With cut-over now ~10ms the ghost table was already renamed, giving `Table 'test._testing_gho' doesn't exist`. It now takes the lock before un-postponing, which is what the test name describes anyway and no longer depends on cut-over being slow.

### Verification

- `script/cibuild` — clean (gofmt, build, `go test -race ./go/...`).
- `script/docker-gh-ost-replica-tests run` — full suite passes on `mariadb:11.4`, `mariadb:11.8`, `mysql:8.0.41` and `mysql:8.4.3`.
- Per-case `Lock & rename duration` measured on `mariadb:11.8` and `mysql:8.4.3` for the table above.


terev:
Related to https://github.com/github/gh-ost/issues/1191 as well.

conf:
@ericyan could you please have a look? 
Thanks.

